PR #13, opened 2018-04-19
Defined WorkspaceClientCapabilites to fix Eclipse JDT LS errors

peanutbutterspice:
## Overview
This addresses a null pointer exception that is thrown by the Eclipse JDT language server (https://github.com/eclipse/eclipse.jdt.ls) when the workspace has no capabilities defined.

## Analysis
The null pointer is thrown here: https://github.com/eclipse/eclipse.jdt.ls/blob/f364a491a701add3639c42803b0afc8da6df4b50/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/preferences/ClientPreferences.java#L95
Which is called from here: https://github.com/eclipse/eclipse.jdt.ls/blob/edffe44f04172cc05174df43f1acf220427f202d/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/InitHandler.java#L157

I can provide any additional information required, etc.

Edit:

Also note the typo in WorkspaceClientCapabilities. The typo is on their end: https://github.com/gluon-lang/languageserver-types/blob/master/src/lib.rs#L636

### DIFF
--- a/src/general.rs
+++ b/src/general.rs
@@ -9,7 +9,9 @@ use url::Url;
 pub fn initialize(root_path: &str, meta: EditorMeta, ctx: &mut Context) {
     let params = InitializeParams {
         capabilities: ClientCapabilities {
-            workspace: None,
+            workspace: Some(WorkspaceClientCapabilites {
+                ..WorkspaceClientCapabilites::default()
+            }),
             text_document: Some(TextDocumentClientCapabilities {
                 synchronization: None,
                 completion: Some(CompletionCapability {


ul:
Thank you for contribution! Looks good, I just wonder if we can replace

```
WorkspaceClientCapabilites {
    ..WorkspaceClientCapabilites::default()
}
```

with just `WorkspaceClientCapabilites::default()` ?

I created PR for typo you mentioned: https://github.com/gluon-lang/languageserver-types/pull/54

peanutbutterspice:
Not to my knowledge, no. Keep in mind I'm no rust expert. My time debugging this issue has been my only real foray into using the language at all. Based on the doc here: https://doc.rust-lang.org/beta/std/default/trait.Default.html though, those two periods are required.

ul:
No worries, I'm novice in Rust as well and `kak-lsp` is my way to learn language =)

Yes, `..` is required to spread default structure inside newly created one, but in that particular case we can skip creating that new structure in the first place, because we are not altering it (in example you referred one field is changed and other are kept default using spread)

I'll merge it as is and then simplify later. Thanks again for contribution, I hope `kak-lsp` is working for you, please don't hesitate with further PRs/issues.

peanutbutterspice:
Ah, yeah, that makes sense. I'll keep that in mind in the future.

I've got some other stuff planned that can be added as well, I've just gotta get the time to make the changes.

Anyway, thanks for the quick response.

PR #14, opened 2018-04-26
Debug messages, improved completion and cquery semantic highlighting

topisani:
Right, this has a few things, because as always, i forgot to commit in the middle of playing around. I understand if you want me to split it up or only want some parts.

I added some debug messages because i couldnt get it working, and wanted to know what was going on. Turned out you need to add the `-c` flag to gnu-netcat - if that flag doesnt work on OSX, we can remove it and put a note in the readme.

Then i changed completion to

 - use `insertText` by default, instead of `label`
 - display the `label` in the menu, with `kind` next to it, formatted as `MenuInfo`
 - display `detail` and `doc` in the info box.

This plays a lot nicer with cquery and most other lsps, and shouldnt be a problem for any.

Lastly, the whole point of me looking at this, is to switch the cquery plugin to using this, so i added cquery semantic highlighting. Now, i dont know if you want this in the upstream. On the one hand, its fairly simple and self contained, on the other, you may want to keep the repo official lsp stuff only.
Anyway, the cquery stuff should give no issues with non-cquery lsps.

This is the first time i've ever touched rust, so I'm sorry if something is completely off. It does all work nicely though, and great job on this, i think this is the right way to do lsp for kakoune (which is nice, cause  there are at least 3 implementations out there)

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -1,6 +1,9 @@
-decl str lsp_cmd 'nc localhost 31337'
+decl str lsp_cmd 'nc localhost 31337 -c'
 decl -hidden completions lsp_completions
-decl -hidden range-specs lsp_errors 
+decl -hidden range-specs lsp_errors
+
+#cquery
+decl -hidden range-specs cquery_semhl
 
 def lsp-did-change %{
     decl -hidden str lsp_draft %sh{ mktemp }
@@ -100,6 +103,7 @@ method  = "textDocument/didSave"
 def lsp-enable %{
     set buffer completers "option=lsp_completions:%opt{completers}"
     add-highlighter buffer/ ranges lsp_errors
+    add-highlighter buffer/ ranges cquery_semhl
 
     lsp-did-open
 
--- a/src/language_features/completion.rs
+++ b/src/language_features/completion.rs
@@ -5,6 +5,7 @@ use regex::Regex;
 use serde::Deserialize;
 use types::*;
 use url::Url;
+use std;
 
 pub fn text_document_completion(params: EditorParams, meta: &EditorMeta, ctx: &mut Context) {
     let req_params = TextDocumentCompletionParams::deserialize(params.clone())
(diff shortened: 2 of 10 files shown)


ul:
It's very neat improvement, thank you for contribution!

Let me to address your comment now and to review PR code during weekend.

> Right, this has a few things, because as always, i forgot to commit in the middle of playing around. I understand if you want me to split it up or only want some parts.

No worries, PR is perfectly readable.

> I added some debug messages because i couldnt get it working, and wanted to know what was going on.

I'm okay with them, just want to replace later with multi-level logging and allow user to control verbosity level. Created #16 to not forget.

> Turned out you need to add the -c flag to gnu-netcat - if that flag doesnt work on OSX, we can remove it and put a note in the readme.

It works fine on OSX with `-c` flag, but I'm surprised that it's needed on your system, I wonder what was the exact point of failure? Was it reading stream to string or parsing?

> Then i changed completion to ...

Kudos for doing this! It looks much nicer now.

> Lastly, the whole point of me looking at this, is to switch the cquery plugin to using this, so i added cquery semantic highlighting. Now, i dont know if you want this in the upstream. On the one hand, its fairly simple and self contained, on the other, you may want to keep the repo official lsp stuff only.
Anyway, the cquery stuff should give no issues with non-cquery lsps.

Support for server-specific protocol extensions is welcome. kak-lsp never meant to be reference implementation, its primary goal is to be useful tool.

rc/lsp.kak
```diff
@@ -1,6 +1,9 @@
-decl str lsp_cmd 'nc localhost 31337'
+decl str lsp_cmd 'nc localhost 31337 -c'
 decl -hidden completions lsp_completions
-decl -hidden range-specs lsp_errors 
+decl -hidden range-specs lsp_errors
+
+#cquery
```
ul: nit:  option name already has self-explanatory prefix `cquery` so we can skip comment here

src/language_features/completion.rs
```diff
@@ -36,14 +37,35 @@ pub fn editor_completion(
...
         .map(|x| {
+            let mut doc: String = match &x.documentation {
+                None => "".to_string(),
+                Some(doc) => match doc {
+                    Documentation::String(st) => st.clone(),
+                    Documentation::MarkupContent(mup) => mup.value.clone(),
+                },
+            };
+            if let Some(d) = x.detail {
+                doc = d.clone() + "\n\n" + &doc;
```
ul: I think `format!` macro could be a perfect fit here: `doc = format!("{}\n\n{}", d, doc);`

src/language_features/cquery.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+impl<'de> serde::Deserialize<'de> for StorageClass {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        use enum_primitive::FromPrimitive;
+
+        let i = try!(u8::deserialize(deserializer));
```
ul: Could be made even more ergonomic with `?` operator: `let i = u8::deserialize(deserializer)?;`

src/language_features/cquery.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+impl<'de> serde::Deserialize<'de> for SemanticSymbolKind {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        use enum_primitive::FromPrimitive;
+
+        let i = try!(u8::deserialize(deserializer));
```
ul: Ditto `?` operator

src/language_features/cquery.rs
topisani: Oh, i just copied this from the lsp-types package

ul: Probably it's because `?` operator was stabilized recently and more mature codebases have `try!`.

topisani:
no, thats cool. Are you on the kakoune IRC, and under which handle?

ul:
was not; now is; feel free to ping me by `ul` handle

PR #21, opened 2018-05-10
Only send didChange when the buffer did change

mawww:
Track the last sent change in a lsp_timestamp option, and only
send didChange (writing the buffer to a tempoary file) if this
does not match the current timestamp.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -3,10 +3,18 @@ decl -hidden completions lsp_completions
 decl -hidden range-specs lsp_errors
 
 decl -hidden range-specs cquery_semhl
+decl -hidden str lsp_draft
+decl -hidden int lsp_timestamp -1
 
-def lsp-did-change %{
-    decl -hidden str lsp_draft %sh{ mktemp }
-    eval -no-hooks write %opt{lsp_draft}
+def lsp-did-change %{ try %{
+    %%sh{
+        if [ $kak_opt_lsp_timestamp -eq $kak_timestamp ]; then
+            echo "fail"
+        else
+            draft=$(mktemp)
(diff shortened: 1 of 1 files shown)


ul:
Thanks for contribution! Now I learned trick with `try` + `echo "fail"` =)

rc/lsp.kak
```diff
@@ -3,10 +3,18 @@ decl -hidden completions lsp_completions
...
 
 decl -hidden range-specs cquery_semhl
+decl -hidden str lsp_draft
+decl -hidden int lsp_timestamp -1
 
-def lsp-did-change %{
-    decl -hidden str lsp_draft %sh{ mktemp }
-    eval -no-hooks write %opt{lsp_draft}
+def lsp-did-change %{ try %{
+    %%sh{
```
mawww: Just realised I developped that on the explicit-reparsing branch, so this `%%` should be replaced with a `%`. I plan to merge that branch, but as it will break every scripts, I want to bundle that push with a general cleanup of command parsing in case we need to break more things (avoiding to have all script writers update their scripts twice).

ul: Strange enough it works fine on master (or maybe I'm missing something). Should we replace `%%` with `%` now or just replace `%` with `%%` in other places after branch merge?

mawww: I suspect it fails silently due to the try block... `%%sh{...}` is going to be parsed as a string delimited by `%` on master (the same way `%~...~` is `~` delimited, on current master `%%...%` is `%` delimited). I suggest to remove that extra `%` for now, and reintroduce it once explicit-reparsing is enabled.

ul: Indeed it fails, I was unfocused when checking it last night. Fixed it now. Thanks for heads up!

PR #28, opened 2018-05-15
Only errors will be highlighted with the Error face

BurNiinTRee:
Currently it's difficult to tell errors apart from warning without moving the cursor over the problematic section. This changes the face used to highlight diagnostics other than actual warnings to Information.


ul:
Nice improvement, thank you for contribution!

PR #30, opened 2018-05-15
Added a service file for a systemd user service

BurNiinTRee:


kak-lsp.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Kak LSP daemon
+
+[Service]
+ExecStart=/home/lars/.local/bin/kak-lsp
```
ul: Is it possible to use `~` or `$HOME` or some equivalent here? If not, we probably should leave a note in README about extra step of editing this file.

BurNiinTRee:
There you go

PR #35, opened 2018-05-19
use shell arihmetic expansion instead of 'expr'

occivink:
Arithmetic expansion is posix, and slightly slimper/cheaper than calling `expr`


ul:
Neat, thanks for contribution!

PR #42, opened 2018-05-27
Reference AUR package

GordStephen:
I've created an AUR package to automate/structure installation for ArchLinux users here: https://aur.archlinux.org/packages/kak-lsp-git/ - this adds a reference to it in the README.

Regarding the package itself, happy to incorporate any suggestions / changes to the install process. You can see the PKGBUILD file [here](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=kak-lsp-git).

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -95,7 +95,7 @@ systemctl --user enable --now kak-lsp
 
 == Language servers
 
-`kak-lsp` doesn't manage installation of language servers, please install them by yourself for the languages you plan to use `kak-lsp` with. Please consult  https://github.com/ul/kak-lsp/wiki/How-to-install-servers[How to install servers] wiki page for quick installation of language servers supported by `kak-lsp` out of the box.
+`kak-lsp` doesn't manage installation of language servers, please install them by yourself for the languages you plan to use `kak-lsp` with. Please consult the https://github.com/ul/kak-lsp/wiki/How-to-install-servers[How to install servers] wiki page for quick installation of language servers supported by `kak-lsp` out of the box.
 
 == Usage
 
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -68,6 +68,8 @@ mv kak-lsp.toml ~/.config/kak-lsp/
 
 === From the source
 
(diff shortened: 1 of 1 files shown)


ul:
Thank you for contribution!

PKGBUILD looks fine to me, the only thing which might need to be fixed is [install -Dt "${pkgdir}/etc/" kak-lsp.toml](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=kak-lsp-git#n33) — unfortunately at the moment kak-lsp doesn't try to read from that location and after installation it will be not immediately usable. User need to copy config manually into `~/.config/kak-lsp/`. I think it might be worth mention in package description and/or in installation message. I raised #44 to fix this problem (actually I will do it right now because it's easy to do and result should be quite beneficial)

GordStephen:
Yeah, there's a post-install message about that (as well as activating the systemd unit) - https://aur.archlinux.org/cgit/aur.git/tree/kak-lsp-git.install?h=kak-lsp-git

ul:
Ah, I missed it, thanks for pointing out correct place. Anyway, with the latest master copying file should be required only for customization (which I guess virtually everyone does because default config is far from being complete or universal).

Foxboron:
@GordStephen Yo! Thanks for the AUR submission.

A few things to point out in the PKGBUILD:
-  The `v` from the version should be parsed out.  Mentioned in the [VCS package guidelines](https://wiki.archlinux.org/index.php/VCS_package_guidelines#Git)
- The example you have put into `/etc/kak-lsp.toml` should be removed and placed into `/usr/share/$pkgname/examples/kak-lsp.toml`
- I also think you should be using `install -Dm644` and such when installing files.

GordStephen:
Thanks for the feedback! I'll push a new release.

GordStephen:
@Foxboron I [incorporated](https://aur.archlinux.org/cgit/aur.git/commit/?h=kak-lsp-git&id=d107590775937859750099ae5b68ae5571123433) those changes, if you have any other suggestions let me know.

Foxboron:
They look great! Awesome work :)

PR #49, opened 2018-05-30
add links to LSP, Kakoune and Rust in README

Delapouite:
Hello

I followed the way you present the project on reddit annoucements.
It may help user not already familiar with all those terms.


ul:
Thank you for contribution!

Issue #34, opened 2018-05-18
Optimize editor_references


ul:
Group locations by file and scan each file once.

Pantoofle:
Sorry for the notification spam. My bad. Failed the formating phase and had to create a new pull request

ul:
No worries =) Let's polish PR just a little bit more and get it merged.

PR #60, opened 2018-06-08
Apply Issue #34 optimization

Pantoofle:
Now the locations are grouped by filename and then sorted by line number.
This way, we only scan files once and load only one buffer.


Pantoofle:
Needed one more commit to uniform the tabs/spaces and pull the actual master branch

PR #61, opened 2018-06-08
Apply Issue #34 optimization

Pantoofle:
Now the locations are grouped by filename and then sorted by line number.
This way, we only scan files once and load only one buffer.


src/language_features/references.rs
```diff
@@ -43,46 +44,73 @@ pub fn editor_references(
         ReferencesResponse::Array(locations) => Some(locations),
         ReferencesResponse::None => None,
     } {
+        // Sort locations by (filename, line)
+        let mut locations = locations.to_vec();
```
ul: I think this could be replaced by just putting `mut` into pattern match: `if let Some(mut locations) = match result {`

src/language_features/references.rs
```diff
@@ -43,46 +44,73 @@ pub fn editor_references(
...
-                let filename = path.to_str().unwrap();
-                let file = File::open(filename);
+            .group_by(|location|{
+              location.uri.to_file_path()
+            })
+            .into_iter()
+            .map(|(filename, group)| {
+                let filename = filename.unwrap();
+                let name = filename.to_str().unwrap();
+                let file = File::open(name);
```
ul: `File::open` can take path as well, no need to convert it to string. Also we could lift making relative path string from line 87 because it will be the same each iteration of loop. I imagine something like that:

```
                let filename = filename.unwrap();
                let file = File::open(&filename);
                let name = filename
                    .strip_prefix(&ctx.root_path)
                    .ok()
                    .and_then(|p| Some(p.to_str().unwrap()))
                    .or_else(|| filename.to_str())
                    .unwrap();
```

src/language_features/references.rs
Pantoofle: Okay, I'll adapt this

src/language_features/references.rs
Pantoofle: Oh yeah, of course. I blindly copy/pasted this part. I'll adapt it

PR #65, opened 2018-06-17
Fix completion of dollar prefixed variables

JJK96:
While using a php language server (which I also added to the wiki). I noticed that completion of variables prefixed with a `$` results in something like `$\$variable`.

This is the result of two things in the code.

# 1 

In rc/kak.lsp, the regex selecting the word that is to be replaced by the completion only takes \w characters. 
I changed it so that it also selects the leading `$`.
This results in variables completed to `\$variable`, which is a bit better.

# 2

In src/language_features/completion.rs the characters `[:|$]` are escaped.
I do not see a problem with removing this escaping, so I would gladly see pointed out why this escaping is necessary.

For now, disabling the escaping for the return entry fixes the issue.

### DIFF
--- a/src/language_features/completion.rs
+++ b/src/language_features/completion.rs
@@ -41,7 +41,7 @@ pub fn editor_completion(
         CompletionResponse::Array(items) => items,
         CompletionResponse::List(list) => list.items,
     };
-    let re = Regex::new(r"(?P<c>[:|$])").unwrap();
+    let re = Regex::new(r"(?P<c>[:|])").unwrap();
     let maxlen = items.iter().map(|x| x.label.len()).max().unwrap_or(0);
 
     let items = items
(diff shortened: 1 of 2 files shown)


src/language_features/completion.rs
```diff
@@ -66,9 +66,9 @@ pub fn editor_completion(
             }
             format!(
                 "{}|{}|{}",
-                re.replace_all(&x.insert_text.unwrap_or(x.label), r"\$c"),
+                x.insert_text.unwrap_or(x.label),
```
ul: I think we still want to escape `:` and `|` because they are delimiters in completions list (see https://github.com/mawww/kakoune/blob/master/doc/interfacing.asciidoc#completion-candidates). Let's just remove `$` from regex on line 44. I'm not sure why it's there, looks like I copy-pasted it from somewhere though from quick look to Kakoune sources `$` doesn't have any special meaning in completions list.

JJK96: I removed the $ from the regex.

PR #70, opened 2018-07-05
Fix command syntax after breaking changes in Kakoune

mawww:


ul:
Sorry for delay with it, I've got a bit busy these days and I want to migrate to the latest Kakoune by myself first before merging this PR.

PR #71, opened 2018-07-07
Update cquery to newest breaking kakoune changes

topisani:
Recent kakoune changes required some updates to the list format and `%sh` blocks


SolitudeSF:
#70

topisani:
hah, should probably have checked here first - mawww's is a lot more thorough than mine

PR #72, opened 2018-07-10
Add bash language server

maximbaz:
The "go to definition" and completions are useful even for zsh, so until there is a dedicated zsh language server, let's keep this enabled for zsh as well.


ul:
Thanks for contribution!

PR #79, opened 2018-07-27
Add line-flags for diagnostics

topisani:
Ive been missing this for some time, its really useful when scrolling through a file looking for errors.

Currently theres no way to set individual faces for each line spec, so errors and warnings both use the same face `LineFlagErrors´. However, the characters are customizable with the options `lsp_diagnostics_line_{error,warning}_sign`. It defaults to `*` for errors and `!` for warnings. Then the user can feel free to use whatever fancy unincode character they want.

I dont really know anything about rust, so i may have duplicated a little too much code, but it works, and is probably easy to clean up

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -1,6 +1,8 @@
 # faces used by inline diagnostics
 set-face global DiagnosticError red
 set-face global DiagnosticWarning yellow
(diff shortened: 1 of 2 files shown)


ul:
This is great UX improvement, thank you!

PR #84, opened 2018-08-05
Correct behaviour when formatting change is same line and character

stoand:
(Tested with Typescript)

Before:
`const j= 1234;`  
formats to
`const    1234;`

Now:
`const j= 1234;`
formats to
`const j = 1234;`


ul:
Thank you for contribution!

PR #93, opened 2018-10-03
add dart to kak-lsp.toml

tskinn:


ul:
Nice to see more servers supported out of the box, thank you for contribution!

Issue #97, opened 2018-10-07
[Feature request] next- and previous error


JJK96:
Currently I find myself scrolling through the file to find where the error occurred, especially if I have edited in multiple places, and the error is not currently in my view. 

I think it would be nice to be able to jump to an error quickly.

It does not seem too hard to do since we have the lsp-diagnostics buffer which can just be parsed for the line and column of the error.

I will probably try implementing it myself soon.

Let me know your opinions

ul:
It's easy to put a list of jump positions into an option when publishing diagnostics for the buffer. I think using that list would be easier than scanning `*diagnostics*` buffer. Do you want me to do Rust part, and you'll then implement Kakoune/shell part? (IIRC you are not interested much in Rust, but I may be wrong).

JJK96:
Sure! I'll take a look at it soon.

JJK96:
How about using the `lsp_errors` option as the jump list?

ul:
Should be totally fine if it's convenient enough to parse. Idea with introducing another option was to ease Kakoune/shell side work, but not to introduce extra entity is generally better, of course.

JJK96:
I think it should be easy to parse, since it is basically a list of coordinates including labels for the kind of diagnostic

PR #102, opened 2018-10-10
Find next or previous error

JJK96:
Implements #97

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -438,11 +438,11 @@ Jump to the next or previous diagnostic error" %{
     evaluate-commands %sh{
         previous=false
         errorCompare="DiagnosticError"
-        if [[ $1 == "--previous" ]]; then
+        if [ $1 = "--previous" ]; then
             previous=true
             shift
         fi
-        if [[ $1 == "--include-warnings" ]]; then
+        if [ $1 = "--include-warnings" ]; then
             errorCompare="Diagnostic"
         fi
         #expand quoting, stores option in $@
@@ -452,19 +452,19 @@ Jump to the next or previous diagnostic error" %{
         current=""
         prev=""
         selection=""
-        for e in $@; do
-            if [[ $e == *"$errorCompare"* ]]; then
+        for e in "$@"; do
+            if [ -z "${e##*${errorCompare}*}" ]; then # e contains errorCompare
                 e=${e%,*}
                 line=${e%.*}
                 column=${e#*.}
-                if [[ $line -eq $kak_cursor_line && $column -eq $kak_cursor_column ]]; then
+                if [ $line -eq $kak_cursor_line ] && [ $column -eq $kak_cursor_column ]; then
                     continue #do not return the current location
                 fi
                 current="$line $column"
                 if [ -z "$first" ]; then
                     first="$current"
                 fi
-                if [[ $line -gt $kak_cursor_line || (($line -eq $kak_cursor_line) && ($column -gt $kak_cursor_column)) ]]; then
+                if [ $line -gt $kak_cursor_line ] || { [ $line -eq $kak_cursor_line ] && [ $column -gt $kak_cursor_column ]; }; then
                     #after the cursor
                     if $previous; then
                         selection="$prev"
@@ -491,9 +491,9 @@ Jump to the next or previous diagnostic error" %{
                 selection="$first"
             fi
         fi    
-        echo "edit $kak_buffile $selection"
-        }
+        printf "edit %b %b" "$kak_buffile" "$selection"
     }
+}
 
 def lsp-auto-hover-enable -docstring "Enable auto-requesting hover info for current position" %{
     hook -group lsp-auto-hover global NormalIdle .* %{
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -433,6 +433,68 @@ def lsp-diagnostic-lines-disable -docstring "Disable diagnostics line flags"  %{
     remove-highlighter global/lsp_error_lines
 }
 
+def lsp-find-error -params 0..2 -docstring "lsp-find-error [--previous] [--include-warnings]
+Jump to the next or previous diagnostic error" %{
+    evaluate-commands %sh{
+        previous=false
+        errorCompare="DiagnosticError"
+        if [[ $1 == "--previous" ]]; then
+            previous=true
+            shift
+        fi
+        if [[ $1 == "--include-warnings" ]]; then
+            errorCompare="Diagnostic"
+        fi
+        #expand quoting, stores option in $@
+        eval "set -- ${kak_opt_lsp_errors}"
+
+        first=""
+        current=""
+        prev=""
+        selection=""
+        for e in $@; do
+            if [[ $e == *"$errorCompare"* ]]; then
+                e=${e%,*}
+                line=${e%.*}
+                column=${e#*.}
+                if [[ $line -eq $kak_cursor_line && $column -eq $kak_cursor_column ]]; then
+                    continue #do not return the current location
+                fi
+                current="$line $column"
+                if [ -z "$first" ]; then
+                    first="$current"
+                fi
+                if [[ $line -gt $kak_cursor_line || (($line -eq $kak_cursor_line) && ($column -gt $kak_cursor_column)) ]]; then
+                    #after the cursor
+                    if $previous; then
+                        selection="$prev"
+                    else 
+                        selection="$current"
+                    fi
+                    if [ ! -z "$selection" ]; then
+                        # if a selection is found
+                        break
+                    fi
+                else
+                    prev="$current"
+                fi
+            fi
+        done
+        if [ -z "$first" ]; then
+            # if nothing found
+            echo "echo -markup '{Error}No errors found'" 
+        fi
+        if [ -z "$selection" ]; then #if nothing found past the cursor
+            if $previous; then
+                selection="$current"
+            else 
+                selection="$first"
+            fi
+        fi    
+        echo "edit $kak_buffile $selection"
+        }
+    }
+
 def lsp-auto-hover-enable -docstring "Enable auto-requesting hover info for current position" %{
     hook -group lsp-auto-hover global NormalIdle .* %{
         lsp-hover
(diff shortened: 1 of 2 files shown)


Delapouite:
Thanks JJK96. I think this PR needs to be rebased or squashed.

JJK96:
Done

ul:
Thank you for implementing this! Unfortunately, I'm not good at shell scripting to do thorough review, but I tested it and it works well.

Few improvements which would be nice to have as a part of this feature:

1. Fix typo in the docstring: "pervious".
2. Add command description to the README.
3. Add command as a subcommand of `lsp` command.
4. Add bindings for both next and previous error to `lsp` user mode.

Last two things are still not documented, but good to keep them up-to-date now, and I'm going to document them for the next release.

rc/lsp.kak
```diff
@@ -393,6 +393,68 @@ def lsp-diagnostic-lines-disable -docstring "Disable diagnostics line flags"  %{
     remove-highlighter global/lsp_error_lines
 }
 
+def lsp-find-error -params 0..2 -docstring "lsp-find-error [--pervious] [--include-warnings]
+Jump to the next or previous diagnostic error" %{
+    evaluate-commands %sh{
+        previous=false
+        errorCompare="DiagnosticError"
+        if [[ $1 == "--previous" ]]; then
```
maximbaz: I have a feeling this isn't POSIX-compatible so won't work with `dash`, because https://github.com/mawww/kakoune/issues/2152 is being ignored using `dash` instead of `bash` is the only way to achieve reasonable startup performance.

JJK96:
> Thank you for implementing this! Unfortunately, I'm not good at shell scripting to do thorough review, but I tested it and it works well.
> 
> Few improvements which would be nice to have as a part of this feature:
> 
>     1. Fix typo in the docstring: "pervious".
> 
>     2. Add command description to the README.
> 
>     3. Add command as a subcommand of `lsp` command.
> 
>     4. Add bindings for both next and previous error to `lsp` user mode.
> 
> 
> Last two things are still not documented, but good to keep them up-to-date now, and I'm going to document them for the next release.



> Thank you for implementing this! Unfortunately, I'm not good at shell scripting to do thorough review, but I tested it and it works well.
> 
> Few improvements which would be nice to have as a part of this feature:
> 
>     1. Fix typo in the docstring: "pervious".
> 
>     2. Add command description to the README.
> 
>     3. Add command as a subcommand of `lsp` command.
> 
>     4. Add bindings for both next and previous error to `lsp` user mode.
> 
> 
> Last two things are still not documented, but good to keep them up-to-date now, and I'm going to document them for the next release.

Fixed these things

rc/lsp.kak
JJK96: I used shellcheck to find non-posix compliant statements, so this should work in dash now too

skrps: It still isn't POSIX-compliant.

JJK96: Do you have any specific points?

skrps:
I've made a thorough review about POSIX compliance and it just disappeared LOL

I remade all the comments.

rc/lsp.kak
skrps: In a POSIX shell, the test section of an if condition is not a builtin; instead, it uses `test(1)` in order to compute `true` or `false`.  In this case, you should do:

    [ "$1" = "--previous" ]

JJK96:
Probably your git wasn't fully up to date

rc/lsp.kak
```diff
@@ -452,19 +452,19 @@ Jump to the next or previous diagnostic error" %{
...
-                if [[ $line -eq $kak_cursor_line && $column -eq $kak_cursor_column ]]; then
+                if [ "$line" -eq "$kak_cursor_line" ] && [ "$column" -eq "$kak_cursor_column" ]; then
                     continue #do not return the current location
                 fi
                 current="$line $column"
                 if [ -z "$first" ]; then
                     first="$current"
                 fi
-                if [[ $line -gt $kak_cursor_line || (($line -eq $kak_cursor_line) && ($column -gt $kak_cursor_column)) ]]; then
+                if [ "$line" -gt "$kak_cursor_line" ] || { [ "$line" -eq "$kak_cursor_line" ] && [ "$column" -gt "$kak_cursor_column" ]; }; then
```
skrps: You can use parentheses instead of spawning a subshell.  From the [spec](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html):

> The following form can be used instead:
>
>     ( test expr1 && test expr2 ) || test expr3

rc/lsp.kak
```diff
@@ -452,19 +452,19 @@ Jump to the next or previous diagnostic error" %{
...
         selection=""
-        for e in $@; do
-            if [[ $e == *"$errorCompare"* ]]; then
+        for e in "$@"; do
+            if [ "${e#*$errorCompare}" != "$e" ]; then # e contains errorCompare
                 e=${e%,*}
                 line=${e%.*}
                 column=${e#*.}
-                if [[ $line -eq $kak_cursor_line && $column -eq $kak_cursor_column ]]; then
+                if [ "$line" -eq "$kak_cursor_line" ] && [ "$column" -eq "$kak_cursor_column" ]; then
```
skrps: Quotes should not be needed.

skrps:
Maybe...

It's because of those things that I find more and more fond of email-based patch reviews.

rc/lsp.kak
JJK96: As far as I can tell `()` spawns a subshell while `{}` does not https://askubuntu.com/questions/606378/when-to-use-vs-in-bash/606386

JJK96:
@skrps thanks for the detailed code review

rc/lsp.kak
skrps: Well, less characters...  Also, quotes should only be needed for strings containing whitespaces; AFAIK, the variables involved here don't have whitespaces, so quotes shouldn't be needed.

skrps: Remember: all variables are strings in a POSIX shell.  One shouldn't need to quote them as one does in C, Go or other programming languages, *unless* those strings carry some special symbol that could (or could not, depends of your intent) be erroneously interpreted by the shell.

rc/lsp.kak
JJK96: According to [spec](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html) 
```
test "$1" -a "$2"

should be written as:

test "$1" && test "$2"
```

rc/lsp.kak
```diff
@@ -433,6 +433,68 @@ def lsp-diagnostic-lines-disable -docstring "Disable diagnostics line flags"  %{
...
+        first=""
+        current=""
+        prev=""
+        selection=""
+        for e in $@; do
+            if [[ $e == *"$errorCompare"* ]]; then
+                e=${e%,*}
+                line=${e%.*}
+                column=${e#*.}
+                if [[ $line -eq $kak_cursor_line && $column -eq $kak_cursor_column ]]; then
```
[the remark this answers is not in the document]
JJK96: According to [spec](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html) 
```
test "$1" -a "$2"

should be written as:

test "$1" && test "$2"
```

rc/lsp.kak
skrps: > As far as I can tell `()` spawns a subshell while `{}` does not https://askubuntu.com/questions/606378/when-to-use-vs-in-bash/606386

Oh yeah, you're right here.

ul:
Big thanks to everyone involved in the review! Is PR ready to merge? What's left otherwise (i.e. `echo` -> `printf` replacement)?

Btw, I'm using `echo` in few other places in `lsp.kak`, should it be replaced? And in general, I'd be grateful for `lsp.kak` review, feel free to raise issues/PRs if you notice any non-portable or non-safe stuff.

JJK96:
Some things pointed out by @skrps still need to be applied, i.e. unnecessary quoting, printf instead of echo. I can do that tomorrow.

skrps:
> I'm using `echo` in few other places in `lsp.kak`, should it be replaced?

Depends on what you're using variables when calling `echo`.

I can review the shell script portions of `lsp.kak`; I should raise a PR soon if need be.

rc/lsp.kak
```diff
@@ -452,19 +452,19 @@ Jump to the next or previous diagnostic error" %{
         current=""
         prev=""
         selection=""
-        for e in $@; do
-            if [[ $e == *"$errorCompare"* ]]; then
+        for e in "$@"; do
+            if [ "${e#*$errorCompare}" != "$e" ]; then # e contains errorCompare
```
skrps: Just occurred to me that you can also do:

    [ -z "${e##*${errorCompare}*}" ]

which generates the same result and it's more readable, at least for me: if `e` contains the content of `errorCompare`, the parameter expansion will return an empty string; so you test if the string is empty and ta-da!

Could sound nitpicky, but this version keeps `e` from being expanded twice, compared to the original version.

rc/lsp.kak
maximbaz: > I used shellcheck to find non-posix compliant statements, so this should work in dash now too

@JJK96 I recommend `checkbashism` for this purpose. For example, for the line 

```sh
if [[ "$1" == "--previous" ]]; then
```

It outputs the following:

```
$ checkbashisms tmp.sh
possible bashism in tmp.sh line 5 (alternative test command ([[ foo ]] should be [ foo ])):
if [[ "$1" == "--previous" ]]; then
possible bashism in tmp.sh line 5 (should be 'b = a'):
if [[ "$1" == "--previous" ]]; then
```

rc/lsp.kak
JJK96: You're right, I adapted it

rc/lsp.kak
```diff
@@ -433,6 +433,68 @@ def lsp-diagnostic-lines-disable -docstring "Disable diagnostics line flags"  %{
...
+            echo "echo -markup '{Error}No errors found'" 
+        fi
+        if [ -z "$selection" ]; then #if nothing found past the cursor
+            if $previous; then
+                selection="$current"
+            else 
+                selection="$first"
+            fi
+        fi    
+        echo "edit $kak_buffile $selection"
```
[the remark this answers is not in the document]
JJK96: Adapted

PR #103, opened 2018-10-11
Send initialized notification

matklad:
According to the spec, the client must send initialized notification
before sending any other requests or notifications.

https://microsoft.github.io/language-server-protocol/specification#initialized

>The initialized notification is sent from the client to the server after the client received the result of the initialize request but before the client is sending any other request or notification to the server. The server can use the initialized notification for example to dynamically register capabilities. The initialized notification may only be sent once.

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -472,6 +472,7 @@ fn dispatch_server_response(
                     .expect("Failed to parse initialized response")
                     .capabilities,
             );
+            ctx.notify(notification::Initialized::METHOD.into(), InitializedParams {});
             let mut requests = Vec::with_capacity(ctx.pending_requests.len());
             for msg in ctx.pending_requests.drain(..) {
                 requests.push(msg);


ul:
Good catch, thanks for contribution!

ul:
Hey @matklad I know you are pretty experienced in Rust, am I allowed to ask you for code review of kak-lsp if you have time and interest in it? I'm especially unsatisfied on how I manage threads so it's hard to organize clean synchronized exit, but I guess there is a lot of other not-so-well-designed things.

matklad:
Sure, will take a look!

matklad:
BTW, I am a huge fan of maintaining a short ARCHITECTURE.md document which describes a high-level architecture of the project, which some pointers to code. ([example](https://github.com/rust-analyzer/rust-analyzer/blob/010096d3994a7329bb8415ac5921f2e8005daaec/ARCHITECTURE.md)). If you could draft such a document in, say, one hour, that would help me with diving into the code quicker :)

ul:
Great idea, thank you! I'll ping you when file will be ready then (hope to have some time to work on it this Sunday).

PR #106, opened 2018-10-11
Simplify poisoning pattern

matklad:
I took a quick peek at the code and I think I've identified one of the major sources of complexity here: the `posoned_tx_rx` pattern. It is used to cancel stuff all over the place. However in Rust we have a wonderful tool for cancellation: RAII (I've even [blogged](https://matklad.github.io/2018/03/02/stopping-a-rust-worker.html) about it). 

The main idea is that, instead of doing `select!(receiver, poison)`, we just do `for msg in receiver` and treat a closed receiver as a shutdown signal. A cool thing about this pattern is that it automatically works when the sender is dropped due to panic. 

I've applied it only in one instance, but it should be applicable to other uses of poisoned channels as well. 

I am also not sure that this particular patch actually works: I don't have a good understanding of interactions between various actors, however, I am pretty sure this patch could be tweaked to work.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -11,7 +11,7 @@ pub struct Context {
     pub diagnostics: FnvHashMap<String, Vec<Diagnostic>>,
     pub editor_tx: Sender<EditorResponse>,
     pub lang_srv_poison_tx: Sender<()>,
-    pub lang_srv_tx: Sender<ServerMessage>,
+    pub lang_srv_tx: Option<Sender<ServerMessage>>,
     pub language_id: String,
     pub pending_requests: Vec<EditorRequest>,
     pub request_counter: u64,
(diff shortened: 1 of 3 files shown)


src/context.rs
```diff
@@ -63,8 +63,9 @@ impl Context {
             method,
             params: Some(params.unwrap()),
         };
-        self.lang_srv_tx
-            .send(ServerMessage::Request(Call::MethodCall(call)));
+        if let Some(tx) = &self.lang_srv_tx {
+            tx.send(ServerMessage::Request(Call::MethodCall(call)));
+        }
```
matklad: Note that this `if let` is suspicious: what should we do if the  `lang_srv_tx` is `None`? Note, however, that the very same problem was here from the start: it could have been the case that the other side of the channel was poisoned, making this message silently lost. This approach just make this subtlety visible in code.

matklad:
Couple of random thoughts:

* its important to internalize that crossbeam_channel, unlike std::mpsc, is **strictly** unidirectional. That is, dropping the sender notifies the receiver. However, dropping the receiver has no effect on the sender.

* Because dropping a sender is itself an event you can listen too, a useful pattern is `Sender<Void>`  (where `enum Void {}` is an uninhabited type) which allows you to check if anything is alive (see here for an examle of this: https://github.com/rust-lang-nursery/rls/pull/923).

ul:
It works well, thank you! I will try to digest it and blog post and apply to other places.

PR #109, opened 2018-10-14
simplify controllers lifecycle; close #108

### LINKED ISSUE #108: Simplifying controllers lifecycle
One problem with controller is that it spawns a lo of threads and shares `Context` via an `Arc`, which muddles ownership structure and makes drop-based cancellation harder.

I think it should be possible to confine Controller to a single thread, and simplify stuff in the process. Let me sketch a design!

First, let's introduce a `ControllerHandle` struct, which will be owned by the root event loop, the one that currently owns the `Controllers` `HashMap`. This loop needs to do three things:

* dispatch requests to controllers (we'll use `Sender<EditorRequest>` for that)
* be notified when controller dies (b/c, for example, the lang srv died). We'll use `Receiver<Void>` pattern for that
* reap stopped controllers. 

So, we get

```rust
struct ControllerHandle {
    sender: Sender<EditorRequest>,
    is_alive: Receiver<Void>,
    thread: JoinHandle<()>,
}
```

The first version of controllers loop might look like

```rust
let (is_alive_rx, is_alive_tx) = bounded::<Void>(0);
std::thread::spawn(|| {
    let alive = is_alive_tx;
    for request in editor_request_receiver {
        ...
    }
    drop(alive); // signal that we are done, will happen implicitly without `drop`
})
```

The main loop will then select over all `is_alive` (you can pass in an iterator to the `select!` macro), roughly like this:

```rust
let mut controllers: Vec<ControllerHandle> = ...;
let receivers = controllers.iter().map(|c| &c.is_alive);
select! {
    // The third argument to `recv` is optional and is assigned a
    // reference to the receiver the message was received from.
    recv(receivers, msg,lo from) => {
        assert!(msg.is_none()); // msg type is Void, so we only can get a closed event
        let controller_idx = controllers.iter().position(|c| c.is_alive == from).unwrap();
        let controller = controllers.swap_remove(controller_idx);
        controller.thread.join();
    }
}
```

Note that if we want to add an ability to explicitly stop the controller from the top-level event loop, we can change `sender: Sender<EditorRequest>` to `sender: Option<Sender<EditorRequest>>`. Stoping the controller would be just setting `sender` to None. Note that reaping of the controller would happen automatically, because `is_alive_tx` will get dropped. 

Now, the inner event loop of the controller can't be just `for request in editor_requests_rx`, because we need to deal with language server as well. But I think we can still use a single thread and loop and just use `select!` macro do dispatch between various events? Like this:

```rust
fn controller_loop(
    editor_requests: Receiver<EditorRequest>,
    srv_rx: Receiver<LangSrvMsg>,
    srv_tx: Sender<LangSrvMsg>,
    is_alive: Sender<Void>,
) {
    enum Event {
        EditorRequest(EditorRequest), // recived message from editor_requests
        EditorDied, // editor_requests is closed
        LangSrvRequest(LangSrvMsg),
        LangSrvDied,
    }
    let mut ctx = Context::new(....); // Note: shared state is mut on the stack, no need for arcs & mutexes
    loop {
          let event: Event = select!(...);
          match event {
                    Event::EditorDied => break,
                    Event::LangSrvDied => break, // log error maybe?
                    Event::EditorRequest(req) => ...,   
                    Event::LangSrvRequest(req) => ...,   
          }
    }
    // Note that we correctly signal termination in any of the cases:
    // - editor receviver is closed, that is, main loop has explicitly terminated this controller
    // - langsrv receiver has died, which means that LSP server unexpectedly did a belly up.
    // - this loop itself panicked during to a bug
    drop(is_alive);
}
```

ul:
@matklad have I applied you suggestions properly? I think I got the idea, but there are could be some subtle nuances which I could have missed. Though I tested changes manually and they seem to work correctly for the most common cases, and the only hole left is exiting by timeout (it's still a hard process::exit, I plan to fix it soon (and now it's much easier to do!), just want to do it together with pid/socket cleanup issue).

I was again surprised how well Rust compiler supports such refactors. Literally, I fearlessly moved big chunks of code around, fixed compiler errors, and it worked!

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -14,7 +14,7 @@ use types::*;
 use util::*;
 
 struct ControllerHandle {
-    sender: Sender<EditorRequest>,
+    sender: Option<Sender<EditorRequest>>,
     is_alive: Receiver<Void>,
     thread: JoinHandle<()>,
 }
@@ -107,7 +107,9 @@ pub fn start(config: &Config, initial_request: Option<&str>) {
 
                 if controllers.contains_key(&route) {
                     let controller = controllers.get(&route).unwrap();
-                    controller.sender.send(request);
+                    if let Some(sender) = controller.sender.as_ref() {
+                        sender.send(request);
+                    }
                 } else {
                     // because Kakoune triggers BufClose after KakEnd
                     // we don't want textDocument/didClose to spawn new controller
@@ -134,16 +136,16 @@ fn exit_editor_session(controllers: &mut Controllers, request: &EditorRequest) {
         "Editor session `{}` closed, shutting down associated language servers",
         request.meta.session
     );
-    // have to clone keys as controllers are mutated inside loop
-    for route in controllers.keys().cloned().collect::<Vec<_>>() {
+    for (route, controller) in controllers.iter_mut() {
         if route.session == request.meta.session {
-            // should be safe to unwrap because we are iterating controllers' keys
-            let controller = controllers.remove(&route).unwrap();
             info!("Exit {} in project {}", route.language, route.root);
             // to notify kak-lsp about editor session end we use the same `exit` notification as
             // used in LSP spec to notify language server to exit, thus we can just clone request
             // and pass it along
-            controller.sender.send(request.clone());
+            if let Some(sender) = controller.sender.as_ref() {
+                sender.send(request.clone());
+            }
+            controller.sender = None;
         }
     }
 }
@@ -161,11 +163,12 @@ fn stop_session(controllers: &mut Controllers) {
         params: toml::Value::Table(toml::value::Table::default()),
     };
     info!("Shutting down language servers and exiting");
-    for k in controllers.keys().cloned().collect::<Vec<_>>() {
-        // should be safe to unwrap because we are iterating controllers' keys
-        let controller = controllers.remove(&k).unwrap();
-        controller.sender.send(request.clone());
-        info!("Exit {} in project {}", k.language, k.root);
+    for (route, controller) in controllers.iter_mut() {
+        if let Some(sender) = controller.sender.as_ref() {
+            sender.send(request.clone());
+        }
+        controller.sender = None;
+        info!("Exit {} in project {}", route.language, route.root);
     }
     stderr().flush().unwrap();
     stdout().flush().unwrap();
@@ -196,7 +199,7 @@ fn spawn_controller(
 
     ControllerHandle {
         is_alive: is_alive_rx,
-        sender: controller_tx,
+        sender: Some(controller_tx),
         thread,
     }
 }
--- a/src/session.rs
+++ b/src/session.rs
@@ -1,0 +1,202 @@
+use controller;
+use crossbeam_channel::{bounded, Receiver, Sender};
+use editor_transport;
+use fnv::FnvHashMap;
+use languageserver_types::notification::Notification;
+use languageserver_types::*;
+use project_root::find_project_root;
+use std::io::{stderr, stdout, Write};
+use std::process;
+use std::thread::{self, JoinHandle};
+use std::time::Duration;
+use toml;
+use types::*;
+use util::*;
+
+struct ControllerHandle {
+    sender: Sender<EditorRequest>,
+    is_alive: Receiver<Void>,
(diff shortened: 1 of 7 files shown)


src/session.rs
```diff
@@ -0,0 +1,202 @@
...
+    // have to clone keys as controllers are mutated inside loop
+    for route in controllers.keys().cloned().collect::<Vec<_>>() {
+        if route.session == request.meta.session {
+            // should be safe to unwrap because we are iterating controllers' keys
+            let controller = controllers.remove(&route).unwrap();
+            info!("Exit {} in project {}", route.language, route.root);
+            // to notify kak-lsp about editor session end we use the same `exit` notification as
+            // used in LSP spec to notify language server to exit, thus we can just clone request
+            // and pass it along
+            controller.sender.send(request.clone());
```
matklad: To properly clean up this without timeouts, we need to drop `controller.sender` first, and then block on `controller.is_alive.recv()`. Now with the code written, I think that we probably should move `sender` out of `Controller`, so that they could be destroyed automatically?

Smth like this:


```rust
struct ControllerHandle {
    is_alive: Sender<Void>,
    thread: JoinHandle<()>,
}

let mut handles: Vec<ControllerHandle> = Vec::new();
let mut controllers: HashMap<Route, Sender<EditroRequest>> = HashMap::new();
'event_loop: loop {
    let mut dead_controller = None;
    select! {
        recv(handles.iter().map(|it| &it.is_alive), msg, from) => {
            assert!(msg.is_none());
            dead_controller = handles.iter().position(|it| it.is_alive == from)
        }
        recv(commmands_receiver, ...) => {
            let route = ...
            let (handle, sender) = spawn_controller(route);
            handles.push(handle);
            controllers.insert(route, sender);
         }
    }
    if let Some(idx) = dead_controller {
        handles.swap_remove(idx).thread.join()
    }
}

drop(controllers); // stopped controllers
for handle in handles {
    assert!(handle.is_alive().recv().is_none());
    handle.thread.join().
}

```

matklad: Note that this has a nice bonus: it supports multiple exiting controllers for a single route.

That is, we can start a controller for route `foo`, add it to the controllers map, then stop it, remove from the map, and, while it is still in `handles`, add another controller with the `foo` route.

matklad: Hm, nope, this won't work as is: when we remove a handle from `handles`, we need to remove the corresponding route as well, and my proposal does not allow to do that. Perhaps it's just best to keep the `sender` in Controller, but make it an `Option` to allow for clean-up.

src/session.rs
```diff
@@ -0,0 +1,202 @@
...
+    for k in controllers.keys().cloned().collect::<Vec<_>>() {
+        // should be safe to unwrap because we are iterating controllers' keys
+        let controller = controllers.remove(&k).unwrap();
+        controller.sender.send(request.clone());
+        info!("Exit {} in project {}", k.language, k.root);
+    }
+    stderr().flush().unwrap();
+    stdout().flush().unwrap();
+    thread::sleep(Duration::from_secs(1));
+    process::exit(0);
```
matklad: This is a dangerous pattern, calling `process::exit` won't unwind the stack or call destructors, and that is usually desirable. The safe way to use `process::exit` is as a sole call in `main`:

```rust
fn main() {
    let code = main_inner();
    ::std::process::exit(code);
}

fn main_inner() -> i32 {
     ...
}
```

Similarly, not joining a thread can also lead to some info loss. I think the current code has a "lost thread" problem when communicating with language server. We don't wait for this thread:

https://github.com/ul/kak-lsp/blob/9b36a30d595d411059546e4a019718ae1120fec7/src/language_server_transport.rs#L67-L73

That means that, during shutdown, we may kill all controllers and shut down the process *before* these writer threads have a change to write the `exit` notification for proper LSP shutdown. I try to use [structured concurency](https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/) pattern in similar cases: if I spawn a thread, I make sure that it joined on all paths by the parent thread.

ul: Good catch! I already wanted to eliminate `exit` use, but was unaware of language server writer thread problem. I'll fix it soon.

PR #115, opened 2018-10-16
Add support for ccls with a couple of custom features

topisani:
https://github.com/MaskRay/ccls has replaced cquery for me, and probably many others. It supports a few more nice features, and i implemented some of them in kak-lsp.

Once again (like with cquery), I'm not sure if this is the structure you want, and my rust is pretty terrible, so there might be some issues. Everything seems to work though.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -321,6 +321,99 @@ insertSpaces = %s
 ' "${kak_session}" "${kak_client}" "${kak_buffile}" "${kak_timestamp}" "${kak_opt_lsp_tab_size}" "${kak_opt_lsp_insert_spaces}" | ${kak_opt_lsp_cmd}) > /dev/null 2>&1 < /dev/null }
 }
 
+# CCLS Extension
+
+def ccls-navigate -docstring "Navigate C/C++/ObjectiveC file" -params 1 %{
+    lsp-did-change
+    nop %sh{ (printf '
+session   = "%s"
+client    = "%s"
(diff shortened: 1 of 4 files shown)


topisani:
Thanks for that, i somehow forgot about it again. nice work on kak-lsp btw, this is the reason why i can use my favorite editor with my favorite language ;)

ccls has some even more advanced features, like generating browsable call/inheritance trees. It's something ive considered implementing at some point, but it would probably require either a custom terminal UI or a fair amount of kak code. To some extend i feel like this goes outside the scope of kak-lsp? but at the same time, thats how it would be the easiest to implement.

An alternative could be allowing users to implement request handlers directly from kak, which would make support for extensions a lot easier - any thoughts on this?

ul:
Thanks for updating the wiki!

> To some extend i feel like this goes outside the scope of kak-lsp? but at the same time, thats how it would be the easiest to implement.

The primary goal of kak-lsp is to provide Kakoune with IDE-like features, and basic LSP was just the most effective investment, but any useful extensions could find their place in kak-lsp long before it would start to look bloated. Of course, I prioritize implementing core LSP features for myself, but I'm more than happy to accept contribution of extensions.

> An alternative could be allowing users to implement request handlers directly from kak, which would make support for extensions a lot easier - any thoughts on this?

I started movement in that direction by introducing `lsp-show-*` and similar commands in `lsp.kak`. If you see how to improve it even further it would be great.

PR #123, opened 2018-10-30
Fix error due to experimental access to crates through prelude

JJK96:
```
error[E0658]: access to extern crates through prelude is experimental (see issue #44660)
  --> src/language_features/ccls.rs:15:12
   |
15 |     if let toml::Value::Table(ref mut table) = params {
   |            ^^^^

error[E0658]: access to extern crates through prelude is experimental (see issue #44660)
  --> src/language_features/ccls.rs:18:13
   |
18 |             toml::Value::try_from(TextDocumentIdentifier {
   |             ^^^^

error[E0658]: access to extern crates through prelude is experimental (see issue #44660)
  --> src/language_features/ccls.rs:50:18
   |
50 |     let result = serde_json::from_value(result).expect("Failed to parse definition response");
   |                  ^^^^^^^^^^
```

### DIFF
--- a/src/language_features/ccls.rs
+++ b/src/language_features/ccls.rs
@@ -2,6 +2,8 @@ use context::*;
 use itertools::Itertools;
 use jsonrpc_core::{Params, Value};
 use languageserver_types::{NumberOrString, Position, Range, TextDocumentIdentifier};
+use toml;
+use serde_json;
 use serde;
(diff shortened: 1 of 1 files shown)


JJK96:
Same problem as in that issue, I'm currently updating to 1.30.

maximbaz:
Because it's so simple 😛

PR #132, opened 2018-11-21
Daemonize in a current working directory

dpc:
It looks like `rls` does not work correctly if cwd was changed to `/`. I can hardly see downsides of deamonizing in the cwd, so I propose this as a patch, and I will fill an issue against rls.


ul:
Thanks for contribution! Yes, that all sounds reasonable.

PR #138, opened 2018-12-01
Update install instructions for Arch Linux

maximbaz:
I added `kak-lsp` to the official package repository for Arch Linux. 

Hopefully other distros will package it too at some point, so I proactively created a whole new section `Package managers` 😄


ul:
Good stuff, thank you! Sorry for delay with merge.

PR #139, opened 2018-12-06
Make the lsp-rename-prompt prompt default to the current name.

### LINKED ISSUE #137: lsp-rename-prompt should default to the current name
Steps to reproduce:

 1. In a file with a supported language server, select a name
 2. `:lsp-rename-prompt`

Expected result:

  - A "New name" prompt with the current name filled in, so it can be edited

Actual result:

  - A "New name" prompt with no initial value, so to make a minor change I have to copy/paste the name from the buffer into the prompt.

Screwtapello:
Fixes #137.

I filed https://github.com/mawww/kakoune/issues/2618 about `prompt` mysteriously not working inside `eval -draft`.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -179,7 +179,12 @@ character = %d
 }
 
 def lsp-rename-prompt -docstring "Rename symbol under the main cursor (prompt for a new name)" %{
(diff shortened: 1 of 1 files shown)


ul:
Thank you!

Hmm, that's surprising behaviour of prompt, let's see what Maxime would say.

Issue #94, opened 2018-10-06
[Feature Request] Show amount of lsp-diagnostic errors in status bar


JJK96:
For me it would be beneficial to see at a glance if there are errors in the current file, instead of having to switch to the lsp-diagnostics buffer or scrolling through the file.

ul:
Binaries should be available soon at https://github.com/ul/kak-lsp/releases/tag/v5.5.0
Just use `lsp_diagnostic_count` option in your modeline and voilà

JJK96:
Thanks you very much
Such a quick response!

JJK96:
Currently `lsp_diagnostic_count` contains the number of errors + the number of warnings. I would like to show only the number of errors in my status bar, maybe it is an option to have two variables?

PR #142, opened 2018-12-13
Split `lsp_diagnostic_count` in errors and warnings (#94)

JJK96:


ul:
Thanks, it could be a useful distinction.

PR #147, opened 2018-12-19
Add Makefile

maximbaz:
I'd like to have a Makefile in the repo, being the source of truth for how to install the app.

This will allow me and everyone else use `$ make` to compile the binary and `$ make install` to install it.


ul:
That looks convenient, thanks for adding!

PR #149, opened 2018-12-23
Fix error when executing lsp-find-error without arguments

JJK96:
shell stderr: <<<
sh: line 3: [: =: unary operator expected
sh: line 7: [: =: unary operator expected
>>>


ul:
Good catch, thank you!

PR #150, opened 2018-12-26
Proper loading by plugin managers

### LINKED ISSUE #117: Proper structure of main script, for loading by plugin managers
Hello. I'm writing a [plugin manager](https://github.com/andreyorst/plug.kak), and currently I'm not able to use your plugin, if it is loaded by the plugin manager.

I've added `noload` option to workaround this issue, but I think that this isn't proper way of how plugin should work. I still want to be able to start lsp manually, I just don't want to avoid loading of main script, and load it via shell command.

My current setup is:

```kak
plug "ul/kak-lsp" "noload" %{
    hook global WinSetOption filetype=(c,cpp,rust) %{
        evaluate-commands %sh{ kak-lsp --kakoune -s $kak_session }
    }
}
```

Let me breakdown this for you:
`plug` - it's a command that does the following:
- checks if first parameter (which is a repo) exists at plugin installation path, and if it exists:
  - parses parameters, like `"noload"`, `"branch: ..."`, `"tag: ..."`;
  - loads all `.kak` files in the repo. If `"noload"` is specified this step is skipped;
  - loads `%{settings}` from the last parameter even if previous step was skipped.

So this configuration skips the plugin files itself, but sets the hooks that will call external program, that will load `lsp.kak` script. I think that this should be handled by `lsp.kak` and not by external program.  So user configuration can be like this:

```kak
plug "ul/kak-lsp" %{
    set-option global lsp_path_to_executable "no/more/symlinking"
    hook global WinSetOption filetype=(c,cpp,rust) %{
        lsp-start [optional parameters]
    }
}
```

So script could handle everything by itself, as a plugin should do. Final proposal workflow is:

- `source lsp.kak` either by plugin manager, or from user configuration, or from autoload folder;
- configure some options like path to kak-lsp inside kakoune's config;
- start lsp via generic command by entering it manually or by hook.

I understand that simply calling `kak-lsp` from shell and passing session id to it feels easier, but I think that this should be handled in the main script, and not in the binary.

andreyorst:
Ok. So I've read some code, and come to conclusion that there's no need in `plugin.kak` at all.

Your plugin is now loading perfectly fine with `plug`, and can be loaded in the same way without it by simply sorcing `lsp.kak`:

```kak
plug "andreyorst/kak-lsp" do %{
    cargo build --release
    cargo install
}
```
This PR fixes #117.

I've also did bit of refactoring by replacing all short commands with standard ones, like `decl` to `declare-option`. Are there any tests for kakscript so I could check if I didn't break anything?

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -87,8 +87,8 @@ plug "ul/kak-lsp" do %{
 }
 ----
 
-You can replace `cargo install` with `ln -s target/release/kak-lsp ~/.local/bin/`
-where `~/local/bin/` can be replaced to something in your `$PATH`.
+You can replace `cargo install` with `ln -sf target/release/kak-lsp ~/.local/bin/`
+where `~/.local/bin/` can be replaced to something in your `$PATH`.
 
 == Language servers
 
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -80,12 +80,12 @@ cp kak-lsp.toml ~/.config/kak-lsp/
 If you don't mind using plugin manager, you can install kak-lsp
 with https://github.com/andreyorst/plug.kak[plug.kak]. Add this code to your `kakrc`:
 
----
+----
 plug "ul/kak-lsp" do %{
     cargo build --release
     cargo install
 }
----
+----
 
 You can replace `cargo install` with `ln -s target/release/kak-lsp ~/.local/bin/`
 where `~/local/bin/` can be replaced to something in your `$PATH`.
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -23,8 +23,8 @@ tar xzvf kak-lsp-v6.0.0-x86_64-apple-darwin.tar.gz
 # replace `~/.local/bin/` with something on your `$PATH`
 mv kak-lsp ~/.local/bin/
 
-mkdir -p ~/.config/kak-lsp 
-mv kak-lsp.toml ~/.config/kak-lsp/ 
+mkdir -p ~/.config/kak-lsp
+mv kak-lsp.toml ~/.config/kak-lsp/
 ----
 
 ==== Linux
@@ -42,8 +42,8 @@ tar xzvf kak-lsp-v6.0.0-x86_64-unknown-linux-musl.tar.gz
 # replace `~/.local/bin/` with something on your `$PATH`
 mv kak-lsp ~/.local/bin/
 
-mkdir -p ~/.config/kak-lsp 
-mv kak-lsp.toml ~/.config/kak-lsp/ 
+mkdir -p ~/.config/kak-lsp
+mv kak-lsp.toml ~/.config/kak-lsp/
 ----
 
 ==== FreeBSD
@@ -55,8 +55,8 @@ tar xzvf kak-lsp-v6.0.0-x86_64-unknown-freebsd.tar.gz
 # replace `~/.local/bin/` with something on your `$PATH`
 mv kak-lsp ~/.local/bin/
 
-mkdir -p ~/.config/kak-lsp 
-mv kak-lsp.toml ~/.config/kak-lsp/ 
+mkdir -p ~/.config/kak-lsp
+mv kak-lsp.toml ~/.config/kak-lsp/
 ----
 
 === From the source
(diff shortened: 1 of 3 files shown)


README.asciidoc
```diff
@@ -70,11 +70,26 @@ cargo build --release
...
 
+=== With plug.kak
+
+If you don't mind using plugin manager, you can install kak-lsp
+with https://github.com/andreyorst/plug.kak[plug.kak]. Add this code to your `kakrc`:
+
+---
+plug "ul/kak-lsp" do %{
+    cargo build --release
+    cargo install
```
maximbaz: Consider using `make` (if not also `make install`), to let @ul have control of what is the proper way of compiling and installing the tool, to have room for future changes. For example, right now your command is not respecting `Cargo.lock`. There was some discussion about disabling LTO, maybe one day it would be possible to change from `$ make` to `$ make LTO=no`

andreyorst: make has problems. Rust already has a build system builtin, and it's amazing. Well I guess I'll wait for @ul to response. If he will say to use make, I'll surely do, but for now let it stay this way. If the `--locked` parameter is the only difference I don't really see the problem here. Sure, what you're saying about make, and ability to silently change command is true, I just don't see why it can't be done with cargo alone. I've thought that there were some cargo configuration file which should handle such things, no?

ul:
Thanks for the contribution! I glad to see that kak-lsp is friendlier to plug.kak now.

>  Are there any tests for kakscript so I could check if I didn't break anything?

Unfortunately, kak-lsp has a sad testing story at the moment, as is in "no automated tests at all for both Rust and Kakoune script code". Though it seems to be close to the threshold of complexity when tests become a vital investment.

Regarding the discussion about `make`:

1. Definitely, we are not going to use `make` as _a build system._ @andreyorst is right, Rust has its own, and it's versatile enough (we always can introduce `build.rs` and do whatever we want).
2. OTOH `make` seems to be a good choice as _a cross-platform task runner,_ and I totally agree with @maximbaz that we could neatly hide common tasks behind `make` targets.

In this particular case, I would be happy to bounce responsibility for the decision to @andreyorst as the author of plug.kak, because:

1. First things first, there is still no canonical way to install kak-lsp from the sources. If you install it from the sources, it implies that you know what you do and you want that flexibility: `cargo install` or `mv target/release/kak-lsp /somewhere/in/PATH` or `ln -s $(pwd)/target/release/kak-lsp /somewhere/in/PATH` or `install -Dm755 -t /somewhere/in/PATH target/release/kak-lsp` or whatever.
2. Arguably it's better to have a dedicated section in README and/or Makefile target curated by the author of plug.kak with the risk that sections could go outdated for a while than me trying to guess if I break plug.kak in a subtle way when I update shared target in Makefile.

TL;DR

* We use Makefile only as a task runner, and I feel it's okay in this role.
* It's up to @andreyorst to find out what works best with plug.kak. Any of discussed solutions doesn't do harm to other ways of kak-lsp usage.

README.asciidoc
```diff
@@ -70,11 +70,26 @@ cargo build --release
...
+
+----
+plug "ul/kak-lsp" do %{
+    cargo build --release
+    cargo install
+}
+----
+
+You can replace `cargo install` with `ln -s target/release/kak-lsp ~/.local/bin/`
+where `~/local/bin/` can be replaced to something in your `$PATH`.
```
ul: `.` seems to be missing: `~/local/bin/` -> `~/.local/bin/`

maximbaz:
With the introduction of Makefile I wanted it to become a canonical way of building and installing `kak-lsp`, so that there's one source of truth on which `cargo` command to use to compile the binary and which set of files need to be copied to consider `kak-lsp` fully "installed". So ideally README would just point to Makefile, saying that it reflects the officially supported way of building and installing the tool (and it might change in the future), but people are of course free to do whatever they want.

ul:
> With the introduction of Makefile I wanted it to become a canonical way of building and installing `kak-lsp`

I fully support this initiative, my only consideration is that it doesn't fit all use cases. E.g. plug.kak doesn't require you to copy lsp.kak anywhere, it sources it automatically.

maximbaz:
It would be nice though if plug.kak did respect Makefile, if it installed full package (license, example kak-lsp.toml and other files), even if it doesn't strictly use all the files.

maximbaz:
> That's a task for system package manager, not a plugin manager. Plugin manager works in term of editor, not a whole system.

To be honest, I fail to see the difference for end-users. Remember, we are only talking about Makefile as the "task runner", nobody is asking you to run `sudo make install`. Makefile simply contains the list of files that the author of the project considers to be a canonical installation bundle. 

It is perfectly fine if end-users look at the project and decide that they want to perform a partial installation, but you and me as creators of plugin manager and system package respectively are hiding installation details from users, and if the author of a tool expressed their intention for installation instructions in a form of Makefile, it is only fair to both author and our users to respect it.

----

Anyways, if you and @ul are both happy with this PR, go for it! I merely wanted to inform you that there's a Makefile now, and README.md potentially could be updated to say "compile with `make`, install with `make install`", and hide `cargo` commands inside the Makefile. And then the section about `plug.kak` would be the only place that recommends different set of installation steps. But if you are okay with it, so am I 😉

ul:
Let's go with the current approach and see. It doesn't look like it will cost much to change if it will prove itself wrong.

@maximbaz we can update README with `make` invocation instead of the direct call to cargo, at least for `build` target. For the installation part, we might want to ensure that users will not accidentally rewrite their custom configs when they use installation instruction for the upgrade (I guess `install(1)` does backups automatically but it worth to double check). What do you think?

maximbaz:
Agree 👍

PR #154, opened 2019-01-09
little more explicit configuration via plug.kak

andreyorst:


README.asciidoc
```diff
@@ -82,8 +82,13 @@ with https://github.com/andreyorst/plug.kak[plug.kak]. Add this code to your `ka
 
 ----
 plug "ul/kak-lsp" do %{
-    cargo build --release
-    cargo install
+    cargo build --release --locked
+    cargo install --force
+} config %{
+    hook global WinSetOption filetype=rust %{
+        lsp-start
```
ul: Normally explicit start is not required.

README.asciidoc
```diff
@@ -82,8 +82,13 @@ with https://github.com/andreyorst/plug.kak[plug.kak]. Add this code to your `ka
...
+        lsp-enable
```
ul: It's better to call `lsp-enable-window` in WinSetOption hook.

andreyorst: ok

andreyorst:
> If minimal config is working then it might worth just to keep it as is and link a wiki page with an elaborate example like https://github.com/ul/kak-lsp/wiki/Usage-with-plug.kak

Yeah, I think that README should just cover installation, and point to the wiki. Tho only change needed thought is to have `--force` option.

PR #157, opened 2019-01-13
 lsp-did-open: send text from kakoune to kak-lsp

### LINKED ISSUE #156: LSP does not work for newly created (not saved) files
Hey @ul,

Thanks for the plugin, it works really great. I found annoying bug, here's some details:

**Repro steps**:

* Create but not save a file using `kak` with enabled `kak-lsp`. Example: `kak foo.py`.
* Type `imp` and wait for completion.

**Expected behaviour**:

* A list of completion items from the language server.

**Actual behaviour**:

* An error is shown instead of completions.

---

I tracked down the problem to  `text_document_did_open` function. The problem is that kak-lsp program tries to read a file from filesystem in order to send its content to a language server, however, new (not saved) files cannot be found yet.

https://github.com/ul/kak-lsp/blob/e0bb7c16bcfaa05fefa166c52052ffb88d0e9080/src/text_sync.rs#L10-L32

I think `lsp.kak` should send a buffer's content to `kak-lsp` for `lsp-did-open` even the same way it sends the content for `lsp-did-change` event. Alternatively, we can send an empty "text" if file not found in `text_document_did_open` and trigger `lsp-did-change` right after `lsp-did-open` in `lsp.kak`.

ikalnytskyi:
kak-lsp is used to read a file's content from a filesystem on `didOpen` before sending it to a language server, and this is something that does not work in case we are dealing with newly created (unsaved) file. In order to fix that problem, we need to send a buffer context from kakoune directly to kak-lsp, similarly to `didChange` message.

Closes #156

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -283,7 +283,17 @@ method   = "capabilities"
 }
 
 define-command -hidden lsp-did-open %{
-    nop %sh{ (printf '
+    evaluate-commands %sh{
+        if [ $kak_opt_lsp_timestamp -eq $kak_timestamp ]; then
+            echo "fail"
+        else
+            echo "evaluate-commands -draft -no-hooks %{set-option buffer lsp_timestamp %val{timestamp}; execute-keys '%'; set-option buffer lsp_draft %val{selection}}"
+        fi
+    }
+    nop %sh{ (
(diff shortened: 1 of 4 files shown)


ul:
Thank you for the contribution! Works like a charm.

PR #168, opened 2019-02-18
Add missing space to mappings to discard history

andreyorst:
mappings that are executed from `goto` mode are displayed in history.


ul:
Thanks, it's a good catch!

PR #169, opened 2019-02-21
Add grep-jump style commands for *symbols* and *references*

hinshun:
Adds the following commands piggybacking `grep-jump` from `grep.kak`:
- `lsp-references-next-match`: Jump to the next references match
- `lsp-references-previous-match`: Jump to the previous references match
- `lsp-symbols-next-match`: Jump to the next symbols match
- `lsp-symbols-previous-match`: Jump to the previous symbols match


ul:

Nice idea and implementation =)

PR #175, opened 2019-03-02
Add instrunctions to install on Void Linux

maxice8:


ul:

PR #181, opened 2019-03-18
add --path . to plug installation method

andreyorst:
This is weird, but today my cargo ranted on me that I'm not able to use `cargo install`:
```
error: Using `cargo install` to install the binaries for the package in current working directory is no longer supported, use `cargo install --path .` instead
```


ul:
Yeah, they changed it recently in cargo. Thanks for the update!

Issue #182, opened 2019-03-20
Add capability to set project root with environment variable


benjumanji:
Some build systems use recursively placed build files. In large scala projects it isn't weird to see `build.sbt` in multiple places, but the scala language server really needs to be started at the parent build.sbt. I don't think it's reasonable to expect kak-lsp to be able to support this, so allowing people to just _tell_ you where the root is would be a helpful feature. 

Other weirdo build systems that might benefit: bazel / tup / pants / waf

I'd also be happy to work on this.

ul:
Yes, PR is welcome for this feature!

benjumanji:
Project Root selection
==================

## Problem
Current solution allows the specification of a set of markers per filetype for
the project root. On encountering a new buffer we walk from the buffer towards
the filesystem root checking at each directory if we have a match for a marker.
We stop at the first marker. This works well build systems with a single root
file but for systems that may scatter their build files all over the project
directory (in some extreme cases in _every_ folder c.f.
[tup](http://gittup.org).

## Solution
Provide a hook to allow projects roots to be explicitly designated on a
per-filetype basis. This patch uses environment variables. For maximum
flexibility the following scheme is used: on new buffer detect filetype, and
look for all environment variables prefixed with `KAK_LSP_%FILETYPE%*` and add
the value of each to a set of roots and proceed as before, first match wins.
This will allow for both regular repos and monorepos browsed from the root to
specify subprojects accurately. Consider the following tree:
```
~/src
├── bar
│   ├── build.sbt
│   └── ext
│       ├── sub1
│       │   ├── build.sbt
│       │   └── src
│       │       └── main
│       │           └── scala
│       └── sub2
│           ├── build.sbt
│           └── src
│               └── main
│                   └── scala
└── foo
    ├── build.sbt
    └── ext
        ├── sub1
        │   ├── build.sbt
        │   └── src
        │       └── main
        │           └── scala
        └── sub2
            ├── build.sbt
            └── src
                └── main
                    └── scala
```
Then an enviroment with `KAK_LSP_SCALA_FOO="$HOME/src/foo"` and
`KAK_LSP_SCALA_BAR="$HOME/src/bar"` would allow both sets of files to
successfully be associated with the correct build. Of course users that only
browse from single sources will be catered too as well and need not suffix at
all. Note that the `FOO` and `BAR` suffixes are _arbitrary_ and have no meaning
when it comes to matching.

benjumanji:
[...]

PR #184, opened 2019-03-23
Allow project roots to be specified by environment

benjumanji:
Allow per language marker files to be overridden by directly specifying
the project root for a given buffer. When searching for the project root
for a buffer first see if there are any environment variables prefixed
by `KAK_LSP_PROJECT_ROOT_%LANG%`. If so then iterate over all matching
variable values checking if they are a prefix of the path to the current
buffer. If so the first matching prefix is selected as the project root,
else return the buffers path. If there are no matching environment
variables then proceed as before.

### DIFF
--- a/src/project_root.rs
+++ b/src/project_root.rs
@@ -1,7 +1,18 @@
 use glob::glob;
+use std::collections::HashSet;
+use std::env;
 use std::path::PathBuf;
 
-pub fn find_project_root(roots: &[String], path: &str) -> String {
+pub fn find_project_root(language: &str, markers: &[String], path: &str) -> String {
+    let vars = gather_env_roots(language);
+    if vars.is_empty() {
+        roots_by_marker(markers, path)
+    } else {
+        roots_by_env(&vars, path).unwrap_or_else(|| roots_by_marker(markers, path))
+    }
+}
+
+pub fn roots_by_marker(roots: &[String], path: &str) -> String {
     let mut pwd = PathBuf::from(path);
     if pwd.is_file() {
         pwd.pop();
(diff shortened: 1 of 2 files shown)


benjumanji:
I have roughly no idea what I am doing with rust so please feel free to mercilessly point out flaws.

benjumanji:
solves #182

src/project_root.rs
```diff
@@ -1,7 +1,18 @@
...
+use std::collections::HashSet;
+use std::env;
 
-pub fn find_project_root(roots: &[String], path: &str) -> String {
+pub fn find_project_root(language: &str, markers: &[String], path: &str) -> String {
+    let vars = gather_env_roots(language);
+    if vars.is_empty() {
+    	roots_by_marker(markers, path)
+    } else {
+        roots_by_env(&vars, path)
```
ul: It worth running `cargo fmt` to make formatting consistent with the rest of the codebase.

src/project_root.rs
```diff
@@ -1,7 +1,18 @@
...
 use std::path::PathBuf;
+use std::collections::HashSet;
+use std::env;
 
-pub fn find_project_root(roots: &[String], path: &str) -> String {
+pub fn find_project_root(language: &str, markers: &[String], path: &str) -> String {
+    let vars = gather_env_roots(language);
+    if vars.is_empty() {
+    	roots_by_marker(markers, path)
+    } else {
```
ul: I think that it might be good to fall back to `root_by_marker` not only when no appropriate env vars are found, but also when `roots_by_env` didn't match any of vars. In that case, visiting random project while running the same kak-lsp session would give the expected result.

src/project_root.rs
```diff
@@ -24,3 +35,31 @@ pub fn find_project_root(roots: &[String], path: &str) -> String {
...
+  let p = PathBuf::from(path);
+  let pwd =
+      if p.is_file() {
+          p.parent().unwrap().to_path_buf()
+      } else {
+          p
+      };
+  roots
+      .iter()
+      .find(|x| x.starts_with(&pwd))
```
ul: Shouldn't it be `.find(|x| pwd.starts_with(x))`?

src/project_root.rs
```diff
@@ -24,3 +35,31 @@ pub fn find_project_root(roots: &[String], path: &str) -> String {
         }
     }
 }
+
+pub fn gather_env_roots(language: &str) -> HashSet<PathBuf> {
+    let prefix = format!("KAK_LSP_PROJECT_ROOT_{}", language.to_uppercase());
+    debug!("Searching for vars starting with {}", prefix);
+    let p = prefix.as_str();
```
ul: nit: I didn't check, but it should be enough just to use `&prefix` below without explicit conversion.

src/project_root.rs
```diff
@@ -24,3 +35,31 @@ pub fn find_project_root(roots: &[String], path: &str) -> String {
...
+    let mut roots: HashSet<PathBuf> = HashSet::new();
+    env::vars()
```
ul: Idiomatic Rust would be like:
```
env::vars()
      .into_iter()
      .filter(|(k, _v)| k.starts_with(p))
      .map(|(_k, v)| PathBuf::from(v))
      .collect()
```
Note that intermediate variable `root` is eliminated and there is no semicolon after `collect` so it will be returned as the result of the function.

benjumanji: Oh that's neat.

benjumanji:
Thanks for the feedback!

ul:

PR #185, opened 2019-03-24
Add docstrings for the various diagnostics-related commands.

Screwtapello:
I was having difficulty getting diagnostic flags to show up, and found it frustrating that  `lsp-diagnostic-lines-enable` sounded like it should do what I wanted, but instead it just complained "wrong argument count".

It turns out my actual problem was I hadn't called `lsp-enable`, but I figure more detailed docstrings never hurt.


ul:
Thanks for adding those!

Issue #161, opened 2019-01-24
lsp_tab_size option possibly unnecessary


JJK96:
Kakoune has the option `tabstop` which does exactly the same as `lsp_tab_size` for as far as I can tell.

So is `lsp_tab_size` necessary?

ul:
Good catch, I think it's not necessary then.

PR #190, opened 2019-04-04
README: lsp_tab_size was replaced with the standard tabstop option.

Screwtapello:
In #161, the lsp_tab_size option was removed, since Kakoune already provides the tabstop option. However, removing this option from the documentation leaves no clue that it affects the lsp-formatting command, so we update the lsp-formatting documentation to mention the options it uses.


ul:
Good catch, thanks for updating docs!

PR #196, opened 2019-05-06
Send window/progress messages to editor

ricochet1k:
Rust Language Server sends useful progress messages to communicate what it is doing, and the VS Code rust extension displays them in the statusbar.

This allows users to override the `lsp-handle-progress` command. For example: 
```
declare-option -hidden str modeline_progress ""
define-command -hidden -params 4 -override lsp-handle-progress %{
  set global modeline_progress %sh{
    echo $1${2:+": $2"}${3:+" $3%"}${4:+" ✓"}
  }
}
```
And then put `%opt{modeline_progress}` in your `modelinefmt` to see what RLS is doing.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -685,6 +685,11 @@ define-command -hidden lsp-replace-selection -params 1 -docstring %{
     nop %sh{ rm $kak_opt_lsp_text_edit_tmp }
 }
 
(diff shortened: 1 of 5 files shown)


ul:

Yes, handling progress is a nice feature to have.

ul:
Added your recipe to the wiki: https://github.com/ul/kak-lsp/wiki/Handling-progress-notifications

Issue #110, opened 2018-10-14
Simplify dispatching


ul:
Ref https://github.com/ul/kak-lsp/commit/3f121aff1259428874fb17a1014cc54e40b39c72#r30889210

ricochet1k:
I'm working on a big refactor based on this idea and a few others I've come across in the process, just FYI.

ul:
Amazing news! Just be aware of https://github.com/ul/kak-lsp/tree/better-character-offset-handling branch. You might want to base your work on it to avoid major conflict as I hope to get this branch into the master soon.

ricochet1k:
Turns out, I will not be able to submit any pull requests, due to the Unlicense and copyright assignment restrictions, because I work for Google. https://opensource.google.com/docs/patching/#forbidden

If you want to switch to a more traditional open source license, such as Apache 2 or MIT like most Rust projects, then I would be able to help. There is more information about licenses here: https://opensource.google.com/docs/using/license/

Delapouite:
Does this license restriction also applies when you make a personal contribution from home on Sundays?

ricochet1k:
Yes it does. As part of my condition of employment, Google owns copyright of all code I write. You can see a hint of that here: https://opensource.google.com/docs/iarc/

alexherbo2:
_All your base are belong to us._

ricochet1k:
I'm applying for an exception. It will take a few weeks to hear back, but they might grant me a copyright release.

ul:
The primary objective of choosing Unlicense was to support Kakoune's model and to remove all obstacles in the community's way. Ironically, it constrained contribution.

I personally am not attached to the Unlicense or any other Public Domain license because I intend to liberate kak-lsp effectively, not formally. If it requires switching to a license which sounds less free, but pragmatically gives more freedom, I'm okay with it.

However, I'd like to hear the opinion of other people as kak-lsp is already a kind of _res publica_, and people who spent their time coding, discussing and using kak-lsp may have ideas about licensing they see the most resonating with the project.

ul:
@ricochet1k how do you think, would dual-licensing under Unlicense/MIT do? There is even a precedent in a Rust ecosystem: https://github.com/docopt/docopt.rs/commit/11b5ff3d3939b7d4ba0e09e457f802a37bb5c5dc

ul:
done: 55dfc83409b9b7d3556bacda8ef8b71fc33b58cd

robertmeta:
[...]

PR #197, opened 2019-05-06
Use callbacks instead of a separate dispatch function for responses

ricochet1k:
Changes ctx.call and ctx.notify to take the request type as a generic parameter, which simplifies calling and allows automatic parsing of the response type, and also eliminated the need for dispatch_server_response.

Created types to represent CCLS requests.

Also, EditorMeta was being passed by reference and then cloned frequently even though the caller had no need for it anymore. Changed to move instead and was able to remove lots of clone()s.

This is related to #110 but does not complete it.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,8 +1,11 @@
 use crate::types::*;
 use crossbeam_channel::Sender;
-use jsonrpc_core::{self, Call, Id, Params, Version};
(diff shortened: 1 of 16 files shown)


ul:
It's definitely a step forward to a cleaner implementation, thanks for doing it and for going through hoops of applying for a copyright release.

PR #199, opened 2019-05-07
CodeAction support

ricochet1k:
Enables basic support for code actions. I haven't tested this extensively, but it is enough to allow RLS's "Deglob Imports" action to work correctly.

In the future, I'd like to make it possible to show the number of available code actions in the modeline, maybe by setting a list option instead of immediately executing the menu command?

### DIFF
--- a/src/language_features/codeaction.rs
+++ b/src/language_features/codeaction.rs
@@ -1,0 +1,69 @@
+use crate::context::*;
+use crate::types::*;
+use crate::util::*;
+use itertools::Itertools;
+use lsp_types::request::*;
+use lsp_types::*;
+use serde::Deserialize;
+use url::Url;
+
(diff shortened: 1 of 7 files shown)


src/language_features/codeaction.rs
```diff
@@ -0,0 +1,69 @@
...
+    ctx: &mut Context,
+) {
+    let result = match result {
+        Some(result) => result,
+        None => return,
+    };
+    match result {
+        CodeActionResponse::Commands(cmds) => {
+            if cmds.is_empty() {
+                return;
```
ul: How do you think, would it be better using `lsp-show-error` or another way of communicating back that there are no actions available? To not keep the user waiting in uncertainty for the response.

PR #208, opened 2019-06-02
Add Nim support

SolitudeSF:
Nim project is defined by the file with `.nimble` extension. I suppose i cannot add a substring as valid root?

### DIFF
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -96,3 +96,8 @@ args = ["--stdio"]
 
 [language.php.initialization_options]
 storagePath = "/tmp/intelephense"
+
+[language.nim]
+filetypes = ["nim"]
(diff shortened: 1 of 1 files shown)


ul:

You can use glob patterns, i.e. https://github.com/ul/kak-lsp/blob/5a168cebe32b423580d97f6a817c444de69da2f9/kak-lsp.toml#L46

So I guess for Nimble case it would look like `*.nimble`

SolitudeSF:
does the order of the `roots` list matters? since one git repo can have mutiple nimble projects (in subdirs).

ul:
It doesn't matter for that case as all markers are checked at each level before going level up.

There is a way to specify an arbitrary directory as a project root, though. Ref https://github.com/ul/kak-lsp/wiki/Project-root-selection for details.

PR #209, opened 2019-06-13
elm language server settings

bburdette:
I've been using these settings for elm for a few days.  So far so good.


ul:
Thank you for your contribution! It worth adding a short instruction about how to install language server to wiki: https://github.com/ul/kak-lsp/wiki/How-to-install-servers

bburdette:
Thanks for the merge.  wiki updated!

PR #218, opened 2019-07-05
Update to lsp-types 0.58.0 and regex 1.0

daboross:
I've just submitted this to update to lsp-types 0.58, which brings support for mixed CodeAction and Command responses to code action requests.

This also includes updating to regex 1.0, which as I understand it was largely a commitment to stability? It has some small breaking changes, but the features affected aren't used in this codebase.


ul:
Thank you for the upgrade!

PR #227, opened 2019-07-23
Use Box<FnOnce>, stabilized in 1.35.0

daboross:
This removes code working around the lack of `Box<FnOnce>`!

If kak-lsp previously had a minimum supported rust version, this raises that to 1.35.0. Understandable if this isn't wanted because of that. If that's fine, this should be a pretty small change.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -20,9 +20,7 @@ pub struct Document {
     pub text: ropey::Rope,
 }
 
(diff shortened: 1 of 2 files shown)


ul:
Thanks for the cleanup! kak-lsp's minimum supported version of Rust is the latest stable, that's it.

PR #229, opened 2019-07-31
add lsp-disable and lsp-disable-window

andreyorst:
As suggested in https://github.com/ul/kak-lsp/issues/228 I've added two commands `lsp-disable` and `lsp-disable-window`. However after executing those hovering still works for some reason. Did I missed something?

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -952,6 +952,17 @@ define-command -hidden lsp-enable -docstring "Default integration with kak-lsp" 
     }
 }
 
+define-command -hidden lsp-disable -docstring "Disable kak-lsp in the window scope" %{
+    remove-highlighter global/cquery_semhl
+    remove-highlighter global/lsp_references
+    lsp-inline-diagnostics-disable global
+    lsp-diagnostic-lines-disable global
+    unmap global goto d '<esc>: lsp-definition<ret>' -docstring 'definition'
+    unmap global goto r '<esc>: lsp-references<ret>' -docstring 'references'
+    remove-hooks global lsp
+    lsp-exit
+}
+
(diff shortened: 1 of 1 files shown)


andreyorst:
ah. found that. Interestingly enough that `lsp-auto-hover` hooks are always `global` scoped. Can't this cause problems with `window` scoped lsp?

rc/lsp.kak
```diff
@@ -961,8 +975,8 @@ define-command lsp-enable-window -docstring "Default integration with kak-lsp in
     lsp-inline-diagnostics-enable window
     lsp-diagnostic-lines-enable window
 
-    map window goto d '<esc>: lsp-definition<ret>' -docstring 'definition'
-    map window goto r '<esc>: lsp-references<ret>' -docstring 'references'
+    map window goto d '<esc>: lsp-definition<ret>'
```
ul: Any specific reason to remove docstring here and below?

ul:
> Interestingly enough that lsp-auto-hover hooks are always global scoped. Can't this cause problems with window scoped lsp?

Yes, I think I overlooked it when introduced `lsp-enable-window`. Most likely we want `auto` hooks to be scope-parametrized as well.

ul:
Thanks for adding this command! Let's merge it (after restoring docstrings) and then have `auto` commands fixed as well.

rc/lsp.kak
andreyorst: Oi!

andreyorst:
I've put docstrings back. It seems that I've copied map and unmap commands around and unmap doesn't accept docstring, and I've removed those docstrings by mistake.

PR #238, opened 2019-09-09
Insert before the cursor for all text edits

lePerdu:
Currently, text is isnerted before the cursor when at column 0 and after the cursor at other columns. This is not correct, though: it "skips" over the position between columns 1 and 2 (in Kakoune coordinates) and makes inserts that aren't at the beginning of a line off by 1. Text should instead always be inserted before the cursor in Kakoune, not just when at the beginning of the line.

As far as I can see, `KakouneTextEditCommand::InsertAfter` isn't needed at all: cursors in Kakoune can refer to the newline character (or end-of-file if there is no newline at the end of the file) of each line, so even if the LSP column is at the end of the line, the insert can still be represented as an insertion before the cursor in Kakoune. (I hence removed it and the corresponding `lsp-insert-after-selection` command.)

Here is an example of the current behavior of `lsp-format`:

With this buffer:
```
int main() {
    if (true) {
        return long_function_name(alongargumentname, anotherargument, "a string", 100000000);
    }
}
```

Calling `lsp-format` (with clangd) sends back an insert with a new line and some indentation results in this
```
int main() {
    if (true) {
        return long_function_name(a
                longargumentname, anotherargument, "a string", 100000000);
    }
}
```

With this PR, `lsp-format` results in this (the output from the `clang-format` command):
```
int main() {
    if (true) {
        return long_function_name(
                alongargumentname, anotherargument, "a string", 100000000);
    }
}
```

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -688,19 +688,6 @@ define-command -hidden lsp-show-message -params 2 -docstring %{
     info %arg{2}
 }
 
-define-command -hidden lsp-insert-after-selection -params 1 -docstring %{
-    Insert content after current selections while keeping cursor intact.
-    It is used to apply text edits from language server.
-} %{
-    declare-option -hidden str lsp_text_edit_tmp %sh{ mktemp }
-    declare-option -hidden str lsp_text_edit_content %arg{1}
-    execute-keys %sh{
-        printf "%s" "$kak_opt_lsp_text_edit_content" > $kak_opt_lsp_text_edit_tmp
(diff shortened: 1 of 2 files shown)


ul:
Thank you for your contribution! And sorry for the delay...

Good to see this code simplified and corrected, I briefly tested it manually with the language servers I use and it seems to work. (It's a shame that we don't have an automated test suite for it, I hope I'll find time to build it, or anyone will be keen to contribute).

PR #242, opened 2019-09-25
Feature/workspace edit code actions

topisani:
Still new to rust, so not certain about the implementation, but it works.

This supports the code actions issued by ccls, and is the only way i have tested it


ul:
Looks good to me, thank you for your contribution!

Issue #241, opened 2019-09-22
kak-lsp appears to not run when kak-lsp.toml exists


wbthomason:
I have a strange bug: When I have no `~/.config/kak-lsp/kak-lsp.toml`, `kak-lsp` starts and runs correctly (for a Python file). When there is a `kak-lsp.toml` (configuring C++ and Rust, but not Python - from the wiki, I got the impression that I could not configure Python's LSP in this file?), `kak-lsp` seems to silently fail - I get no completions, lints, etc. I'm not quite sure how to go about debugging this.

Apologies if this is a naive mistake; I'm super new to Kakoune and kak-lsp both.

ul:
Hello, this indeed sounds like a strange behaviour of kak-lsp. Let's double-check things one by one:

1. kak-lsp reads only one configuration file, from the directory which is system-dependent. Please check https://github.com/ul/kak-lsp/#configuring-kak-lsp for details. When the configuration file is missing kak-lsp loads default one which is embedded into the binary.

2. The configuration file doesn't extend default configuration but rather overwrites it. If you want one more language server to existing ones, or change the existing one, then you need to copy the entire config and edit it.

3. You can debug kak-lsp by the following instruction https://github.com/ul/kak-lsp/#troubleshooting

Please let me know if you get any new input so I can propose a few other things.

wbthomason:
Ah, it seems that (2) caused my issue - I did not realize that the configuration file does not extend the default, but overwrites it, and thus the configuration file I had was definitely incomplete. Copying the default config and customizing it fixes the behavior I mentioned above. 

Is this documented somewhere on the repo? I'd be happy to make a PR adding it to the README if it isn't (I didn't see it there, but I may have missed something).

Thanks!

ul:
I don't think it's properly documented anywhere. PR is welcome!

PR #244, opened 2019-09-30
Document configuration file: overwriting vs extending

wbthomason:
To prevent issues such as #241


ul:
Thank you for the docs improvement!

PR #247, opened 2019-10-11
Pass buffer content using stdio instead of environment variables

### LINKED ISSUE #245: Pass buffer content using stdin instead of environment variables
At the moment `lsp-did-change` and `lsp-did-open` both use environment variables to pass the content of the buffer.

While the overall environment can get quite big, each single environment variable appears to be limited to a hard 128kib, at least on linux ([see](https://unix.stackexchange.com/questions/336934/raise-128kib-limit-on-environment-variables-in-linux)). 

If an environment variable is bigger than this, it causes the shell scope to not even be run. This can be tested with the following command which is a noop with a selection smaller than ~130K chars, and wipes it beyond that
`reg dquote %sh{ printf '%s' "$kak_selection" }; exec R`

I think a solution here would be to pass the buffer content over stdin using `<a-|>`, which does not have that limitation. Since anyway the shell scopes only rely on side effects, this should be straightforward.

There might be other places where this limitation is relevant.

occivink:
Fixes #245 

Other notable changes:
* lsp-did-change was not operating asynchronously, but lsp-did-open was. They are now both asynchronous
* the three chained sed calls have been merged into one (I think that should be portable? Would be nice if on macos could test, since afaik it has noticeable differences)

To test this, you can check that with a file bigger than 128kib, changing the buffer content causes both kak-lsp and the language server to use _some_ cpu. Before, it did not and kept idling.

However jumping to definitions was working before, even on files that big. It's not entirely clear to me what didChange is supposed to do, but it does not seem entirely necessary?

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -69,12 +69,21 @@ define-command -hidden lsp-did-change -docstring "Notify language server about b
     evaluate-commands %sh{
         if [ $kak_opt_lsp_timestamp -eq $kak_timestamp ]; then
             echo "fail"
-        else
-            echo "evaluate-commands -draft -no-hooks %{set-option buffer lsp_timestamp %val{timestamp}; execute-keys '%'; set-option buffer lsp_draft %val{selection}}"
         fi
     }
-    nop %sh{ (
-lsp_draft=$(printf '%s.' "${kak_opt_lsp_draft}" | sed 's/\\/\\\\/g' | sed 's/"/\\"/g' | sed "s/$(printf '\t')/\\\\t/g")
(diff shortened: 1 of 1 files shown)


ul:
Thanks for the fix!

> However jumping to definitions was working before, even on files that big. It's not entirely clear to me what didChange is supposed to do, but it does not seem entirely necessary?

didChange is the way LSP server knows actual buffer content when changes are not saved to the file. I guess if you will edit a buffer in the region affected by the size limitation so definition will be moved then jump to definitions will end up in an incorrect position.

PR #251, opened 2019-10-19
Replace bingo with gopls

robertmeta:
Bingo has been discontinued and go-langserver is very slow (at least on go.mod projects), the best drop-in replacement is gopls (https://github.com/golang/tools/tree/master/gopls).


ul:
Thanks for the update!

PR #255, opened 2019-11-03
fix void install command

LennyPenny:
the correct command is `xbps-install` :)


ul:
Good catch, thank you =)

PR #256, opened 2019-11-05
Open goto definition in jump client

### LINKED ISSUE #250: Go to definition should open in jump client
# Steps to reproduce

1. Open two client in the same session
2. In the second client execute:
```
rename-client jump
set global jumpclient jump
```
3. In the first client use the go-to-definition command

# Observed behaviour

Jumps to the definition in the first client

# Expected behaviour

Jumps to the definition in the second client

JJK96:
Fix #250


ul:
Thank you for the fix!

PR #260, opened 2019-11-21
Improve debugging experience.

dpc:
With this change the initial request to `kak-lsp` will also be logged into a file.

The full file I'm using it with: https://github.com/ul/kak-lsp/wiki/Usage-with-plug.kak


dpc:
Actually, this is not working like I want it to.

PR #262, opened 2019-11-23
Code action with diagnostics

osiox:
Following change adds diagnostics to codeAction request and is required to make code actions work with eclipse jdt.ls (along with [this PR to jdt.ls](https://github.com/eclipse/eclipse.jdt.ls/pull/1278)).

Note: I actually do not know rust, please point out if there is a better way to write it.


ul:
Thank you for your contribution! I think your style is fine. The only minor thing is that it worth running `cargo fmt` for formatting but I'll fix it on the master.

PR #266, opened 2019-12-14
Add LaTeX support

Ordoviz:
I couldn’t think of any sensible roots so I simply choose `.git`. Server is [texlab](https://github.com/latex-lsp/texlab).


ul:
Thank you for your contribution! Would you mind adding an entry here https://github.com/ul/kak-lsp/wiki/How-to-install-servers about the installation of texlab?

Ordoviz:
Done.

PR #269, opened 2019-12-22
Semantic highlighting proposal

topisani:
Support for https://github.com/microsoft/vscode-languageserver-node/pull/367

Tested with `clangd`

~~The main reason for the WIP status is that it depends on [this pr](https://github.com/gluon-lang/lsp-types/pull/127) in lsp-types. For now i have changed the path in `Cargo.toml` to use a local `lsp-types` repo, which will obviously be changed back before merging.~~ *EDIT:* This has now been merged

The second point is, currently the textmate scopes are being directly translated to kakoune faces.
Since textmate scopes are heirarchical, we should have a way of defining `scope -> face` mappings, which would let `entity.type.cpp` map to the face for `entity.type`.


src/language_features/semantic_highlighting.rs
```diff
@@ -0,0 +1,82 @@
...
+        .as_ref()
+        .unwrap()
+        .scopes
+        .as_ref()
+        .expect("Server sent semantic highlight notification without setting capability");
+    let offset_encoding = ctx.offset_encoding.to_owned();
+    let ranges = params
+        .lines
+        .iter()
+        .map(|info| {
```
ul: nit:  [flat_map](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map) might fit there.

src/language_features/semantic_highlighting.rs
```diff
@@ -0,0 +1,82 @@
...
+                        "{}|{}",
+                        lsp_range_to_kakoune(&range, &document.text, &offset_encoding),
+                        face
+                    )
+                })
+                .join(" ")
+        })
+        .join(" ");
+    let command = format!(
+        "eval -buffer %§{}§ %§set buffer lsp_semantic_highlighting {} {}§",
```
ul: It's safer to use `editor_quote` instead. It's unfortunate that we haven't cleaned up existing usage of `§` in `ccls.rs` yet, it gives a bad example.

src/types.rs
```diff
@@ -199,3 +200,97 @@ pub enum OffsetEncoding {
     #[serde(rename = "utf-16")]
     Utf16,
 }
+
+#[derive(Debug, Eq, PartialEq, Clone, Default, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ServerCapabilities {
```
ul: Is it possible to use https://docs.rs/lsp-types/0.67.1/lsp_types/struct.ServerCapabilities.html directly?

topisani: Oh Yeah, my mistake, we do use the lsp-types one. This was before I realized it would be better to just contribute to that project

src/language_features/semantic_highlighting.rs
topisani: Ah yes, ofc, I had some issue and replaced it with this and forgot to change it back when that wasn't it

topisani:
Btw, if you need any help managing issues etc, hit me up!

src/language_features/semantic_highlighting.rs
topisani: ah no, it had something todo with some scoping errors i didn't fully understand, and this was the easiest way for me to fix it

ul:
> Btw, if you need any help managing issues etc, hit me up! I see a lots been going on lately

Yes, a lot, and at the same time, I've got much less spare resources. Your help is much appreciated!

src/language_features/semantic_highlighting.rs
ul: This is weird, I don't understand either why does it complain about `line` lifetime as `u64` could be just copied. However, we can use `move` and also take `offset_encoding` by reference instead of cloning it (performed implicitly by `to_owned`):

```
        .flat_map(|info| {
            let line: u64 = info.line as u64;
            let offset_encoding = &ctx.offset_encoding;
            info.tokens.iter().map(move |t| {
                let scope = scopes
                    .get(t.scope as usize)
                    .expect("Semantic highlighting token sent for out-of-range scope");
                let face = get_face_for_scope(scope);
                let range = Range {
                    start: Position::new(line, t.character.into()),
                    end: Position::new(line, (t.character + u32::from(t.length)).into()),
                };
                format!(
                    "{}|{}",
                    lsp_range_to_kakoune(&range, &document.text, offset_encoding),
                    face
                )
            })
        })
```

topisani:
Damn, i see i messed up the git stuff. All that effort trying to keep these two branches separated...

topisani:
http://ix.io/25x2

This is the list of scopes available from eclipse.jdt.ls, and its a lot less managable than the clangd list. So we do need a scoped mapping to faces. Regarding how, on the initialized notification, the server sends a list of available scopes, and then, when the actual highlighting is sent, its just indexes into that initial list. That means, we do not need dynamic scope->face mapping, we can just map that list of scopes to a list of faces upon initialization. It also means that the splitting/lookup/matching can be done in rust, and not kakrc, which is nice.

I think having the configuration in `kakrc` as a `str-to-str` map from `scope` to `face`, is what would make the most sense. We can even have a setter for that option, which sends it to kak-lsp, and lets that remap the scopes. This also lets colorschemes define a custom mapping.

topisani:
the option in `kakrc` turned out to be infeasable - editing it in the `toml` file is fine, at least for now. 

Tested with clangd and eclipse.jdt.ls

ul:
> Damn, i see i messed up the git stuff. All that effort trying to keep these two branches separated...

I can help to clean it up if I'll get spare time before you. It should be a checkbox somewhere allowing me to push into your branch.

src/language_features/semantic_highlighting.rs
```diff
@@ -19,27 +20,16 @@ pub fn semantic_highlighting_notification(params: Params, ctx: &mut Context) {
...
         .lines
         .iter()
         .flat_map(|info| {
             let line: u64 = info.line as u64;
             let offset_encoding = &ctx.offset_encoding;
             info.tokens.iter().map(move |t| {
-                let scope = scopes
-                    .get(t.scope as usize)
+                let face = faces.get(t.scope as usize)
                     .expect("Semantic highlighting token sent for out-of-range scope");
```
ul: Does it mean kak-lsp would crash if the scope is not mapped to face in the config? Would it be better to log an error message (probably even into Kakoune's `*debug*` buffer) and skip this token?

ul: Ah, you are mapping scopes advertised by the language server but not configured to empty strings in `make_scope_map`. So case when face is not found *is* exceptional.

src/language_features/semantic_highlighting.rs
```diff
@@ -75,7 +69,32 @@ pub fn debug_scopes(meta: EditorMeta, ctx: &mut Context) {
...
+pub fn make_scope_map(ctx: &mut Context) -> std::vec::Vec<std::string::String> {
+    let faces: HashMap<String, &String> = ctx.config.semantic_scopes.iter().map(|(k,v)| (k.replace("_", "."), v)).collect();
+
+    let scopes = ctx
+        .capabilities
+        .as_ref()
+        .and_then(|x| x.semantic_highlighting.as_ref())
+        .and_then(|x| x.scopes.as_ref());
+
+    if scopes == None {
```
ul: nit: I've seen `if scopes.is_none()` noted as more preferable in few places.

src/language_features/semantic_highlighting.rs
```diff
@@ -75,7 +69,32 @@ pub fn debug_scopes(meta: EditorMeta, ctx: &mut Context) {
...
+      return Vec::new();
+    }
+    let scopes = scopes.unwrap();
+
+    let faces: Vec<String> = scopes.iter().map(|scopes| {
+      for scope in scopes {
```
ul: nit: I think more idiomatic Rust could look more like

```rust
pub fn make_scope_map(ctx: &mut Context) -> std::vec::Vec<std::string::String> {
    let faces: HashMap<String, &String> = ctx
        .config
        .semantic_scopes
        .iter()
        .map(|(k, v)| (k.replace("_", "."), v))
        .collect();

    let scopes = ctx
        .capabilities
        .as_ref()
        .and_then(|x| x.semantic_highlighting.as_ref())
        .and_then(|x| x.scopes.as_ref());

    if scopes == None {
        return Vec::new();
    }
    let scopes = scopes.unwrap();

    map_scopes_to_faces(scopes, faces)
}

fn map_scopes_to_faces(
    scopes: &Vec<Vec<String>>,
    faces: HashMap<String, &String>,
) -> std::vec::Vec<std::string::String> {
    let find_face = |scope: &String| {
        scope
            .chars()
            .enumerate()
            .filter_map(|(i, x)| {
                if x == '.' {
                    // Trading extra HashMap lookups for creating intermediate collection seems to be fair.
                    faces.get(&scope[0..i])
                } else {
                    None
                }
            })
            .last()
            .map(|face| face.to_string())
    };
    scopes
        .iter()
        .map(|scopes| {
            scopes
                .iter()
                .find_map(find_face)
                .unwrap_or_else(|| String::new())
        })
        .collect()
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn map_scopes_to_faces_should_map_unknown_scope_to_empty_face() {
        let scopes = vec![
            vec![String::from("just.noise")],
            vec![String::from("some.scope")],
        ];
        let mut faces: HashMap<String, &String> = Default::default();
        let some_face = String::from("some.face");
        faces.insert(String::from("some.another.scope"), &some_face);
        let faces = map_scopes_to_faces(&scopes, faces);
        assert_eq!(Some(&String::from("")), faces.get(1));
    }

    #[test]
    fn map_scopes_to_faces_should_map_scope_by_prefix() {
        let scopes = vec![
            vec![String::from("just.noise")],
            vec![
                String::from("some.non-matching.scope"),
                String::from("some.nested.scope"),
            ],
        ];
        let mut faces: HashMap<String, &String> = Default::default();
        let some_face = String::from("some.face");
        faces.insert(String::from("some.nested"), &some_face);
        let faces = map_scopes_to_faces(&scopes, faces);
        assert_eq!(Some(&String::from("some.face")), faces.get(1));
    }

    #[test]
    fn map_scopes_to_faces_should_map_scope_by_longest_prefix() {
        let scopes = vec![
            vec![String::from("just.noise")],
            vec![
                String::from("some.scope.matching.short.prefix"),
                String::from("some.nested.scope"),
            ],
        ];
        let mut faces: HashMap<String, &String> = Default::default();
        let some = String::from("some");
        let some_face = String::from("some.face");
        faces.insert(String::from("some"), &some);
        faces.insert(String::from("some.nested"), &some_face);
        let faces = map_scopes_to_faces(&scopes, faces);
        assert_eq!(Some(&String::from("some.face")), faces.get(1));
    }
}
```

I extracted `map_scopes_to_faces` for easier testing. Note the last test would fail with this code (as well as with original code). I don't know if this case is realistic enough or not for us to care about.

This is one of the ways to account for this edge case:

```rust
fn map_scopes_to_faces(
    scopes: &Vec<Vec<String>>,
    faces: HashMap<String, &String>,
) -> std::vec::Vec<std::string::String> {
    let find_face = |scope: &String| {
        scope
            .chars()
            .enumerate()
            .filter_map(|(i, x)| if x == '.' { Some(&scope[0..i]) } else { None })
            .enumerate()
            .filter_map(|(n, element)| faces.get(element).map(|face| (n, face)))
            .last()
            .map(|(n, face)| (n, face.to_string()))
    };
    scopes
        .iter()
        .map(|scopes| {
            scopes
                .iter()
                .filter_map(find_face)
                .max_by_key(|(n, _)| *n)
                .map(|(_, x)| x)
                .unwrap_or_else(|| String::new())
        })
        .collect()
}
```

topisani: I'll use this code instead
Are you sure its supposed to behave like that in the last test case?

topisani: I do see the logic in the longest name being the most specific. But i assumed they were ordered by the order they should be applied in.

I this makes sense when only mapping some scopes to faces though

topisani:
I think this all works as intended now

ul:
Excellent, thank you!

Issue #40, opened 2018-05-22
Consider using textEdit for completion


ul:
> 23-Wed 01:00:04	Franciman	not sure it will be significant, but it may be useful. Do you remember the completion problem I had, and you said that you are keeping track of the offset. I found out that in the response to the textDocument/complete request, there can also be a textEdit which contains the range of text to apply the text transformation to (in layman's words write the completion :P)
> 23-Wed 01:01:08	Franciman	so instead of keeping track of the offset you can use the start position of the range of the text edit. This has also the advantage of being able to support, in future, the use of the textEdit, when you select a completion, which may be useful
> 23-Wed 01:02:04	Franciman	since insertText field is deprecated in favor of the textEdit, when the text to insert is different from the label

vnagarnaik:
Could you prioritize this feature please? Along with adding support for "additionalTextEdits"?

I am using a C++ language server that provides completions and suggests the appropriate #include header file through "additionalTextEdits". This would make working in kakoune much more productive.

I don't know much rust to be helpful with an actual PR.

ul:
Yes, I think it would be useful to have this done. It should be much easier to implement now because we have textEdit counterparts in lsp.kak. The only problem which remains is that we should cancel Kakoune's autocompletion when item is selected and do our own text edit instead. I'm not sure how to do it in elegant way, going to ask around at IRC channel.

ul:
Unfortunately I still don't know how to solve bottleneck I flagged in my previous comment. I raised an issue in Kakoune repo to get some advice.

ul:
This issue seems to be solvable now when we've got completions-commands branch in Kakoune, which I think eventually will be merged.

It's still non-trivial, though.

1. Applying and clearing textEdit and additionalTextEdits live while the user is cycling through completions menu is fragile as it involves many operations and can interrupt completions engine. I think the complexity of the solution doesn't worth it, and we will stick with default completions behaviour for cycling through items and will apply edits only when completion is confirmed.

2. As language server is not aware of multiselection, kak-lsp should take care of projecting textEdit to all selections, either by offsetting textEdit for the main selection or by making multiple requests to language server (which has an interesting side-effect that we can take the intersection of all responses to have only candidates which are relevant for all selections). We also need to assume something about additionalTextEdits, for example, that they are always used only for auto-inserting imports and we should apply them only once even in multiselection case.

Avi-D-coder:
tsserver sends adtional text edits adding namespace prefixes. 

[...]

PR #270, opened 2019-12-23
Add Snippet completion

topisani:
Uses slightly modified code from https://github.com/occivink/kakoune-snippets to parse snippets.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -1034,6 +1034,7 @@ lsp-stop-on-exit-enable
 hook -always -group lsp global KakEnd .* lsp-exit
 
 # SNIPPETS
+# This is a slightly modified version of occivink/kakoune-snippets
 
 decl -hidden range-specs lsp_snippets_placeholders
 decl -hidden int-list lsp_snippets_placeholder_groups
@@ -1041,27 +1042,29 @@ decl -hidden int-list lsp_snippets_placeholder_groups
 face global SnippetsNextPlaceholders black,green+F
 face global SnippetsOtherPlaceholders black,yellow+F
 
-def -hidden lsp-snippets-insert-completion -params 2 %{
-  exec -draft "<a-;><a-/>%arg{1}<ret>d"
-  eval -draft -verbatim lsp-snippets-insert %arg{2}
-  remove-hooks window lsp-post-completion
-  hook -once -group lsp-post-completion window InsertCompletionHide .* %{
-    try %{
-      lsp-snippets-select-next-placeholders
-      # On the next key, start replacing the placeholder contents
-      on-key %{ exec "<a-;>c%val{key}" }
+# First param is the text that was inserted in the completion, which will be deleted
+# Second param is the actual snippet
+def -hidden lsp-snippets-insert-completion -params 2 %{ eval -save-regs "a" %{
+    reg 'c' "%arg{1}"
+    exec -draft "<a-;><a-/>%val[main_reg_c]<ret>d"
+    eval -draft -verbatim lsp-snippets-insert %arg{2}
+    remove-hooks window lsp-post-completion
+    hook -once -group lsp-post-completion window InsertCompletionHide .* %{
+        try %{
+            lsp-snippets-select-next-placeholders
+            # On the next key, start replacing the placeholder contents
+            on-key %{ exec "<a-;>c%val{key}" }
+        }
     }
-  }
-}
+}}
 
-def lsp-snippets-insert -hidden -params 1 %<
-    eval %sh<
-        #<<
-        if ! command -v perl >/dev/null 2>&1; then
+def lsp-snippets-insert -hidden -params 1 %[
+    eval %sh{
+        if ! command -v perl > /dev/null 2>&1; then
             printf "fail '''perl'' must be installed to use the ''snippets-insert'' command'"
         fi
-    >
-    eval -draft -save-regs '^"' %<
+    }
+    eval -draft -save-regs '^"' %[
         reg '"' %arg{1}
         exec <a-P>
         # replace leading tabs with the appropriate indent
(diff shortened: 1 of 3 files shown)


rustfmt.toml
```diff
@@ -0,0 +1 @@
+tab_spaces = 4
```
ul: I don't mind, just curious is there any specific reason to set this option explicitly to the default value?

rc/lsp.kak
```diff
@@ -1018,3 +1032,199 @@ define-command lsp-disable-window -docstring "Disable kak-lsp in the window scop
...
+# SNIPPETS
+
+decl -hidden range-specs lsp_snippets_placeholders
+decl -hidden int-list lsp_snippets_placeholder_groups
+
+face global SnippetsNextPlaceholders black,green+F
+face global SnippetsOtherPlaceholders black,yellow+F
+
+def -hidden lsp-snippets-insert-completion -params 2 %{
+  exec -draft "<a-;><a-/>%arg{1}<ret>d"
```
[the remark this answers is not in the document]
topisani: Yep, i have some issues with that. I've fixed the angle bracket issue locally by storing it in a register.

The other place I've seen an issue is in multiline snippets. I think the right thing to do is to extract some prefix which is all that's given as the insertion text. In most cases we could probably just select all word characters from the begining of the text, but couldn't really find a good way to do that in rust - still quite new to it

rustfmt.toml
topisani: For some reason it kept formatting with 2 spaces for me, so I set this. Didn't realize it was a default. Maybe I have some global rustfmt file somewhere?

rc/lsp.kak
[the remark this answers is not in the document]
ul: > In most cases we could probably just select all word characters from the begining of the text, but couldn't really find a good way to do that in rust - still quite new to it.

I was about to propose a solution like this https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=590fdbaf82345754a1eeb778624a1dad but you already pushed a regex solution which I think fine too.

> Btw, I also intend to add support for completions with text edits (like adding imports), I plan for those to be applied in a hook like this when the menu is closed, how does that sound

Wow, are you going to fix #40? That would be awesome!

rustfmt.toml
ul: Could be the case, rustfmt does support global config: https://github.com/rust-lang/rustfmt/blob/master/Configurations.md

ul:
How do I cycle through placeholders? I think it worth adding this into a README (along with `snippet_support` option and other things related to snippets).

topisani:
Definitely, I have some stuff for mapping tab to go through the completion menu when open, and when closed cycle through placeholders. I'll add that to the readme

rc/lsp.kak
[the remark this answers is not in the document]
topisani: yes - the `additionalTextEdits` are now easy with this. The `textEdit` itself is a bit harder, still dont really know how to "undo" that `textEdit` once it has been applied.

For `additionalTextEdits`, id just wait and apply it on the `InsertCompletionHide` hook

src/language_features/completion.rs
```diff
@@ -70,12 +71,24 @@ pub fn editor_completion(
...
-                "{}|{}|{}",
-                escape_bar(&x.insert_text.unwrap_or(x.label)),
-                escape_bar(&doc),
-                escape_bar(&entry),
-            ))
+            let insert_text = &x.insert_text.unwrap_or(x.label);
+            let do_snippet = ctx.config.snippet_support;
+            let do_snippet = do_snippet && x.insert_text_format.map(|f| f == InsertTextFormat::Snippet).unwrap_or(false);
+            if do_snippet {
+              	let snippet = insert_text;
```
ul: nit: formatting seems to be off here and below

rc/lsp.kak
```diff
@@ -1018,3 +1032,202 @@ define-command lsp-disable-window -docstring "Disable kak-lsp in the window scop
...
+decl -hidden range-specs lsp_snippets_placeholders
+decl -hidden int-list lsp_snippets_placeholder_groups
+
+face global SnippetsNextPlaceholders black,green+F
+face global SnippetsOtherPlaceholders black,yellow+F
+
+# First param is the text that was inserted in the completion, which will be deleted
+# Second param is the actual snippet
+def -hidden lsp-snippets-insert-completion -params 2 %{ eval -save-regs "a" %{
+    reg 'c' "%arg{1}"
```
ul: Why does it save `a` reg but writes into `c`?

src/language_features/completion.rs
```diff
@@ -41,6 +41,7 @@ pub fn editor_completion(
     let unescape_markdown_re = Regex::new(r"\\(?P<c>.)").unwrap();
     let maxlen = items.iter().map(|x| x.label.len()).max().unwrap_or(0);
     let escape_bar = |s: &str| s.replace("|", r"\|");
+    let snippet_prefix_re = Regex::new(r"^[^\[\(<\n\$]+").unwrap();
```
ul: Are all other characters safe? Might worth restricting to word characters like `r"^\w+"` or even just ASCII alphanumeric like `r"^[[:alnum:]]+"`. I'm not sure because I don't fully understand what's fine to provide Kakoune.

topisani: I think it should be fine, as long as we regex::escape the stuff that's used to select it afterwards. I did realize that I accidentally regex::escaped both though...

topisani: I now regex escape the correct one. I believe i should in theory be able to send anything that's not a newline, but i thought something like this would make sense. It works even for snippets like `sizeof...(${1:parameter-pack})` now (the dots being handled correctly). I can't seem to find any other example where it breaks in rust or C++ at least...

I'd keep it like this, word characters only might break in languages that allow snake-case or snippets with operators etc.

topisani:
AFAIK this can be merged now

ul:
Thanks for polishing it! Let's merge but please have another PR with the README update as otherwise it would be hard to discover and properly configure this feature.

PR #279, opened 2020-01-16
Fix a quoting issue in lsp-find-error

thwischm:
The missing quotes prevented the command from working if the file path contains spaces.


ul:
Good catch, thank you for your contribution!

PR #281, opened 2020-01-30
Update dart language server command

mogenson:
We need to use a shell to find the correct path to the dart analysis
server source file. This should be in a known location relative to the
dart binary, regardless of whether the dart sdk was installed
stand-alone or part of the flutter toolkit. However the dart sdk and
binary location could be anywhere since it's commonly user installed
from source.


ul:
Thanks for the update!

PR #286, opened 2020-02-05
Don't display an error message for a CONTENT_MODIFIED error.

jneem:
According to the lsp specification [1], "if clients receive a
ClientModified error, it generally should not now it in the UI for
the end-user. Clients can resend the request if appropriate."

Note that this patch doesn't resend the request.

Since there's only one language-server error that we recognize, this
patch hard-codes it. There are currently two lsp-specific error codes,
so it doesn't make sense to pull in any dependencies for their
definition. If it ever does, the "lsp-server" crate exports the error
codes.

[1] https://microsoft.github.io/language-server-protocol/specification

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -13,6 +13,11 @@ use lsp_types::notification::Notification;
 use lsp_types::request::Request;
 use lsp_types::*;
 
+// This is an error code defined by the language server protocol, signifying that a request was
+// cancelled because the content changed before it could be fulfilled. In this case, the user
+// should not be notified.
+const CONTENT_MODIFIED: i64 = -32801;
+
 /// Start controller.
 ///
 /// Controller spawns language server for the given language and project root (passed as `route`).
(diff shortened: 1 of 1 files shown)


src/controller.rs
```diff
@@ -127,12 +132,15 @@ pub fn start(
                                             "{} language server doesn't support method {}",
                                             ctx.language_id, method
                                         ),
+                                        ErrorCode::ServerError(CONTENT_MODIFIED) => String::new(),
```
ul: It might be better to reorg code to not use msg emptiness for control flow, something like this:

```rust
                                    match failure.error.code {
                                        ErrorCode::ServerError(CONTENT_MODIFIED) => {}
                                        code => {
                                            let msg = match code {
                                                ErrorCode::MethodNotFound => format!(
                                                    "{} language server doesn't support method {}",
                                                    ctx.language_id, method
                                                ),
                                                _ => format!(
                                                    "{} language server error: {}",
                                                    ctx.language_id, editor_quote(&failure.error.message)
                                                ),
                                            };
                                            ctx.exec(meta, format!("lsp-show-error {}", editor_quote(&msg)));
                                        }
                                    };

```

This also would leave the handling of ` ErrorCode::ServerError(CONTENT_MODIFIED)` extensible in case we will want to log or retry it in the future.

jneem: Thanks for the quick feedback!

src/controller.rs
```diff
@@ -122,17 +127,23 @@ pub fn start(
...
-                                    ctx.exec(meta, format!("lsp-show-error {}", editor_quote(&msg)));
+                                    match failure.error.code {
+                                        ErrorCode::ServerError(CONTENT_MODIFIED) => {},
+                                        code => {
+                                            let msg = match code {
+                                                ErrorCode::MethodNotFound => format!(
+                                                    "{} language server doesn't support method {}",
+                                                    ctx.language_id, method
+                                                ),
+                                                ErrorCode::ServerError(CONTENT_MODIFIED) => String::new(),
```
ul: Sorry for nitpicking, this line is redundant now =)

jneem: Oops, sorry...

ul:
Thank you for your contribution! And for the inspiration to try ra_lsp_server =)

PR #287, opened 2020-02-09
Support incremental semantic highlighting

topisani:
`clangd` now implements incremental highlighting, so i implemented it here.

It stores the updated lines in the context, and then calls `lsp-update-semantic-highlighting` in the editor, which sends the current semhl option to `kak-lsp`. This is then parsed, and any regions on updated lines are removed, and the new regions are added.

This is made a lot easier than one would think, because the semhl is sent from the server one line at a time, so no fancy overlap checking is requried.

Tested a little bit with clangd, it appears to work, and correctly removes old regions.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -607,6 +607,19 @@ cat ${pipe}
 rm -rf ${tmp}
 }}
 
+define-command lsp-update-semantic-highlighting -hidden %{
+    nop %sh{ (printf '
+session      = "%s"
+client       = "%s"
+buffile      = "%s"
+filetype     = "%s"
+version      = %d
+method       = "update-semantic-highlighting"
+[params]
+current = "%s"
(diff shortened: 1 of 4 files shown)


ul:
Code looks good to me but I don't have a chance to test it extensively as I don't use any language server with semantic highlighting support. Let's merge and see if any feedback would come from others.

arrufat:
I've just updated my kak-lsp and I didn't notice any difference, how can I check it?

topisani:
It should work as before with any server that always updates the full buffer. Clangd is the only one I know of that does incremental highlighting, and I'm not sure that has been released yet.

This still seems to be quite stable, though I occasionally do get misplaced highlighting, but I believe that to be a sync issue, not a logic one

arrufat:
@topisani thanks for your answer. Actually, after some usage, I also got some misplaced highlighting (which I hadn't noticed before), and even running `lsp-update-semantic-highlighting` manually didn't fix it.

topisani:
Lsp-update-semantic-highlight won't fix it. It just applies the latest updated semhl, and if that has already happened, it shouldn't change anything, especially not for the better.

I've always gotten some misplaced highlights occasionally, but the async nature of this may have made it slightly worse, can't tell

arrufat:
Great, thanks for the clarification, I didn't notice the misplaced highlighting before, but probably because I want paying attentio!n. Thank you!

PR #295, opened 2020-03-06
kak-lsp.toml: Slight style change

alexherbo2:


ul:
Thanks for cleaning it up!

PR #300, opened 2020-03-20
Fix crash when language server sends out-of-bounds range

Phlogistique:
Currently, kak-lsp panics if a language server sends an out-of-bound range of characters.

This broke `lsp-formatting-sync` with `dhall-lsp-server`, which deliberately uses an out-of-bound value to convey that the whole file should be replaced:
https://github.com/dhall-lang/dhall-haskell/blob/431f8cfdd6af6c0a09a1a1f2f8056202b6e02a8c/dhall-lsp-server/src/Dhall/LSP/Handlers.hs#L307

Instead of panicking, the range is now clamped to the size of the file.

### DIFF
--- a/src/position.rs
+++ b/src/position.rs
@@ -26,7 +26,8 @@
 //!   And just works when `offset_encoding: utf-8` is provided in the config.
 use crate::types::*;
 use lsp_types::*;
-use ropey::Rope;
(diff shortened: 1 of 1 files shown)


ul:
Thank you for making kak-lsp more resilient =)

PR #301, opened 2020-03-20
Add help for troubleshooting crashes

Phlogistique:
The instructions in the README do not help if what you want to troubleshoot is a panic, or if you want to run a debugger. This provides an alternative way to run `kak-lsp` that makes troubleshooting more like usual daemon software.


ul:
Thanks for adding this bit!

PR #307, opened 2020-04-03
rust-analyzer inlay hints

tadeokondrak:
Requires https://github.com/mawww/kakoune/pull/3440.

To test:
- Use [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer)
- `hook global NormalIdle .* rust-analyzer-inlay-hints`
- `hook global InsertIdle .* rust-analyzer-inlay-hints`
- Wait (rust-analyzer takes a while before it starts sending them)

I'm not sure what the best way to make these automatically requested is.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -262,7 +262,7 @@ That is, the keys of the `lsp_server_configuration` option are a `.`-delimited p
 Inlay hints are a feature supported by https://github.com/rust-analyzer/rust-analyzer[rust-analyzer], which show inferred types, parameter names in function calls, and the types of chained calls inline in the code. To enable support for it in kak-lsp, add the following to your `kakrc`:
 
 ----
-hook global BufCreate .*\.rs %{
+hook global WinSetOption filetype=rust %{
   hook window -group rust-inlay-hints BufReload .* rust-analyzer-inlay-hints
   hook window -group rust-inlay-hints InsertIdle .* rust-analyzer-inlay-hints
   hook -once -always window WinSetOption filetype=.* %{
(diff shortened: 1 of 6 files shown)


ul:
The linked PR to Kakoune is open but https://github.com/mawww/kakoune/blob/master/doc/pages/changelog.asciidoc states "replace-ranges highlighter now support empty ranges". Was implemented in parallel?

tadeokondrak:
My implementation had issues with non-current timestamps, so it was implemented differently, but the new implementation doesn't work before spaces, so type hints on `let` doesn't work with it. Marking the PR as a draft until at least that's fixed.

tadeokondrak:
Looks like they are working now 🎉

tadeokondrak:
Okay, it works in kakoune master now, but enabling it still seems to be hacky (the way I posted sends that request every keystroke), and the colour might not be dim enough to stay out of the way.

ul:
I'd suggest defining a custom face for inlay hints, potentially defaulting to one of the built-in faces (like `References` defaulting to `MatchingChar` https://github.com/ul/kak-lsp/blob/master/rc/lsp.kak#L15). This way we could make it match people's themes and solve the problem of intrusive colours as everyone would be able to tune it to their liking.

Regarding integration, we can have it as is and then listen to feedback from the community about more efficient forms and eventually incorporate one of the solutions. Please add a section to the README describing the feature and how to enable it.

ul:
Thank you for this feature! It's one of the things I was jealous about when I looked into how Rust is supported in JetBrain's CLion.

PR #315, opened 2020-04-27
proposed Semantic Tokens support

tadeokondrak:
The successor to the previous proposed semantic highlighting API.

Tested with rust-analyzer, I think clangd also has support.

Test with `:lsp-semantic-tokens`. Like my other PR, I'm not sure the right way to run this automatically is.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -273,6 +273,22 @@ hook global WinSetOption filetype=rust %{
 
 You can change the face of the hints with `set-face global InlayHint <face>`.
 
+== Semantic Tokens
+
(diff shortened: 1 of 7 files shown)


ul:
Great to see more work on semantic highlighting! Would you mind adding a brief description to the  README?

It seems that we are getting a lot of requests which are expected to be hooked into `{Normal,Insert}Idle`. I wonder if we can add some kind of batching support in the future, so Kakoune sends only a single request which is then transformed by kak-lsp daemon into multiple requests to language server and stitched back into a single response to the editor.

PR #320, opened 2020-04-30
Inlay diagnostics support

tadeokondrak:
This PR adds "inlay diagnostics" (because the name "inline diagnostics" is already taken):

It works well sometimes and saves time hovering over lines for errors:

But it's unfortunately also quite buggy:
multiple at once
hovering over the newline it's replacing

I'm not sure if it should be merged, but I'm posting it here in case anyone has ideas on how to make it work better, or if anyone wants to use it anyway (I will), since hovering over lines is a big pain (in my opinion).


ul:
I actually like this kind of representation. If no one has objections to implementation, I think we should merge it.

tadeokondrak:
I have been using this since I opened the PR and it has been pretty nice, the main fixable issue is that multiple errors on the same line are not separated by spaces, I'll fix that then mark this as ready.

tadeokondrak:
Used commas, because they look better:

PR #321, opened 2020-05-01
Fix completion errors

bartol:
Hi, I have recently started using this awesome tool with gopls language server.
When I start typing completion menu pops up but if I try to insert item it outputs following error:
```
1:1: 'eval' 7:271: 'lsp-snippets-insert-completion' 1:2: 'eval' 4:79: 'eval' unknown option '-verbatim'
```
I am using kak-lsp version 8.0.0 and gopls version 0.4.0.
Removing `-verbatim` resolves errors and completion works as expected. Tested on both release 8.0.0 and master.


ul:
@topisani can tell more about snippets, but I assume that `-verbatim` is needed for this code to work properly with some edge cases. The option itself was introduced in 2019.12.10 release of Kakoune. What version do you use?

bartol:
Ah yes, it was my bad. I've just assumed that kakoune was on lastest version. Seems like versions from apt are old. Sorry!

PR #323, opened 2020-05-02
Fix incorrect scope for lsp_semantic_tokens

tadeokondrak:
Fixes https://github.com/ul/kak-lsp/issues/322


ul:
Thanks for a quick response!

Issue #275, opened 2020-01-07
ccls completion incorrect


Luis-Hebendanz:
Hello there,
I wanted to thank you for this nice addon! 

If ccls is used, and I select a completion item from the list while I am writing a function the whole function with its signature is being used not only the function name.

![image](https://user-images.githubusercontent.com/22085373/71911479-09e21400-3174-11ea-8325-42fe55264e50.png)

![image](https://user-images.githubusercontent.com/22085373/71911635-5af20800-3174-11ea-8e2d-404d8d8b63dd.png)

topisani:
Are you sure you are using an up to date `kak-lsp` build? i used to get this when the label was just used for insertion, but that behavior has been changed some time ago

Luis-Hebendanz:
```
$ kak-lsp --version 
kak-lsp 7.0.0
```

ul:
It seems to be the latest release version not master (which should report `kak-lsp 7.0.0-snapshot`). @topisani how recent is the change you are referring to? Do you refer your work on snippets?

The latest release tho should ask language server to provide non-snippet labels by default via the client capabilities. However, it might be the case that language server ignores it.

topisani:
I think this is related to the `insert_text` vs `label` usage. It might have been broken with my recent snippets support, but I wouldn't have thought so. It might also be something with ccls' `detailedLabel` setting, I think this does something to that extend...

arrufat:
I can confirm that it does not work as of https://github.com/ul/kak-lsp/commit/a7960510e7320cae5a980c1105660bfc900f779e.

The `detailedLabel` flag does not change that.

krobelus:
Duplicate of #40

krobelus:
we fall back to the label when we get a text edit that we can't handle yet
however, common cases are already supported by #326 
(also clangd doesn't seem to have this problem, probably it doesn't force us to use textEdit)

PR #326, opened 2020-05-04
Use textEdit completions that replace token left of cursor

krobelus:
This is a targeted fix for the cases where Kakoune insertion would do
exactly the same as the text edit.
If the language server sends simple textEdit objects that replace
precisely the token left of the cursor, then we can just pass them to
Kakoune as normal completions. This is strictly better than using the
completion's label, because Kakoune's completion engine does exactly
what the text edit is supposed to do.

See below for an example that is not fixed by this commit. I believe
it makes sense to have this partial support until #40 is properly
implemented, because it seems safe and fixes what is arguably the most
common completion scenario: completing the suffix of a token.

Reproduce with this example C file using ccls. Cursor is `|`.

	void foobar(){}
	void main(){
		fo|	// Fixed:	Completion inserts foobar()
		fo|ba	// Not fixed:	Completion inserts foobar() -> void
	}

I thought about also changing the behavior for the "Not fixed" case,
where we could also just pass on the completion.
The result would be that Kakoune replaces "fo" with "foobar()",
ignoring the existing suffix, so we'd end up with "foobar()ba".
This is better than the labels here, but I'm not sure if it's
a good idea to knowingly leak this error.

Improves on #275

### DIFF
--- a/src/language_features/completion.rs
+++ b/src/language_features/completion.rs
@@ -73,7 +73,21 @@ pub fn editor_completion(
                     .collect::<String>();
                 entry += &format!(" {{MenuInfo}}{:?}", k);
             }
-            let insert_text = &x.insert_text.unwrap_or(x.label);
+            // The generic textEdit property is not supported yet (#40).
+            // However, we can support simple text edits that only replace the token left of the
+            // cursor. Kakoune will do this very edit if we simply pass it the replacement string
+            // as completion.
+            let is_simple_text_edit = x.text_edit.as_ref().map_or(false, |text_edit|
+                text_edit.range.start.line + 1 == params.position.line &&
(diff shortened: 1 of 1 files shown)


ul:
Thank you for this improvement, unfortunately, I think #40 will open for a long time if no one will take care of it, as I'm in a period when I can't dedicate large enough timeslot to it.

tadeokondrak:
Hmm, I seem to be getting snippets sometimes on completion even when they're disabled: .
I'm not sure, but I think it might be caused by this PR.

krobelus:
@tadeokondrak that should not happen, how to reproduce? I'm not sure where to configure snippets for a Rust language server.
Possibly it can help us to start kak-lsp in verbose mode and extract the `DEBG From server:` that contain the offending `textEdit`s if any.

PR #330, opened 2020-05-14
Request plain-text content for hover and signature information

puckipedia:
This was missed two years ago, when fixed for the completion. This
ensures that e.g. gopls no longer tries to send us Markdown.

Overlooked for ... almost 2 years :) https://github.com/ul/kak-lsp/issues/69#issuecomment-404334258


ul:
Great archeology skills =) Thank you for your contribution!

PR #332, opened 2020-05-19
doc: mention that lsp-references is mapped to gr by default

Delapouite:
Hi

When executing `lsp-capabilities`, the mapping for `lsp-definition` appears in the infobox but the one for `lsp-references` is missing


ul:
Thank you!

PR #333, opened 2020-05-19
doc: mention *-previous-match and *-next-match commands

Delapouite:
Hello

In Kakoune commands like `grep-previous-match` or `grep-next-match`
(and their `make-*`) equivalent are hardcoded to search for a buffer
named `*grep*` or `*make*`.

Unfortunately it means that they can't be used for buffers with meaningful
names like `*references*`. So I tried to clarify this point.

In #169 dedicated commands where added for references and symbols,
I've added them explicitly.

In a future PR, for consistency, we may want to add the same for
`*implementations*` and `*diagnostics*`. Beware, there's a nuance with
the last one as it is a `make` not a `grep` buffer.


ul:
Thank you for keeping docs healthy =)

PR #335, opened 2020-05-19
Support custom command rust-analyzer.applySourceChange

tadeokondrak:
Haven't tested it much yet. rust-analyzer's actions are *super* useful, so this is a pretty big improvement for me at least :)

Closes https://github.com/ul/kak-lsp/issues/314

### DIFF
--- a/src/language_features/rust_analyzer.rs
+++ b/src/language_features/rust_analyzer.rs
@@ -1,10 +1,16 @@
 use crate::context::Context;
-use crate::position::lsp_range_to_kakoune;
-use crate::types::{EditorMeta, EditorParams, KakounePosition, KakouneRange};
-use crate::util::editor_quote;
+use crate::position::{lsp_position_to_kakoune, lsp_range_to_kakoune};
+use crate::types::{EditorMeta, EditorParams, KakounePosition};
+use crate::util::{apply_text_edits, editor_quote};
+use crate::workspace;
 use lsp_types::request::Request;
-use lsp_types::{Range, TextDocumentIdentifier};
+use lsp_types::ExecuteCommandParams;
+use lsp_types::InsertTextFormat;
+use lsp_types::TextEdit;
+use lsp_types::VersionedTextDocumentIdentifier;
+use lsp_types::{Range, ResourceOp, TextDocumentIdentifier, TextDocumentPositionParams};
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use url::Url;
 
 pub enum InlayHints {}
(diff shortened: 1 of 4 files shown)


src/language_features/rust_analyzer.rs
```diff
@@ -101,3 +107,89 @@ pub fn inlay_hints_response(meta: EditorMeta, inlay_hints: Vec<InlayHint>, ctx:
...
+        .into_iter()
+        .nth(0)
+        .expect("Missing source change");
+    let SourceChange {
+        workspace_edit:
+            SnippetWorkspaceEdit {
+                changes,
+                document_changes,
+            },
+        cursor_position: _,
```
tadeokondrak, line 163: Oops, forgot about this

tadeokondrak:
Ready, and even fixed a bug I added on the way :)

src/workspace.rs
```diff
@@ -151,77 +224,11 @@ pub fn apply_edit(
...
-                                    }
-                                } else if path.is_file() && fs::remove_file(&path).is_err() {
-                                    error!(
-                                        "Failed to delete file: {}",
-                                        path.to_str().unwrap_or("")
-                                    );
-                                    applied = true;
-                                }
+                        DocumentChangeOperation::Op(op) => {
+                            if apply_document_resource_op(&meta, op, ctx) {
```
ul, line 228: Should it be `applied = apply_document_resource_op(&meta, op, ctx)`? Because default value for `applied` is true (line 213) and if `apply_document_resource_op` would return `false`, `applied` will stay `true` in the current implementation.

tadeokondrak: From what I could tell, `applied` does nothing right now and is just there for future use? I don't think it actually does anything yet.

ul: We send it back to the language server:

```
pub fn apply_edit_from_server(id: Id, params: Params, ctx: &mut Context) {
    let params: ApplyWorkspaceEditParams = params.parse().expect("Failed to parse params");
    let meta = ctx.meta_for_session();
    let response = apply_edit(meta, params.edit, ctx);
    ctx.reply(id, Ok(serde_json::to_value(response).unwrap()));
}
```

Even if langserv does nothing with it we probably should provide correct value.

tadeokondrak: Sorry, what I mean is that in the current code, it's never set to false. It's only set to true at the start and true in every assignment.

ul: Ah, now I get it. I think it's a long-running bug then, it's probably meant to be `false` by default.

ul: Okay, there is a problem with this `applied` but it's not resolved by setting it to the `false` by default. Let's solve it later.

PR #339, opened 2020-05-20
Improve failure handling in workspace::apply_edit

tadeokondrak:
It compiles, but I haven't tested it yet.

### DIFF
--- a/src/general.rs
+++ b/src/general.rs
@@ -31,7 +31,7 @@ pub fn initialize(
                         ResourceOperationKind::Delete,
                         ResourceOperationKind::Rename,
                     ]),
-                    failure_handling: None,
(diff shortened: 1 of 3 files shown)


tadeokondrak:
Not really sure how to check if it sends the right reply (since I don't know any language server other than rust-analyzer that does file operations, and it uses applySourceChange), but it still seems to work.

ul:
@topisani would you mind testing it with clangd (iirc you use it actively)?

topisani:
Yep, gimme just a few days, I'll try it out

ul:
I guess we can merge and see how it works in the wild. I don't see any immediate problems with the code or after trying locally.

PR #343, opened 2020-05-22
Generalize handling of GotoDefinitionResponse and add lsp-type-definition

tadeokondrak:
There were a bunch of places that all handled GotoDefinitionResponse in their own way, and this meant, for example, :lsp-definition couldn't go to multiple `impl` blocks for Rust. This PR merges them into one implementation, and also adds lsp-type-definition.


ul:
It worth updating README as well to reflect the fact that buffer is called `*goto*` now for all those commands.

PR #345, opened 2020-05-23
Update default haskell config to use `hie-wrapper`

414owen:
[haskell-ide-engine](https://github.com/haskell/haskell-ide-engine/) added `hie-wrapper` two years ago, to identify the right ghc version to use.


ul:
Thank you for keeping default config up-to-date!

PR #346, opened 2020-05-23
Rust 1.41.0 fixups

414owen:
Ran `cargo test`. These fixups were needed for `master` to compile.
Not 100% sure why type inference didn't work on these literals.
Needed to use u64 to get access to u64::MAX.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -7,6 +7,7 @@ use jsonrpc_core::Params;
 use lsp_types::*;
 use std::collections::HashSet;
 use std::path::Path;
+use std::u64;
 
 pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
     let params: PublishDiagnosticsParams = params.parse().expect("Failed to parse params");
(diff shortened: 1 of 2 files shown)


ul:
We don't have a clear policy around the Rust version, but I think it'd be reasonable to say that supporting the latest is enough. I may be wrong as package maintainers may need a wider range of versions. What do you think?

I believe this is an announcement relevant to changes in this PR: 
https://blog.rust-lang.org/2020/04/23/Rust-1.43.0.html#type-inference-around-primitives
https://blog.rust-lang.org/2020/04/23/Rust-1.43.0.html#library-changes

414owen:
Ah it's a real shame that's the version number that breaks.
I'm on nixos, so `rustup` isn't really an option, because my dynamic libraries are in nix store locations, and the latest `rustc` in `<nixpkgs-unstable>` is `1.42.0` :)

Feel free to close this.

ul:
Well, there is no harm in being backward-compatible. Would you mind resolving conflicts? Let's merge it after that.

414owen:
@ul appreciate it.

Issue #318, opened 2020-04-29
Batching commands triggered by {Normal,Insert}Idle


ul:
It seems that we are getting more and more of requests which are expected to be hooked into `{Normal,Insert}Idle` (buffer change, semantic highlighting, inlay hints, auto-hover...). I wonder if we can add some kind of batching support, so Kakoune sends only a single request which is then transformed by kak-lsp daemon into multiple requests to language server and stitched back into a single response to the editor.

topisani:
We should also be able to set higher idle intervals for all/some of these requests. Thats probably best done on the kak-lsp side.

414owen:
@ul would we want to stitch them back into a single response for the editor?
Some LS features might be more computationally expensive than others.
If we stitch them back together, we have to wait for the slowest operation.

ul:
It seems I was too fast in PR comment assuming we want stitching responses in this use case. I don't see any significant benefits in doing it. At the same time the drawback of waiting for the slowest response is quite huge.

PR #347, opened 2020-05-24
Add selection formatting

414owen:
Adds support for `textDocument/rangeFormatting`, which formats the current selections.

Tested with `haskell-language-server`.

~Because the `language-server-protocol` doesn't support batching, we're using one message per selection. This means we also get a history item per selection :shrug:.~


ul:
I think we can collect responses and send them back as a single Kakoune command. I don't know how hard is it to do with the current kak-lsp architecture but we definitely need to be able to do it at some point, for example for #318 (tho that one is more about batching requests, batching their responses could be helpful too).

Would you mind spending a bit of time investigating this possibility? Both negative and positive outcomes would be valuable!

414owen:
@ul sure thing. I'll look into it.

414owen:
@ul This has been updated to squash selection formats using the new batch call infrastructure.

PR #348, opened 2020-05-25
fix diagnostics in hover when lsp_hover_max_lines > 0

mlvzk:
I noticed when I had set the lsp_hover_max_lines option my diagnostics in hover were not showing, this change fixed it for me.


ul:
Good catch, thank you!

PR #349, opened 2020-05-25
Give preference to roots according to their position in configuration list

wz1000:
Tweak the root finding logic to search for roots in order of their appearance in
the list.

This is useful when you have subprojects underneath your main project directory,
but would like to set the root to the parent project directory.

For example, we might have nested haskell subprojects like

```
.
├── project.cabal
├── hie.yaml
├── src
│  └── ...
├── subproject
│  ├── subproject.cabal
│  └── ...
└── ..
```

Even though there is a `subproject.cabal` file, we want to give preference to a
`hie.yaml` file which, if it exists, tells the language server exactly how to
load the project. This is not the case with a `*.cabal` file, where the language
server has to do some guesswork (which might fail).

With this patch, we can achieve the desired behaviour by setting the haskell
configuration roots as:

```yaml
[language.haskell]
roots = ["hie.yaml","stack.yaml","*.cabal","Setup.hs"]
```

### DIFF
--- a/src/project_root.rs
+++ b/src/project_root.rs
@@ -13,14 +13,14 @@ pub fn find_project_root(language: &str, markers: &[String], path: &str) -> Stri
 }
 
 pub fn roots_by_marker(roots: &[String], path: &str) -> String {
-    let mut pwd = PathBuf::from(path);
-    while !pwd.is_dir() {
-        pwd.pop();
+    let mut src = PathBuf::from(path);
+    while !src.is_dir() {
+        src.pop();
     }
-    let src = pwd.to_str().unwrap().to_string();
 
-    loop {
-        for root in roots {
+    for root in roots {
+        let mut pwd = src.clone();
+        loop {
             // unwrap should be safe here because we walk up path previously converted from str
             let matches = glob(pwd.join(root).to_str().unwrap());
             if let Ok(mut m) = matches {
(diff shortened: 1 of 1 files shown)


src/project_root.rs
```diff
@@ -17,10 +17,11 @@ pub fn roots_by_marker(roots: &[String], path: &str) -> String {
     while !pwd.is_dir() {
         pwd.pop();
     }
-    let src = pwd.to_str().unwrap().to_string();
+    let src = pwd.clone();
 
-    loop {
-        for root in roots {
+    for root in roots {
+        pwd = src.clone();
```
ul, line 23: nit: I think we can get rid of extra variable and one extra clone:

```diff
@@ -17,10 +17,9 @@ pub fn roots_by_marker(roots: &[String], path: &str) -> String {
     while !pwd.is_dir() {
         pwd.pop();
     }
-    let src = pwd.clone();
 
     for root in roots {
-        pwd = src.clone();
+        let mut pwd = pwd.clone();
         loop {
             // unwrap should be safe here because we walk up path previously converted from str
             let matches = glob(pwd.join(root).to_str().unwrap());
@@ -31,11 +30,11 @@ pub fn roots_by_marker(roots: &[String], path: &str) -> String {
                 }
             }
             if !pwd.pop() {
-              break;
+                break;
             }
         }
     }
-    return src.to_str().unwrap().to_string();
+    return pwd.to_str().unwrap().to_string();
 }
```

If find shadowing confusing here, we can choose a different name for the `pwd` in the loop.

PR #350, opened 2020-05-25
Underline document highlight writes

wz1000:


ul:
Neat! Thanks for adding it.

PR #351, opened 2020-05-26
Fix inlay diagnostics for gopls

tadeokondrak:
Also adds a panic hook, since the integer overflow crashed the process without any information in the log.


tadeokondrak:
Hm, I marked it as a draft because it had an issue with diagnostics that contain `'`, but I'm not sure how to fix this and it doesn't seem related, so opening it back up.

tadeokondrak:
Actually this does seem to make it buggier than it used to be, and I'm not exactly sure why.

PR #353, opened 2020-05-26
LS response bucketing

414owen:
This adds a context method, `batch_call`, which you can pass a `Vec` of language-server requests, and a callback, which is passed an accumulated list of the responses.

This allows us to, for example, flatten responses, so we have a single history item in kakoune, as we want in https://github.com/ul/kak-lsp/pull/347

Note that the responses are not necessarily in the same order as the requests.
I couldn't think of a use case for that, so KISS.
If anyone needs this behavior, it's trivial to implement.


ul:
Glad to see it didn't require rewriting too much =)

PR #362, opened 2020-06-15
Properly unescape = in key-value pair

JJK96:
Unescapes `test=foo\=bar` to `key: test` and `value: foo=bar`.


ul:
Good catch, thank you!

PR #368, opened 2020-06-29
Remove bash dependency by replacing here-string '<<<' with here-doc '<<'

1g0rb0hm:
The following pull request removes the dependency on **bash** by using **here-doc** instead of **here-string** as discussed [here](https://github.com/ul/kak-lsp/commit/23e229eb1ba649697af0c9895001cb161e3d4d04#commitcomment-40231590).

The fix has been tested on **MacOS** using `/bin/dash` as `KAKOUNE_POSIX_SHELL`.


ul:

PR #375, opened 2020-07-20
Rust use component for active toolchain

pickfire:
Use rustup to figure out what is the active toolchain and use that binary
Added rust-analyzer but commented as it is only available in nightly
by using rust-analyzer-preview

### DIFF
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -42,13 +42,13 @@ timeout = 1800 # seconds = 30 minutes
 filetypes = ["rust"]
 roots = ["Cargo.toml"]
 command = "sh"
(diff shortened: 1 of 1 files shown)


krobelus:
I wonder if we should add a fallback if `rustup` is not installed. I think that is the case when a user installs distro-packaged Rust.

ul:
Thank you, I never thought about such wrapper as I try hard to stick with a single toolchain (the latest stable) for my projects but I guess it comes handy for many others.

pickfire:
Ah, I usually just stick with a single toolchain but sometimes I need to use nightly on certain quick test to have `benches` working by using `echo nightly > rust-toolchain` so I thin this would be useful.

PR #379, opened 2020-08-01
Fix panics on out-of-range codepoints

puckipedia:
In some cases (presumably caused by the LSP asynchronously responding, while editing text) it seems that `kak-lsp` can get confused, causing it to try to index into out-of-bounds ranges. This is
easily reproducible by taking e.g. `rust-analyzer`, opening a rust file, and just mashing on the keyboard. (especially editing `use` lines causes this, in my experience.) This partially resolves this by catching these out-of-range positions and translating them to similarly invalid ranges on the other end.

This isn't a comprehensive fix (I believe the UTF-8 codepoint handling is still partially broken),
but this seems to resolve a decent amount of crashes that I have personally encountered
using `rls`, `rust-analyzer`, and `gopls`. Properly bailing out would be preferred, but this would require modifying a large amount of code surrounding this code.

### DIFF
--- a/src/position.rs
+++ b/src/position.rs
@@ -178,11 +178,19 @@ fn lsp_range_to_kakoune_utf_8_code_units(range: &Range) -> KakouneRange {
 
 fn kakoune_position_to_lsp_utf_8_code_points(position: &KakounePosition, text: &Rope) -> Position {
     // -1 because LSP & Rope ranges are 0-based, but Kakoune's are 1-based.
-    let line = position.line - 1;
-    let character = text
-        .line(line as _)
-        .byte_to_char((position.column - 1) as _) as _;
-    Position { line, character }
+    let line_idx = position.line - 1;
+    let col_idx = position.column - 1;
+    if line_idx as usize >= text.len_lines() {
+        return Position { line: line_idx, character: col_idx };
+    }
+
+    let line = text.line(line_idx as _);
+    if col_idx as usize >= line.len_bytes() {
+        return Position { line: line_idx, character: col_idx };
+    }
+
+    let character = line.byte_to_char(col_idx as _) as _;
+    Position { line: line_idx, character }
 }
 
 fn kakoune_position_to_lsp_utf_8_code_units(position: &KakounePosition) -> Position {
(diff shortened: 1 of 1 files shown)


src/position.rs
```diff
@@ -178,11 +178,20 @@ fn lsp_range_to_kakoune_utf_8_code_units(range: &Range) -> KakouneRange {
...
-    Position { line, character }
+    let line_idx = position.line - 1;
+    if line_idx as usize >= text.len_lines() {
+        return Position { line: line_idx, character: position.column - 1 };
+    }
+
+    let line = text.line(line_idx as _);
+
+    if (position.column - 1) as usize >= line.len_bytes() {
+        return Position { line: line_idx, character: position.column - 1 };
```
krobelus, line 189: I'm having similar problems, could it be a timing issue where the language server works with an outdated document? The coordinates are out of range, I think that also happens when only using ASCII characters.

The two early returns are edge cases where we may get incorrect results. Does it make sense to log a warning here? Because normally this case should not happen.

I would extract `position.column - 1` to a variable.

puckipedia: > I'm having similar problems, could it be a timing issue where the language server works with an outdated document? The coordinates are out of range, I think that also happens when only using ASCII characters.

As mentioned in the PR description, I am pretty certain it is this. I think this can happen with non-ASCII too, but I don't have much unicode in my document.

> The two early returns are edge cases where we may get incorrect results. Does it make sense to log a warning here? Because normally this case should not happen.

No, because it's not really an "edge case" per se; it can happen very trivially by just pressing backspace while the LSP is thinking; logging a warning would be way too verbose. IMO this function should've probably returned a `Result`, but ensuring those make it further upstream is a lot of effort, relatively

> I would extract `position.column - 1` to a variable.

Reasonable; this is pretty much the patch i wrote locally when I got annoyed at it repeatedly breaking, and I wanted to at least get some people's comments on this problem :)

ul:
Thank you, I think it's a reasonable workaround until we get a more explicit error handling for the doc and lsp going out of sync.

PR #381, opened 2020-08-05
kak-lsp.toml: Default to clangd instead of cquery.

### LINKED ISSUE #380: Change default language server for C/C++
Today, the default language server for C and C++ is cquery, but the cquery [repository](https://github.com/jacobdufault/cquery) is archived, and the README has a message at the top:

> cquery is no longer under development. clangd and ccls are both good replacements.

ccls might be a good new default, since apparently it's a fork of cquery.

clangd would also be a good new default, because it's an official LLVM project, presumably with the financial support that implies, and should be available everywhere that clang is.

Screwtapello:
cquery development has been abandoned, and clangd is both widely available (as part of clang) and has the resources of the LLVM project behind it.

Fixes #380.

### DIFF
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -105,9 +105,8 @@ offset_encoding = "utf-8"
 
 [language.c_cpp]
 filetypes = ["c", "cpp"]
-roots = ["compile_commands.json", ".cquery"]
-command = "cquery"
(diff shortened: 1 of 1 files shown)


ul:
Thank you for keeping default config up-to-date!

PR #385, opened 2020-08-17
bring tools buffers to front in context of toolsclient

vbauerster:
Same fix as described here: https://github.com/mawww/kakoune/pull/3656


ul:
Thank you!

Issue #235, opened 2019-09-04
Flag-lines highlighter is unstable when there is only one line in document


alekfed:
When I start new documents and there are no errors and warnings, there is no flag line on the left, but as I enter new characters this line appears and disappears on every inserted character so the whole window start to jitter until I start a new line. If I first insert an empty line in the document and start to enter characters on the second line, flag line will appear and won't disappear anymore, so that's the way I usually work - from the second line (and deleting that empty line after inserting some text).

Perhaps, it's not an issue with kak-lsp itself, but it's especially apparent with kak-lsp.

Is there any other solution of this problem instead of "always start from the second line"?

PR #386, opened 2020-08-17
don't occupy space in gutter if there are no any errors

vbauerster:
If there are no any errors `lsp_error_lines` still occupies one space making left gutter unnecessarily wider.
I think it should behave same as [git-diff](https://github.com/mawww/kakoune/blob/2c437cfa02ca568750182209ef6249c6975243a5/rc/tools/git.kak#L31), occupy space only if there is something to show.


vbauerster:
Closing, as it will introduce jitter behavior described in #235

PR #389, opened 2020-08-23
save the " register when showing diagnostics and other buffers

BurNiinTRee:
Currently showing diagnostics and other actions override the default yank/copy/delete register.
This fixes that


ul:
Thank you for fixing this!

PR #393, opened 2020-09-04
Add support for Nix

deviant:
This adds support for Nix files, by way of [rnix-lsp](https://github.com/nix-community/rnix-lsp). I also took the opportunity to sort the existing language config blocks by alphabetical order, since that was becoming increasingly arbitrary.


ul:
Thank you for the config update!

PR #396, opened 2020-09-07
Update semantic tokens to LSP 3.16

topisani:
Updates lsp-types to 0.80.0
Works with clangd 12 (latest master)


ul:
Thank you for the update!

PR #397, opened 2020-09-08
clangd-switch-source-header

topisani:
Implement `textDocument/switchHeaderSource` as specified by https://clangd.llvm.org/extensions.html#switch-between-sourceheader


ul:
Cheers!

PR #400, opened 2020-09-22
Add official terraform language server

xenrox:
At the moment there are two language server implementations. 
I would pick the [offical](https://github.com/hashicorp/terraform-ls) one, since it will work with more terraform versions.


ul:

PR #402, opened 2020-09-24
Escape inlay diagnostics that contain double quotes

### LINKED ISSUE #352: Inlay diagnostics with ' don't display correctly
kak-lsp sends `eval "set buffer lsp_diagnostics 50 ''9.16+0|{DiagnosticError}{\} "encoding/json" imported but not used''"` to the client, but the quotes don't end up matching right and it errors out.

krobelus:
Commit 155b21f (Inlay diagnostics support) copies an eval which is used
to interpret %opt[] tags in line_flags.  It's not necessary here;
diagnostic_ranges is a range-specs which we feed the literal position, some
markup and the diagnostic message from the language server. Clarify that in
the comment. No test, but reproducible with gopls:

	printf 'package main\nimport "encoding/json"' > test.go
	kak -e 'lsp-inlay-diagnostics-enable global; lsp-enable' test.go

Fixes #352

@ul I don't think I have write access yet, otherwise I'd push directly to master if I'm confident in a change.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -87,13 +87,13 @@ pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
         .join(" ");
     // Always show a space on line one if no other highlighter is there,
     // to make sure the column always has the right width
-    // Also wrap it in another eval and quotes, to make sure the %opt[] tags are expanded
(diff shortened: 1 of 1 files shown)


ul:
Sorry for that, I thought org members get write access to repos automatically. I updated the base role in the org to Admin now.

Issue #370, opened 2020-07-06
Dart code completion injects spurious `${1:}`


haarts:
Since upgrading from 7.0.0 to 8.0.0 whenever I code complete the characters `${1:}` are appended. 
![example](https://user-images.githubusercontent.com/10660/86588621-5627d880-bf8c-11ea-9975-f4cfe4a79c1b.png)

I do not have the same issue with Rust code. 

Kakoune version v2020.01.16-78-g3ef8a616
kak-lsp version 8.0.0
Dart VM version: 2.8.4 (stable)

haarts:
I'm dying to have this resolved. How can I help with this?

tw4452852:
I've encountered the same issue. After some investigation, the culprit is here:
https://github.com/ul/kak-lsp/blob/4aaa97260847f40da4da4407f132fb3dce41e247/rc/lsp.kak#L1458

It doesn't recognize the placeholder syntax provided by the dart language server, so I just did a hack for this, replacing the line above with this: `exec s((?<lt>!\\)(\\\\)*|\A)\K(\$(\d+|\{(\d+(:(\\\}|[^}])?)?)\}))<ret>`

PR #403, opened 2020-09-24
Add support for gopls and Dart lsp placeholders

### LINKED ISSUE #390: Some explanations about using placeholders
Hi! First of all, thanks for this amazing work!

I see some time ago support for snippets where introduced in PR https://github.com/ul/kak-lsp/pull/270 by @topisani. Nevertheless, there's still no mention of it in the documentation (either `README.md` or the wiki), and I couldn't figure out how it is supposed to work. When I use it with [gopls](https://github.com/golang/tools/tree/master/gopls), some placeholders appear when I choose a completion suggestion, but they are not selected, and then I have to manually delete them before moving on.

I'm gonna make a small example. Suppose I have the following code:

```go
package main

type teste struct {
    um int
    umas string
}

func main() {
    t := teste{
        u|
    }
}
```
If my cursor is at the `|` position an I'm typping, the completion suggests both `um` and `umas` fields. If I then press `<c-n>` to select any of them (say, `umas`), the text `${1:},` is inserted after the completion but is not selected:

```go
    t := teste{
        umas: ${1:},|
    }
````
The cursor is positioned after the comma and the `${1:}` is not selected, making me having to manually erase everything between `umas:` and the cursor. Am I missing something?

gustavo-hms:
Hi! I've made this change in order to fix #390 and #370. The new regex still matches all the current placeholders' formats and additionally matches `gopls` style placeholders (`${1:}`).

Please tell me if everything is OK.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -1455,10 +1455,11 @@ def lsp-snippets-insert -hidden -params 1 %[
             # select things that look like placeholders
             # this regex is not as bad as it looks
             eval -draft %[
-                exec s((?<lt>!\\)(\\\\)*|\A)\K(\$(\d+|\{(\d+(:(\\\}|[^}])+)?)\}))<ret>
+                exec s((?<lt>!\\)(\\\\)*|\A)\K(\$(\d+|\{(\d+(:(\\\}|[^}])*)?)\}))<ret>
                 # tests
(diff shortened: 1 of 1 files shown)


krobelus:
Look great, thanks for this fix!

PR #418, opened 2020-11-23
Fix docs on semantic token modifier config

ifreund:
https://github.com/kak-lsp/kak-lsp/blob/2e1f9b62df71d2a641e0a8b33c30df15da8478d3/src/types.rs#L28


krobelus:
Thanks!

PR #419, opened 2020-11-28
make use of eval -verbatim and cosmetic patch

vbauerster:
This patch makes use of `eval -verbatim` so there is no need to call [editor_quote](https://github.com/kak-lsp/kak-lsp/blob/07f092ed37712534c7ac3fc2117b54da7a9d94cd/src/util.rs#L114) for subsequent command. Side effect is some performance gain, as no extra allocation is done for each `to_editor` command call.

I've tested mostly with `rust-analyzer` and found no regression or any errors.

### DIFF
--- a/src/editor_transport.rs
+++ b/src/editor_transport.rs
@@ -102,14 +102,13 @@ pub fn start(config: &Config, initial_request: Option<String>) -> Result<EditorT
 }
 
 pub fn start_unix(path: &path::PathBuf, sender: Sender<EditorRequest>) {
-    let listener = UnixListener::bind(&path);
-
-    if listener.is_err() {
-        error!("Failed to bind {}", path.to_str().unwrap());
-        return;
-    }
-
-    let listener = listener.unwrap();
+    let listener = match UnixListener::bind(&path) {
+        Ok(listener) => listener,
+        Err(e) => {
+            error!("Failed to bind: {}", e);
+            return;
+        }
+    };
 
     for stream in listener.incoming() {
         match stream {
--- a/src/editor_transport.rs
+++ b/src/editor_transport.rs
@@ -65,27 +65,25 @@ pub fn start(config: &Config, initial_request: Option<String>) -> Result<EditorT
                     .spawn()
                 {
                     Ok(mut child) => {
-                        {
-                            let stdin = child.stdin.as_mut();
-                            if stdin.is_none() {
+                        let stdin = match child.stdin.as_mut() {
+                            Some(stdin) => stdin,
+                            None => {
                                 error!("Failed to get editor stdin");
                                 return;
                             }
-                            let stdin = stdin.unwrap();
-                            let client = response.meta.client.as_ref();
-                            let command = if client.is_some() && !client.unwrap().is_empty() {
-                                format!(
-                                    "eval -client {} {}",
-                                    client.unwrap(),
-                                    editor_quote(&response.command)
-                                )
-                            } else {
-                                response.command.to_string()
-                            };
-                            debug!("To editor `{}`: {}", response.meta.session, command);
-                            if stdin.write_all(command.as_bytes()).is_err() {
-                                error!("Failed to write to editor stdin");
+                        };
+
+                        let client = response.meta.client.as_ref();
+                        let command = match client.filter(|&s| !s.is_empty()) {
+                            Some(client) => {
+                                format!("eval -client {} -verbatim -- {}", client, response.command)
                             }
+                            None => response.command.to_string(),
+                        };
+
+                        debug!("To editor `{}`: {}", response.meta.session, command);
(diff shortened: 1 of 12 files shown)


src/language_features/semantic_highlighting.rs
```diff
@@ -113,18 +111,11 @@ pub fn make_scope_map(ctx: &mut Context) -> std::vec::Vec<std::string::String> {
...
-        .and_then(|x| x.scopes.as_ref());
-
-    if scopes.is_none() {
-        return Vec::new();
-    }
-    let scopes = scopes.unwrap();
-
-    map_scopes_to_faces(scopes, faces)
+        .and_then(|x| x.scopes.as_ref())
+        .map_or(Vec::new(), |v| map_scopes_to_faces(v, faces))
```
krobelus, line 118: If `Vec::new()` was expensive, we should use `map_or_else`, but it's not so this is perfectly fine.

krobelus (approved):
Thanks a lot for these patches! The `-verbatim` change makes a ton of sense, and I fully agree with the cosmetic changes.
I've left some comments, mostly for myself; I didn't see anything that needs to be changed.

src/editor_transport.rs
```diff
@@ -65,27 +65,25 @@ pub fn start(config: &Config, initial_request: Option<String>) -> Result<EditorT
                     .spawn()
                 {
                     Ok(mut child) => {
-                        {
```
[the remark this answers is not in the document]
vbauerster: > Maybe earlier Rust didn't allow to use `stdin` and `child.wait()` in the same scope since both use mutable references?

I think yes it didn't, moreover `child.wait()` is dropping `stdin`. At least [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d0eb3a32ece5ac6bdfad6036204faa81) playground didn't compile with rust 1.32. Here is [related](https://github.com/PacktPublishing/Mastering-RUST-Second-Edition/issues/1) question.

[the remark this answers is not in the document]
vbauerster: I can add extra scope for backward compatibility.

src/language_features/clangd.rs
```diff
@@ -22,7 +22,7 @@ pub fn switch_source_header(meta: EditorMeta, ctx: &mut Context) {
         move |ctx: &mut Context, meta, response| match response {
             Some(response) => {
               let command = format!(
-                "evaluate-commands -try-client %opt{{jumpclient}} %{{edit {}}}",
+                "eval -try-client %opt{{jumpclient}} -verbatim -- edit -existing {}",
```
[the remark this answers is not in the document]
vbauerster: As far as I remember, it's used inconsistently across the code base. I agree that it's nice to have a rule for that, but better do it in another PR, harmonizing all usages.

src/language_features/semantic_highlighting.rs
```diff
@@ -91,18 +91,16 @@ pub fn editor_update(meta: EditorMeta, params: EditorParams, ctx: &mut Context)
...
-        if sh.scopes.is_some() {
-            let command = format!("echo -debug %§{:?}§", sh.scopes.as_ref().unwrap());
-            ctx.exec(meta, command.to_string());
-        }
-    });
+        .and_then(|x| x.semantic_highlighting.as_ref())
+        .and_then(|x| x.scopes.as_ref());
+
+    if let Some(scopes) = scopes {
+        let command = format!("echo -debug %§{:?}§", scopes);
```
vbauerster, line 101: I'm not sure if logic to debug scopes here is correct, because invoking `lsp-semantic-available-scopes` as suggested [here](https://github.com/kak-lsp/kak-lsp/blob/34e888eea109ad1b2fbaee3b0b7afab46b38f1c8/kak-lsp.toml#L10) never printed anything to the debug buffer at least with rust-analyzer and gopls.

topisani: Don't they use the new semantic tokens api? cause the available scopes is for the old semantic highlighting api iirc

vbauerster: > Don't they use the new semantic tokens api?

Can't answer, just wanted to point out that calling `lsp-semantic-available-scopes` doesn't do anything useful.
Also I can't tell what is difference between `[semantic_scopes]` and `[semantic_tokens]`. Former is referring to texmate scopes.

src/editor_transport.rs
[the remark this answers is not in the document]
krobelus: Thanks for clarifying, so it's really because of "non-lexical lifetimes". No need to add the scope IMO, because the behavior is not changed.

src/language_features/semantic_highlighting.rs
topisani: Yeah it's probably lacking some documentation...

PR #440, opened 2021-01-22
Elaborate/clarify on `lsp` user mode

th1j5:
For me, it is(/was) not clear that the [LSP user mode](https://github.com/kak-lsp/kak-lsp/blob/master/README.asciidoc#usage) didn't have a key combination yet, but is declared.
The linking to kakoune docs didn't totally clarify everything, I'll raise another issue there to see how the [key mapping of user modes](https://github.com/mawww/kakoune/blob/master/doc/pages/modes.asciidoc#user-modes) should be done exactly.

(Was writing an issue, but proposed directly the little change instead)

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -164,7 +164,7 @@ even if Kakoune session is still up and running. Change `server.timeout` in `kak
 duration, or set it to 0 to disable this behaviour. In any scenario making new request would lead to
 attempt to spin up server if it is down.
 
-* `lsp` user mode (see https://github.com/mawww/kakoune/blob/master/doc/pages/modes.asciidoc#user-modes[Kakoune docs] for more details about user modes):
+* The `lsp` user mode is declared, but has no keybinding yet (see https://github.com/mawww/kakoune/blob/master/doc/pages/modes.asciidoc#user-modes[Kakoune docs] for more details about user modes):
 
 |===
 | Binding | Command


README.asciidoc
```diff
@@ -164,7 +164,7 @@ even if Kakoune session is still up and running. Change `server.timeout` in `kak
 duration, or set it to 0 to disable this behaviour. In any scenario making new request would lead to
 attempt to spin up server if it is down.
 
-* `lsp` user mode (see https://github.com/mawww/kakoune/blob/master/doc/pages/modes.asciidoc#user-modes[Kakoune docs] for more details about user modes):
+* The `lsp` user mode is declared, but has no keybinding yet (see https://github.com/mawww/kakoune/blob/master/doc/pages/modes.asciidoc#user-modes[Kakoune docs] for more details about user modes):
```
krobelus, line 167: The convention in Kakoune is that plugins never add keybindings, except in user modes they own.
(our `gd` and `gr` are exceptions). So this should be implied.

Probably most users want to have a binding, so maybe we should explicitly say something like

- You may want to add a binding for `%{: enter-user-mode lsp<ret>}`.

or
- Suggested binding: `map global user l %{: enter-user-mode lsp<ret>}`

In general, this readme is dense reference documentation, but some tutorial bits like this one shouldn't hurt?

th1j5: This sounds good/maybe even better to me!
The main point is that I (as pretty new to kakoune) was searching for a long time about: how do I use this mode, which keys should I use to get in there, ...
The convention you're talking about was not totally clear for me (and I think this should maybe also be clarified in the [kakoune docs](https://github.com/mawww/kakoune/blob/master/doc/pages/modes.asciidoc#user-modes) - not yet created an issue)

As long as this is clarified, anything is good for me.
Thank you for your time

krobelus: Ok, I've added the example here. Thanks for your contribution!

I'm not sure where/if this should go into the Kakoune docs. I think the best place to teach this convention is in plugin's README's. Most of them explain what commands they provide, and list "suggested mappings", then it's clear that no standard keymaps are changed by default. Of course kak-lsp is not a small plugin, so it's easy to miss.

PR #443, opened 2021-01-26
Workspace folders from KAK_LSP_WORKSPACE_FOLDERS environment variable

Niedzwiedzw:


src/general.rs
```diff
@@ -253,7 +280,7 @@ pub fn initialize(
         root_uri: Some(Url::from_file_path(root_path).unwrap()),
         root_path: None,
         trace: Some(TraceOption::Off),
-        workspace_folders: None,
+        workspace_folders: workspace_folders_from_env(),
```
krobelus, line 283: I've hit a similar problem with bash-language-server so I've added shim support
for workspace/worspaceFolders in 2b723953e99ae034d5e595eafd83d5bc8417a9cb.
We currently always send the root path.

We don't really support multiple root paths for now but there might be a use case where it's better to be configurable.
I'd like to to avoid an environment variable (since a Kakoune process might visit several different projects).

krobelus:
Since we already send our root path via `workspace/workspaceFolders`, this should no longer be needed.
Unless there is a scenario where you need multiple workspaces in a single Kakoune session; LMK if that's the case

PR #449, opened 2021-02-09
no need to pass client where buffer context is known

vbauerster:
Some request like `update-semantic-highlighting`, `rust-analyzer/inlayHints` and `textDocument/semanticTokens/full` produce `eval -buffer ...` command which gets concatenated with `eval -client ...` so end result is recursive `eval` command, like for example:

```
eval -client client0 -verbatim -- eval -buffer '/Users/vbauer/projects/tmp/inlay-test/src/main.rs' -verbatim -- set buffer rust_analyzer_inlay_hints 1 '2.15+0|{InlayHint}{\}: String', module: kak_lsp::editor_transport:84
```

Having commands like `eval -client ... eval -buffer ..` is redundant despite it works. I think client information is not necessary as long as there is known buffer in context.

So removing `client` parameter from aforementioned requests will produce single `eval` command which works as before, but avoids recursion:

```
eval -buffer '/Users/vbauer/projects/tmp/inlay-test/src/main.rs' -verbatim -- set buffer rust_analyzer_inlay_hints 1 '2.15+0|{InlayHint}{\}: String', module: kak_lsp::editor_transport:84
```

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -730,14 +730,13 @@ rm -rf ${tmp}
 define-command lsp-update-semantic-highlighting -hidden %{
     nop %sh{ (printf '
 session      = "%s"
-client       = "%s"
 buffile      = "%s"
 filetype     = "%s"
 version      = %d
 method       = "update-semantic-highlighting"
 [params]
 current = "%s"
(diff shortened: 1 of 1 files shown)


krobelus:
I think the `"textDocument/didChange"` request also doesn't need the client because `text_document_did_change` does not use it.
Should we only pass the client when it is needed? Seems cleaner.

PR #453, opened 2021-02-12
EditorResponse.command form String to Cow<'static, str> and related refactoring

vbauerster:
This PR brings some optimization as changing type of `EditorResponse.command` form `String` to `Cow<'static, str>`, which I believe won't allocate if called like `ctx.Exec(meta, "slice str")`.

PS: feel free to reject or cherry pick something.

### DIFF
--- a/src/editor_transport.rs
+++ b/src/editor_transport.rs
@@ -2,6 +2,7 @@ use crate::thread_worker::Worker;
 use crate::types::*;
 use crate::util::*;
 use crossbeam_channel::{bounded, Receiver, Sender};
+use std::borrow::Cow;
 use std::fs;
 use std::io::{Read, Write};
 use std::os::unix::net::{UnixListener, UnixStream};
@@ -73,12 +74,19 @@ pub fn start(session: &str, initial_request: Option<String>) -> Result<EditorTra
                         let client = response.meta.client.as_ref();
                         let command = match client.filter(|&s| !s.is_empty()) {
                             Some(client) => {
-                                format!("eval -client {} -verbatim -- {}", client, response.command)
+                                let command = format!(
+                                    "eval -client {} -verbatim -- {}",
+                                    client, response.command
+                                );
+                                debug!("To editor `{}`: {}", response.meta.session, command);
+                                Cow::from(command)
                             }
-                            None => response.command.to_string(),
+                            None => {
+                                debug!("To editor `{}`: {}", response.meta.session, response.command);
+                                response.command
+                            }
                         };
 
-                        debug!("To editor `{}`: {}", response.meta.session, command);
                         if stdin.write_all(command.as_bytes()).is_err() {
                             error!("Failed to write to editor stdin");
                         }
--- a/src/language_features/formatting.rs
+++ b/src/language_features/formatting.rs
@@ -1,5 +1,4 @@
 use crate::context::*;
-use crate::text_edit::apply_text_edits_to_buffer;
 use crate::types::*;
 use lsp_types::request::*;
 use lsp_types::*;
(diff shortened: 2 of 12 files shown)


src/language_features/range_formatting.rs
```diff
@@ -29,32 +29,30 @@ pub fn text_document_range_formatting(
         meta,
         req_params,
         move |ctx: &mut Context, meta, results| {
-            let result = results.into_iter().flatten().flatten().collect();
```
vbauerster, line 32: it was not obvious than flatten().flatten() is filtering None cases, so explicit filter_map() is better in my opinion.

src/text_edit.rs
```diff
@@ -132,13 +132,13 @@ pub fn apply_text_edits_to_buffer(
...
     text: &Rope,
     offset_encoding: OffsetEncoding,
-) -> String {
+) -> Option<String> {
     // Empty text edits processed as a special case because Kakoune's `select` command
     // doesn't support empty arguments list.
     if text_edits.is_empty() {
         // Nothing to do, but sending command back to the editor is required to handle case when
         // editor is blocked waiting for response via fifo.
-        return "nop".to_string();
```
vbauerster, line 141: I think fn shouldn't know what we suppose to send back to editor, thus just returning None.

src/util.rs
```diff
@@ -180,11 +180,11 @@ pub fn get_kakoune_position(
 /// Apply text edits to the file pointed by uri either by asking Kakoune to modify corresponding
 /// buffer or by editing file directly when it's not open in editor.
 pub fn apply_text_edits(meta: &EditorMeta, uri: &Url, edits: Vec<TextEdit>, ctx: &Context) {
-    let wrapped_edits = edits
+    let edits = edits
         .into_iter()
         .map(|e| OneOf::Left(e))
-        .collect::<Vec<OneOf<TextEdit, AnnotatedTextEdit>>>();
-    apply_annotated_text_edits(meta, uri, &wrapped_edits[..], ctx)
```
vbauerster, line 187: it's not necessary to add `[..]` as coercion to slice is inferred.

src/editor_transport.rs
```diff
@@ -73,12 +74,19 @@ pub fn start(session: &str, initial_request: Option<String>) -> Result<EditorTra
...
+                                Cow::from(command)
+                            }
+                            None => {
+                                debug!("To editor `{}`: {}", response.meta.session, response.command);
+                                response.command
                             }
-                            None => response.command.to_string(),
                         };
 
-                        debug!("To editor `{}`: {}", response.meta.session, command);
```
vbauerster, line 81: I had to move and duplicate debug! call in above match in order to preserve `Cow::borrowed` variant in None case.

krobelus (commented):
The changes seem logical. Of course we should avoid making the code harder to read in exchange for small performance gains, however, `Cow` doesn't seem too bad, and call sites are actually nicer to read because we no longer need to call `to_string()`.
Searching for `to_string()` reveals some more places that can have similar changes in future.

I left some nitpicks, let me know if you want to apply them.

src/language_features/formatting.rs
```diff
@@ -17,40 +16,11 @@ pub fn text_document_formatting(meta: EditorMeta, params: EditorParams, ctx: &mu
         work_done_progress_params: Default::default(),
     };
     ctx.call::<Formatting, _>(meta, req_params, move |ctx: &mut Context, meta, result| {
-        editor_formatting(meta, result, ctx)
+        // result is Option<Vec<TextEdit>>
+        let text_edits = result.unwrap_or_else(Vec::new)
+            .into_iter()
+            .map(|e| OneOf::Left(e))
```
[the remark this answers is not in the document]
vbauerster: Yeah, this is tricky. Frankly I didn't touch it because I thought it would work with enum variants, but it works.

vbauerster:
@krobelus Thanks for reviewing. I've updated.

Issue #406, opened 2020-09-30
Avoid changing directory when creating grep/make buffers


krobelus:
Running `lsp-references` on a symbol will create a scratch buffer with filetype
`grep` that contains relative paths, like so

```
src/test.go:5:6:func foo() {
```

Kakoune's `grep-jump` will run `edit` with those relative paths.  To make that
work, kak-lsp uses `cd %arg{1}` in `lsp-show-goto-choices` and others to change
directory to the project root.  I think that's surprising and should be avoided.

A few possible solutions:

1. Use absolute paths (ugly)
2. Use absolute paths only if necessary: when the cwd is not the project root
   The disadvantage here is that a cd invalidates the locations.
3. Similar to 2: make paths relative to the cwd
4. Add a `dir` variable to `grep.kak` and `make.kak`. Set it to the project root at buffer scope.
5. Reimplement the necessary bits from `grep.kak` and `make.kak` to do the same.
6. Kakoune could honor `%opt{path}` in `edit -existing`. For example `edit -existing stdio.h` would edit `/usr/include/stdio.h`.
   With that we can configure the scratch buffers with `set buffer path %arg{1} %opt{path}`,
   and the jumps will just work.

PR #454, opened 2021-02-15
Don't change kakoune's directory with diagnostics/goto

tadeokondrak:
This is solution ~~2~~ 5 from #406

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -78,6 +78,7 @@ declare-option -hidden range-specs lsp_semantic_highlighting
 declare-option -hidden range-specs lsp_semantic_tokens
 declare-option -hidden range-specs rust_analyzer_inlay_hints
 declare-option -hidden range-specs lsp_diagnostics
+declare-option -hidden str lsp_project_root
 
 ### Requests ###
 
@@ -963,9 +964,15 @@ define-command -hidden lsp-show-error -params 1 -docstring "Render error" %{
 define-command -hidden lsp-show-diagnostics -params 2 -docstring "Render diagnostics" %{
     evaluate-commands -save-regs '"' -try-client %opt[toolsclient] %{
         edit! -scratch *diagnostics*
-        cd %arg{1}
-        try %{ set-option buffer working_folder %sh{pwd} }
-        set-option buffer filetype make
+        set-option buffer filetype lsp-goto
+        set-option buffer lsp_project_root "%arg{1}/"
+
+        # This buffer behaves a bit like a *make* buffer.
+        alias buffer lsp-jump lsp-diagnostics-jump
+        remove-highlighter window/lsp-goto
+        add-highlighter window/lsp-diagnostics ref lsp-diagnostics
+        hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/lsp-diagnostics }
+
         set-register '"' %arg{2}
         execute-keys Pgg
     }
@@ -974,10 +981,9 @@ define-command -hidden lsp-show-diagnostics -params 2 -docstring "Render diagnos
 define-command -hidden lsp-show-goto-choices -params 2 -docstring "Render goto choices" %{
     evaluate-commands -save-regs '"' -try-client %opt[toolsclient] %{
         edit! -scratch *goto*
-        cd %arg{1}
-        try %{ set-option buffer working_folder %sh{pwd} }
-        set-option buffer filetype grep
+        set-option buffer filetype lsp-goto
         set-option buffer grep_current_line 0
+        set-option buffer lsp_project_root "%arg{1}/"
         set-register '"' %arg{2}
         execute-keys Pgg
     }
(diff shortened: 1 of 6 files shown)


src/util.rs
```diff
@@ -222,3 +216,11 @@ pub fn get_file_contents(filename: &str, ctx: &Context) -> Option<Rope> {
                 .and_then(|f| Rope::from_reader(BufReader::new(f)).ok())
         })
 }
+
+pub fn short_file_path<'a>(target: &'a str, current_dir: &str) -> &'a str {
+    Path::new(target)
+        .strip_prefix(current_dir)
```
krobelus, line 222: I wonder if we can make the paths relative to Kakoune's directory, instead of the project root.
Otherwise if I edit `src/main.rs`, use `:cd src`, and then call `:lsp-references`, the `src/main.rs` path is invalid. (also that stopped working at all after the `cd`).

I think we want the behavior of `realpath src/main.rs --relative-to $PWD`, where $PWD is the Kakoune server's working directory.

krobelus: This would be a very dirty implementation, by simply overriding the old commands:

```
define-command -override -hidden lsp-show-goto-choices -params 2 -docstring "Render goto choices" %{
	evaluate-commands -save-regs '"' -try-client %opt[toolsclient] %{
		edit! -scratch *goto*
		set-option buffer filetype grep
		set-option buffer grep_current_line 0
		set-register '"' %sh{
			nl='
			'
			IFS=$nl
			printf %s "$2" | while read line; do
			file=${line%%:*}
			rest=${line#*:}
			# Relative file paths are relative to the project root.
			[ x"${file#/}" = x"${file}" ] && file=$1/$file
			printf '%s:%s\n' "$(realpath --relative-to "$PWD" "$file")" "$rest"
			done
		}
		execute-keys Pgg
	}
}
```

I'm not yet sure how this should be implemented

krobelus:
Thanks for taking this on!

With this variant we should probably keep `grep-jump` working even if Kakoune's working directory is different from the project root - by making the paths relative to Kakoune's `$PWD`.

If we add a "root directory" option to grep-like buffers, then `grep-jump` could just use that instead (option 4).
Option 6 seems also nice although maybe a bit too magical.

krobelus:
> If we add a "root directory" option to grep-like buffers, then grep-jump could just use that instead (option 4).

If we don't want to make paths relative to Kakoune's $PWD, but keep them relative to the project root, then we can probably implement this fairly easily: simply add a buffer-local alias that overrides `grep-jump`. The behavior is the same, but it prepends the project root before jumping to relative paths.

tadeokondrak:
Switched from option 2 to option 5 (copy needed parts from Kakoune).

rc/lsp.kak
```diff
@@ -1613,3 +1614,85 @@ def lsp-snippets-select-next-placeholders %{
...
+        edit -existing "%arg{1}" %arg{2} %arg{3}
+        echo -markup "{Information}{\}%arg{4}"
+        try %{ focus }
+    }
+}
+
+define-command -hidden lsp-diagnostics-jump %{
+    evaluate-commands %{
+        try %{
+            execute-keys gl<a-?> "Entering directory" <ret><a-:>
```
krobelus, line 1669: We don't need this `try` block - it is a Make-specific bit that guesses the directory that Make is in.

rc/lsp.kak
```diff
@@ -1613,3 +1614,85 @@ def lsp-snippets-select-next-placeholders %{
...
+
+declare-option -docstring "name of the client in which all source code jumps will be executed" \
+    str jumpclient
+
+define-command -hidden lsp-goto-jump %{
+    evaluate-commands %{ # use evaluate-commands to ensure jumps are collapsed
+        try %{
+            execute-keys '<a-x>s^((?:\w:)?[^:]+):(\d+):(\d+)?<ret>'
+            set-option buffer lsp_goto_current_line %val{cursor_line}
+            evaluate-commands -try-client %opt{jumpclient} -verbatim -- edit -existing "%opt{lsp_goto_working_directory}/%reg{1}" %reg{2} %reg{3}
```
krobelus, line 1640: This will fail on absolute paths - for example try `:lsp-references` on `App::new` in `src/main.rs`.
We can detect absolute paths:

```
try %{
    execute-keys -draft s^/<ret>
} catch %{
    set-register 1 "%opt{lsp_goto_working_directory}/%reg{1}"
}
evaluate-commands -try-client %opt{jumpclient} -verbatim -- edit -existing %reg{1} %reg{2} %reg{3}
```

krobelus (commented):
Looks good!
At some point I want to unify grep buffer handling, but this works well for now. Plus we probably want to deviate a bit from make.kak.

krobelus:
`grep-jump` in `lsp-next-match` should also be replaced by `lsp-goto-jump`, right?

krobelus:
Added a uniform way to handle all those grep/goto/diagnostics buffers.
I'm fairly happy with this, because it allows to treat `*diagnostics*` like a first-class `*make*` buffer: `lsp-next-location *diagnostics*` works, and also displays the error message in the prompt. We couldn't use `make-next-error` because that hardcoded `*make*` as buffer name.

cc @lenormf maybe you have feedback on the direction

krobelus:
Merging, thanks! I'm glad this is finally fixed.
Moving my additions of `lsp-next-location` to a follow-up.

PR #455, opened 2021-02-15
completion: fix "simple textEdit" detection for non-ASCII chars

tadeokondrak:
The first commit of the branch mentioned in https://github.com/kak-lsp/kak-lsp/issues/40#issuecomment-663838631.

### DIFF
--- a/src/language_features/completion.rs
+++ b/src/language_features/completion.rs
@@ -1,4 +1,5 @@
 use crate::context::*;
+use crate::position::lsp_range_to_kakoune;
 use crate::types::*;
 use crate::util::*;
 use itertools::Itertools;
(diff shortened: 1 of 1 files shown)


krobelus:
Thanks for digging this up, I completely forgot to submit this patch after I got stuck on the textEdit support.

Language servers like `ccls` always send completions as textEdit. Luckily most other language servers don't seem to do this.
This fixes non-ASCII completion of "simple" textEdit completions.
Here is an example that is fixed by this patch.

```c
void føøbår(int x){}

int main() {
/* enter insert mode after a word to trigger completion */
/* fixed: */ fø
/* not fixed: */ føb
}
```
The second example (`føb`) is not fixed. For some reason `ccls` sends an insertion of `føøbår` (into an empty range), when it should actually replace 3 characters.

```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "result": {
    "isIncomplete": false,
    "items": [
      {
        "label": "føøbår(int x) -> void",
        "kind": 3,
        "detail": "",
        "sortText": "   !",
        "filterText": "føøbår",
        "insertTextFormat": 1,
        "textEdit": {
          "range": {
            "start": {
              "line": 5,
              "character": 20
            },
            "end": {
              "line": 5,
              "character": 20
            }
          },
          "newText": "føøbår"
        }
      }
    ]
  }
}
```

This change seems to fix that case. I need to rationalize it though, because this case doesn't seem to happen with ASCII-only inputs. So maybe we shouldn't add this extra hack, because it covers up some broken character handling.. I'm not sure.

```diff
diff --git a/src/language_features/completion.rs b/src/language_features/completion.rs
index c941d6a..3f27912 100644
--- a/src/language_features/completion.rs
+++ b/src/language_features/completion.rs
@@ -91,7 +91,8 @@ pub fn editor_completion(
                         lsp_range_to_kakoune(&text_edit.range, &document.text, ctx.offset_encoding);
                     range.start.line == params.position.line
                         && range.end.line == params.position.line
-                        && range.end.column + 1 == params.position.column
+                        && (range.end.column == params.position.column
+                            || range.end.column + 1 == params.position.column)
                 } else {
                     false
                 }
```

Issue #445, opened 2021-01-31
Spurious warning when using proc macros and rust-analyzer


dmerejkowsky:
Steps to reproduce:

kakoune version: v2020.09.01

rust-analyzer version:dbd0cfb

kak-lsp version: 9.0.0

```rust
// in src/main.rs
use paste::paste;

macro_rules! plop {
    ($name: ident) => {
        paste! {
            let [< _plop_ $name>] = 42 ;
        };
    };
}

fn main() {
    plop!(x);
}
```

```toml
# in Cargo.toml
[package]
name = "foo"
version = "0.1.0"

[dependencies]
paste = "1.0.4"
``` 

```kak
# In kakrc
set-option global lsp_server_configuration rust-analyzer.diagnostics.disabled=["unresolved-proc-macro"]
```

```
:lsp-enable
:lsp-diagnostics

src/main.rs:12:5: warning:proc macro not expanded
````

Maybe there's something wrong with the `lsp_server_configuration` option?

I tried to enable logs for both rust-analyzer and kak-lsp but did not see any relevant info #16

**Comment by dmerejkowsky** [comment]
Found it! Turns out you can *also* set options directly in the kak-lsp.toml file, like this:

```toml
[language.rust]
filetypes = ["rust"]
roots = ["Cargo.toml"]
command = "rust-analyzer"
[language.rust.initialization_options]
diagnostics.disabled = ["unresolved-proc-macro"]
````

**Comment by krobelus** [comment]
This is not specific to Kakoune/kak-lsp. The warning can be reproduced on your example with 

```$
$ rust-analyzer diagnostics src/main.rs
processing crate: foo, module: $PWD/src/main.rs
Diagnostic { message: "proc macro not expanded", range: 388..396, severity: WeakWarning, fix: None, unused: false, code: Some(DiagnosticCode("unresolved-proc-macro")) }

diagnostic scan complete
```

Relevant issues are https://github.com/rust-analyzer/rust-analyzer/issues/7497 and https://github.com/rust-analyzer/rust-analyzer/issues/6448.
So the alternative fix would be to enable proc macro support in rust-analyzer:

```toml
[language.rust.initialization_options]
cargo.loadOutDirsFromCheck = true
procMacro.enable = true
```

PR #456, opened 2021-02-15
rust: fix "unresolved proc macro" warning

dmerejkowsky:
I guess kakoune is not able to implement the proc macro expansion (yet?)

So this can save a lot of headaches for rust-analyzer users

### DIFF
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -180,6 +180,20 @@ args = ["-c", "if command -v rustup >/dev/null; then $(rustup which rls); else r
 # roots = ["Cargo.toml"]
 # command = "sh"
 # args = ["-c", "if command -v rustup >/dev/null; then $(rustup which rust-analyzer); else rust-analyzer; fi"]
+#
+# [language.rust.initialization_options]
+#
+# If you get 'unresolved proc macro' warnings, you have two options
+#
+# 1/ The safe choice is two disable the warning:
+# diagnostics.disabled = ["unresolved-proc-macro"]
+#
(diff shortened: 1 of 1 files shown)


krobelus:
I didn't encounter this diagnostic - can you share steps to trigger it?
I'm curious why this doesn't work, maybe there is a missing feature in kak-lsp.

dmerejkowsky:
> can you share steps to trigger it?

See #445

> I'm curious why this doesn't work, maybe there is a missing feature in kak-lsp.

Yeah me too :)

kak-lsp.toml
```diff
@@ -180,6 +180,8 @@ args = ["-c", "if command -v rustup >/dev/null; then $(rustup which rls); else r
 # roots = ["Cargo.toml"]
 # command = "sh"
 # args = ["-c", "if command -v rustup >/dev/null; then $(rustup which rust-analyzer); else rust-analyzer; fi"]
+# [language.rust.initialization_options]
+# diagnostics.disabled = ["unresolved-proc-macro"]
```
krobelus, line 184: I'm fine with merging this unless you think we should instead show how to enable proc macro support.
There are [some reasons](https://github.com/rust-analyzer/rust-analyzer/issues/6448) why it's not the default, so just disabling the warning seems like the safer choice.

dmerejkowsky:
> so just disabling the warning seems like the safer choice.

Well, this is commented out code anyway, so I guess we can leave the two options and let users choose.

krobelus:
thanks!

PR #468, opened 2021-03-16
Add rootPath for backward compatibility

ebetica:
I'm not sure if this will break other language servers - since both rootPath and rootUri are deprecated, pyright only supports rootPath, but not rootUri, so this was essential in getting pyright working. However, kak-lsp should be moving to workspaceFolders going forwards. I tried it but could not get it working, so hopefully this will do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,5 +10,6 @@ Bug fixes:
 - `lsp-show-{diagnostics,goto-choices,document-symbols}` no longer `cd` to the project root (#454).
 - Fix error when the documentation part of a completion item starts with a dash (`-`) (#460).
 - Fix completions of non-ASCII characters when using textEdit (which is not only partially supported) (#455).
+- Fix `lsp-rename` with `pyright` (#468).
 
 For release notes on v9.0.0 and older see <https://github.com/kak-lsp/kak-lsp/releases>.
--- a/src/general.rs
+++ b/src/general.rs
@@ -280,7 +280,7 @@ pub fn initialize(
         initialization_options,
         process_id: Some(process::id()),
         root_uri: Some(Url::from_file_path(root_path).unwrap()),
-        root_path: None,
+        root_path: Some(root_path.to_string()),
         trace: Some(TraceOption::Off),
         workspace_folders: None,
         client_info: Some(ClientInfo {


krobelus:
Hi, thanks for this patch! Unfortunately it doesn't work for me, and it seems to break go-to definition with pyright 1.1.104.
Maybe there is something wrong about my setup?

My config is

```toml
[language.python]
filetypes = ["python"]
roots = ["requirements.txt", "pyproject.toml", "setup.py", ".git"]
command = "pyright-langserver"
args = ["--stdio"]
offset_encoding = "utf-8"
```

The log should look like the one below, but the last line (`From
server`) was missing when testing with your changes:

```
Mar 20 15:51:54.851 DEBG From editor: 
session   = "434446"
client    = "client0"
buffile   = "/path/to/test.py"
filetype  = "python"
version   = 1
method    = "textDocument/definition"
[params.position]
line      = 10
column    = 1
, module: kak_lsp::editor_transport:129
Mar 20 15:51:54.852 DEBG Searching for vars starting with KAK_LSP_PROJECT_ROOT_PYTHON, module: kak_lsp::project_root:46
Mar 20 15:51:54.854 DEBG Routing editor request to Route { session: "434446", language: "python", root: "/home/johannes/git/kak-lsp" }, module: kak_lsp::session:94
Mar 20 15:51:54.855 DEBG To server: {"jsonrpc":"2.0","method":"textDocument/definition","params":{"position":{"character":0,"line":9},"textDocument":{"uri":"file:///path/to/test.py"}},"id":1}, module: kak_lsp::language_server_transport:177
Mar 20 15:51:54.858 DEBG From server: {"jsonrpc":"2.0","id":1,"result":[{"uri":"file:///path/to/test.py","range":{"start":{"line":2,"character":4},"end":{"line":2,"character":10}}}]}, module: kak_lsp::language_server_transport:151
```

ebetica:
You're are right. I never tested this _without_ a `pyrightconfig.json` in the root directory. It seems that this is actually necessary. Here is an example:
```
$ cat pyrightconfig.json 
{
  "include": [
    "*"
  ],
  "exclude": [
    "**/node_modules",
    "**/__pycache__",
  ],
  "executionEnvironments": [
    {
      "root": ".",
    }
  ]
}
```

I don't know if I have the patience to find out how to properly fix this backward compatibility error, but when kak-lsp moves to the non-deprecated `workplaceFolders` initialization field, this should be solved.

krobelus:
`pyright` still didn't work with you `pyrightconfig.json`, maybe I missed something else?
Anyway, it doesn't seem to break other language servers, so let's use that for now.

ebetica:
If it helps you at all, I am on version pyright 1.1.121

PR #471, opened 2021-04-01
Change ocaml command to ocamllsp

unrelentingtech:
The https://github.com/ocaml-lsp/ocaml-language-server project has been archived, it is not receiving updates anymore.

https://github.com/ocaml/ocaml-lsp is the currently supported "blessed" LSP server (note how it's under the ocaml org).
Its executable is called ocamllsp and it only supports stdio mode (there's no stdio flag).


krobelus:
Thank you! I've updated the wiki accordingly.

PR #478, opened 2021-04-20
Fix warning during compilation

rsaihe:
This commit fixes the following warning that occurs during compilation:

```
warning: panic message is not a string literal
  --> src/thread_worker.rs:29:24
   |
29 |                 panic!(e)
   |                        ^
   |
   = note: `#[warn(non_fmt_panic)]` on by default
   = note: this is no longer accepted in Rust 2021
help: add a "{}" format string to Display the message
   |
29 |                 panic!("{}", e)
   |                        ^^^^^
help: or use std::panic::panic_any instead
   |
29 |                 std::panic::panic_any(e)
   |                 ^^^^^^^^^^^^^^^^^^^^^^

warning: 1 warning emitted
```


krobelus:
yup, thanks!

PR #487, opened 2021-05-29
Send did change synchronously to address #267

### LINKED ISSUE #267: Race condition between didSave and completion
I kept getting the wrong typescript completions when writing `object.`, I got the completions for `object ` instead. I realized what the problem is: `lsp-completion` directly runs `lsp-did-change`: 
https://github.com/ul/kak-lsp/blob/243da6cb8ffa4e5367f5777d2fc48a1a9c30b875/rc/lsp.kak#L103
However, `lsp-did-change` asynchronously sends the buffer contents to the `kak-lsp` binary.  So there is as race between that and the message `lsp-completion` sends to the binary. And indeed didChange loses, presumably because it spawns other binaries such as `sed` and the script for completions is much simpler.

One way to solve this would be to make a command `lsp-did-change-and-then` taking an extra argument to run that will be guaranteed to run after the new buffer contents has been sent.

danr:
I think it is enough to send didChange synchronousy to fix #267. I've only tried it with pyright so there might be some issue together with other language servers.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -175,15 +175,11 @@ define-command -hidden lsp-did-change-and-then -params 1 -docstring %{
             echo "fail"
         fi
     }
-    declare-option -hidden str lsp_callback "evaluate-commands -client %val{client} %arg{1}"
     set-option buffer lsp_timestamp %val{timestamp}
     evaluate-commands -save-regs '|' %{
         set-register '|' %{
-# dump stdin synchronously
 # append a . to the end, otherwise the subshell strips trailing newlines
 lsp_draft=$(cat; printf '.')
-# and process it asynchronously
-(
 # replace \ with \\
 #         " with \"
 #     <tab> with \t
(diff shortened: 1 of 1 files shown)


krobelus:
Hi, it looks reasonable to do this synchronously since it should always be fast.

However, it seems to break completion with `rls` 1.41

Start kak like 
`PATH=target/debug:$PATH kak +110 src/main.rs`
and type `matches.` in a new line. I no longer get field/method completions.
Of course this could be because of some issue with `rls`..

krobelus:
also let me know if you have found a way to trigger the race condition

danr:
Thanks krobelus! I'm swamped right now, I hope I can get a reliable way to reproduce the race condition later.

rc/lsp.kak
```diff
@@ -120,8 +118,8 @@ method   = "textDocument/didChange"
 draft    = """
 %s"""
 ' "${kak_session}" "${kak_buffile}" "${kak_opt_filetype}" "${kak_timestamp}" "${lsp_draft}" | eval ${kak_opt_lsp_cmd} --request
-printf %s "${kak_opt_lsp_callback}" | kak -p "${kak_session}"
-) > /dev/null 2>&1 < /dev/null & }
+printf %s "${kak_opt_lsp_callback}"
```
krobelus, line 121: Wow, I just realized why this broke rls completions - it's because we stop sending the callback.

The current lsp-did-change-then did:

    evaluate-commands -save-regs '|' %{
        set-register '|' %{
            echo ... | kak-lsp --request >/dev/null 2>&1 </dev/null &
            printf %s\\n "${kak_opt_lsp_callback}" | kak -p "${kak_session}"
        }
        execute-keys -draft '%<a-|><ret>'
    }

your version does

    evaluate-commands -save-regs '|' %{
        set-register '|' %{
            echo ... | kak-lsp --request
            printf %s\\n "${kak_opt_lsp_callback}"
        }
        execute-keys -draft '%<a-|><ret>'
    }

`<a-|>` throws away its output, so kak_opt_lsp_callback is not run.

The correct solution is much simpler, I'll push it in a bit.
I'll test a bit more and merge seen.

(I found this because I needed a synchronouse lsp-did-change-and-then for https://github.com/kak-lsp/kak-lsp/issues/582)

PR #488, opened 2021-05-29
Rework Semantic Token modifiers

sclu1034:
This reworks the semantic highlighting feature to properly utilize modifiers into a feature that matches their name.
Rather than specifying separate face mappings for tokens and modifiers, where the latter overwrite the former, they are consolidated into a single configuration structure.
This allows to map the same token to different faces based on the presence of modifiers.

This does include a breaking change within the configuration file.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -304,7 +304,75 @@ hook global WinSetOption filetype=<language> %{
 }
 ----
 
-The faces used for semantic tokens and modifiers can be modified in `kak-lsp.toml`, under the `semantic_tokens` and `semantic_token_modifiers` sections. The modifiers are used first if available, and then the main token type is used if no modifier face is specified.
+The faces used for semantic tokens and modifiers can be modified in `kak-lsp.toml`, under the `semantic_tokens` section.
+The syntax for such an entry is
+
+[source,toml]
+----
+[[semantic_tokens]]
+name = "variable"
+face = "const_variable_declaration"
+modifiers = ["constant", "declaration"]
+----
+
+where `name` is the token's name as reported by the language server (see `lsp-capabilities`), `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config) and `modifiers` is an array of modifier names (also reported by the language server). `modifiers` may be omitted, but `name` and `face` are required.
+
+You may create any arbitrary number of definitions with permutations between the token names and modifiers reported by the server. For an entry to match a token, all the entrie's modifiers must exist on the token. However, the token may have additional modifiers not assigned in the config entry. +
+`kak-lsp` will find the most specific matching configuration to apply, where specificity is defined as the number of matching modifiers. If multiple entries match with the same number of modifiers, the one that was defined last in the configuration wins.
+
+*Example:*
+
+Assuming the following configuration,
+
+[source,toml]
+----
+[[semantic_tokens]]
+name = "variable"
+face = "const_variable_declaration"
+modifiers = ["constant", "declaration"]
+
+[[semantic_tokens]]
+name = "variable"
+face = "const_variable"
+modifiers = ["constant"]
+
+[[semantic_tokens]]
+name = "variable"
+face = "variable"
+----
+
+`kak-lsp` will perform these mappings:
+
+[cols="1,1,2,5"]
+|===
+| Token | Modifiers | Face | Comment
+
+| `variable`
+| `constant`, `declaration`
+| `const_variable_declaration`
+| First entry matches with 2 modifiers.
+
+| `variable`
+| `constant`
+| `const_variable`
+| First and second entry match with 1 modifier, second wins.
+
+| `variable`
+| `declaration`
+| `variable`
+| Only third entry matches. First entry doesn't match, because `constant` is missing.
+
+| `variable`
+|
+| `variable`
+| Third entry matches.
+
+| `function`
+|
+|
+| No entries match and no face is applied.
+
+|===
 
 == Inlay Diagnostics
 
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -317,7 +317,7 @@ modifiers = ["constant", "declaration"]
 
 where `name` is the token's name as reported by the language server (see `lsp-capabilities`), `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config) and `modifiers` is an array of modifier names (also reported by the language server). `modifiers` may be omitted, but `name` and `face` are required.
 
-You may create any arbitrary number of definitions with permutations between the token names and modifiers reported by the server. For an entry to match a token, all the entrie's modifiers must exist on the token. However, the token may have additional modifiers not assigned in the config entry. +
+You may create any arbitrary number of definitions with permutations between the token names and modifiers reported by the server. For an entry to match a token, all the entry's modifiers must exist on the token. However, the token may have additional modifiers not assigned in the config entry. +
 `kak-lsp` will find the most specific matching configuration to apply, where specificity is defined as the number of matching modifiers. If multiple entries match with the same number of modifiers, the one that was defined last in the configuration wins.
 
 *Example:*
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -310,12 +310,12 @@ The syntax for such an entry is
 [source,toml]
 ----
 [[semantic_tokens]]
-name = "variable"
+token = "variable"
 face = "const_variable_declaration"
 modifiers = ["constant", "declaration"]
 ----
 
-where `name` is the token's name as reported by the language server (see `lsp-capabilities`), `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config) and `modifiers` is an array of modifier names (also reported by the language server). `modifiers` may be omitted, but `name` and `face` are required.
+where `token` is the token's name as reported by the language server (see `lsp-capabilities`), `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config) and `modifiers` is an array of modifier names (also reported by the language server). `modifiers` may be omitted, but `token` and `face` are required.
 
 You may create any arbitrary number of definitions with permutations between the token names and modifiers reported by the server. For an entry to match a token, all the entry's modifiers must exist on the token. However, the token may have additional modifiers not assigned in the config entry. +
 `kak-lsp` will find the most specific matching configuration to apply, where specificity is defined as the number of matching modifiers. If multiple entries match with the same number of modifiers, the one that was defined last in the configuration wins.
@@ -327,17 +327,17 @@ Assuming the following configuration,
 [source,toml]
 ----
 [[semantic_tokens]]
-name = "variable"
+token = "variable"
 face = "const_variable_declaration"
 modifiers = ["constant", "declaration"]
 
 [[semantic_tokens]]
-name = "variable"
+token = "variable"
 face = "const_variable"
 modifiers = ["constant"]
 
 [[semantic_tokens]]
-name = "variable"
+token = "variable"
 face = "variable"
 ----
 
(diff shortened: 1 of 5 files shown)


sclu1034:
I must've missed that GitHub picked the first commit's message as PR title.
Updated to match the actual content.

src/language_features/semantic_tokens.rs
```diff
@@ -65,18 +66,53 @@ pub fn tokens_response(meta: EditorMeta, tokens: SemanticTokensResult, ctx: &mut
...
+                            .count();
+
+                        Some((&token_config.face, modifier_count))
+                    })
+                    .collect::<Vec<_>>();
+
+                // Sort by number of matching modifiers, in descending order
+                candidates.sort_by_key(|val| val.1);
+                candidates.reverse();
+                candidates.first().map(|val| format!("{}|{}", range, val.0))
```
krobelus, line 110: It sucks that we can't do a descending sort by simply negating `val.1`. We could use bitwise negation but that's hardly intuitive.

I guess we can use `last()` instead of `first()`

src/general.rs
```diff
@@ -249,16 +250,23 @@ pub fn initialize(
                     token_types: ctx
                         .config
                         .semantic_tokens
-                        .keys()
+                        .iter()
                         .cloned()
-                        .map(|x| x.into())
+                        .map(|token| token.name.into())
+                        // Collect into HashSet first to remove duplicates
```
krobelus, line 256: I'd probably say "a set" instead of "HashSet" in the comment because it's an implementation detail.
For extremely large data, a BTreeSet would probably be more efficient but that's only a guess and it definitely doesn't matter here.

kak-lsp.toml
```diff
@@ -4,20 +4,37 @@ verbosity = 2
...
+name = "keyword"
+face = "keyword"
+
+[[semantic_tokens]]
+name = "operator"
+face = "operator"
+
+[[semantic_tokens]]
+name = "comment"
+face = "comment"
```
krobelus, line 42: Could you also sort them alphabetically?

src/language_features/semantic_tokens.rs
```diff
@@ -3,6 +3,7 @@ use crate::position::lsp_range_to_kakoune;
 use crate::types::{EditorMeta, EditorParams};
 use crate::util::editor_quote;
 use lsp_types::request::SemanticTokensFullRequest;
+use lsp_types::SemanticTokenModifier;
 use lsp_types::{
     Position, Range, SemanticToken, SemanticTokensOptions, SemanticTokensParams,
```
krobelus, line 8: `SemanticTokenModifier` from above should be added here.

src/general.rs
```diff
@@ -249,16 +250,23 @@ pub fn initialize(
...
                         .cloned()
-                        .map(|x| x.into())
+                        // Get all modifiers used in token definitions
+                        .flat_map(|token| token.modifiers)
+                        // Collect into HashSet first to remove duplicates
+                        .collect::<HashSet<SemanticTokenModifier>>()
+                        .into_iter()
                         .collect(),
                     formats: vec![TokenFormat::RELATIVE],
                     overlapping_token_support: None,
```
krobelus, line 272: We still send `semantic_tokens: None,` in `src/general.rs:86` as part of the `WorkspaceClientCapabilities`.

Do we already support that? I didn't check what difference it will make in practice to set that field.

src/general.rs
```diff
@@ -249,16 +250,23 @@ pub fn initialize(
                     token_types: ctx
                         .config
                         .semantic_tokens
-                        .keys()
+                        .iter()
                         .cloned()
-                        .map(|x| x.into())
+                        .map(|token| token.name.into())
```
krobelus, line 255: We don't need to used `cloned()` anymore since we don't need to clone the entire `SemanticTokenConfig` object.
Instead something like `.map(|token| token.name.clone().into())` would work. Same below.

krobelus (commented):
Thanks for this PR - the documentation and examples are real nice!

I'm not sure if we should add a compatibility hack for the config file.
We could
1. Keep parsing the old format, either ignoring old semantic tokens, or converting them on-the-fly.
2. Automatically rewrite the config file  to the new format

To avoid bugging users who  have copied the config file even when they are not using semantic tokens.
I'm not yet sure if  it's worth it, since we even have a pop-up inside Kakoune to show config parse errors.

kak-lsp.toml
```diff
@@ -4,20 +4,37 @@ verbosity = 2
...
-namespace = "module"
-function = "function"
-string = "string"
-keyword = "keyword"
-operator = "operator"
-comment = "comment"
-
-[semantic_token_modifiers]
-documentation = "documentation"
-readonly = "default+d"
```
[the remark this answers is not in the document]
sclu1034: There is no 1:1 "catch-all" equivalent.
However, `documentation` will likely only ever appear on a `comment`, so the near-equivalent for that would be

```toml
[[semantic_tokens]]
name = "comment"
face = "documentation"
modifiers = ["documentation"]
```

I guess I'll add that, since it should make for a good example.

For `readonly`, it's still possible to get the same functionality. But it would need a bit more work, since individual entries for every token that can have that modifiers need to be added.

src/language_features/semantic_tokens.rs
```diff
@@ -65,18 +66,53 @@ pub fn tokens_response(meta: EditorMeta, tokens: SemanticTokensResult, ctx: &mut
...
-                    .chain(ctx.config.semantic_tokens.get(token.as_str()))
-                    .next()
-                    .map(|face| format!("{}|{}", range, face))
+                    .collect();
+
+                let mut candidates = ctx
+                    .config
+                    .semantic_tokens
+                    .iter()
+                    .filter_map(|token_config| {
```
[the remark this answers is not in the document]
sclu1034: Clippy likes to complain about "`filter` followed by `map`" actually.
Moving it to `sort_by_cached_key` could make sense, though.

But I don't think it would have much of an effect, so I'm fine with any option.

kak-lsp.toml
sclu1034: I could, but given that the sort order is tied to functionality here, it might actually be counter-productive to have the reference/example be sorted based on aesthetics.

sclu1034:
I agree that the error message for an invalid config file will be annoying to users, and I'm sure it'd be possible to rework the config parsing in a way that the old format would still be accepted. I'd just have to dig into `serde`.

On the other hand - although this might sound a bit like "I want users to use my feature" - the hard change could also be beneficial, as it would get people to know about this, and possibly other new features.
I know from my own experience that, especially with `plug-update`, additions can simply drift by unnoticed. I only noticed all the inlay hints and semantic tokens features when I happened to be browsing the README some time ago.

But the error message of "expected a sequence, not a map" isn't great for that kind of thing. Maybe `serde` has an option to customize error messages for certain fields.

krobelus:
Yeah I think the breakage is fine if we provide obvious steps to fix the config in the error message and CHANGELOG.md.

We get this error

```rust
Error {
  inner: ErrorInner {
    kind: Custom,
    line: Some(216),
    col: 0,
    at: Some(4955),
    message: "invalid type: map, expected a sequence",
    key: ["semantic_tokens"]
  }
}
```

so we could add an extra line to the error if we get this exact `message` and `key`

```
Failed to parse config file: invalid type: map, expected a sequence for key `semantic_tokens` at line 217 column 1
See https://github.com/kak-lsp/kak-lsp#semantic-tokens for the new configuration syntax for semantic tokens.
```

src/language_features/semantic_tokens.rs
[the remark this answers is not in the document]
krobelus: In this instance clippy doesn't complain about filter+map.
Looks like there were some recent changes that probably made it more specific.
I imagine it should only apply if the closures in filter+map have some duplicate code.

The advantage of smaller filters is smaller scope, though I agree that any of these options is fine.

However, we can just use `Iterator::max_by_key` instead:

```rust
                let candidates = ctx.config.semantic_tokens.iter().filter(|token_config| {
                    token_name == token_config.name &&
                        // All the config's modifiers must exist on the token for this
                        // config to match.
                        token_config
                            .modifiers
                            .iter()
                            .all(|modifier| token_modifiers.contains(&modifier))
                });
                // Not all the token's modifiers must exist in the config.  Therefore, we use
                // the token that matches the most configured modifiers.
                let best = candidates.max_by_key(|token_config| {
                    token_modifiers
                        .iter()
                        .filter(|modifier| token_config.modifiers.contains(modifier))
                        .count()
                });
                best.map(|token_config| format!("{}|{}", range, token_config.face))
```

kak-lsp.toml
krobelus: AIUI sorting by token name doesn't have an impact with the current behavior (only sorting by modifier does).
A larger config probably wants to group the config by token type, to make the order obvious.

We can keep it of course. I just like to sort these because it makes it easier search, and easiei to decide where to add new entries

README.asciidoc
```diff
@@ -304,7 +304,75 @@ hook global WinSetOption filetype=<language> %{
...
 ----
 
-The faces used for semantic tokens and modifiers can be modified in `kak-lsp.toml`, under the `semantic_tokens` and `semantic_token_modifiers` sections. The modifiers are used first if available, and then the main token type is used if no modifier face is specified.
+The faces used for semantic tokens and modifiers can be modified in `kak-lsp.toml`, under the `semantic_tokens` section.
+The syntax for such an entry is
+
+[source,toml]
+----
+[[semantic_tokens]]
+name = "variable"
```
krobelus, line 313: Nit: I think the spec calls this "token type". Maybe we should call it "token" or "token_type"?

src/language_features/semantic_tokens.rs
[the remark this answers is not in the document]
sclu1034: Ah, didn't know about `max_by_key`. Neat.

src/general.rs
sclu1034: I don't think so. The only value that `SemanticTokensWorkspaceClientCapabilities` includes currently is a flag about supporting the request `workspace/semanticTokens/refresh`.

As far as I understand that request, we would need to proxy it to Kakoune and have the editor respond with `textDocument/semanticTokens/full` for all open files.

sclu1034:
Rebased to drop the first two commits that were already cherry-picked into master.

src/language_features/semantic_tokens.rs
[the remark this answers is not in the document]
krobelus: Hold on, did this change the behavior?
Previously we always used the last token-config if they had equal number of matches (because Vec::sort is a stable sort).
I think that's no longer guaranteed (didn't test it though).

So maybe https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold would be the "right" way to implement this..

---

In general I'm not sure what's the best UI for this prioritization feature.
"first wins" / "last wins", or maybe adding an explicit "priority" field. I'm interested in how other editors do it.

krobelus:
Thanks for the updates - looks great!

src/language_features/semantic_tokens.rs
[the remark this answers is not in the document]
sclu1034: > Hold on, did this change the behavior?

It didn't:

> If several elements are equally maximum, the last element is returned
> <https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max_by_key>

---

I don't think there's much of a difference in UX between the three options.

One additional option would be to only match against "token + identical modifiers" and "token + no modifiers". That would be easier to understand, but quite annoying to configure, since some servers will regularly apply 3-5 modifiers, and there are no tools to visualize that (afaik).

krobelus:
Merged as `a3c9589..40a3c9b`, thanks!

PR #491, opened 2021-06-03
Fix download version in installation instructions

caksoylar:
Looks like the second occurrences of v9 in lines were missed in the search-replace.


krobelus:
Whoops, thank you!

PR #496, opened 2021-06-28
Add default configuration for Java via eclipse.jdt.ls

dzamo:


krobelus:
wow this is complicated. They should totally provide a wrapper script.
Anyway, thanks!

dzamo:
It's horribly complicated and ugly ☹️.  And looks like it just needs 
someone to roll up their sleeves and do the scripting over there:

https://github.com/eclipse/eclipse.jdt.ls/issues/748

One day, one day.

PR #502, opened 2021-07-24
add default julia setting in kak-lsp.toml

koehlerson:
I added default settings for julia support. Requires `LanguageServer, StaticLint, SymbolServer` installed globally in julia. Activates the `project=@.` environment. This means that it searches recursively upwards for a julia Project.toml

### DIFF
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -184,6 +184,28 @@ roots = ["package.json"]
 command = "json-languageserver"
 args = ["--stdio"]
 
+# Requires Julia packages "LanguageServer", "StaticLint" and "SymbolServer"
+[language.julia]
+filetypes = ["julia"]
+roots = ["Project.toml", ".git"]
+command = "julia"
+args = [
+    "--startup-file=no",
+    "--history-file=no",
+    "-e",
+    """
+        using LanguageServer;
+        using Pkg;
+        import StaticLint;
+        import SymbolServer;
+        import REPL;
+        env_path = dirname(Pkg.Types.Context().env.project_file);
+        server = LanguageServer.LanguageServerInstance(stdin, stdout, env_path, "");
+        server.runlinter = true;
+        run(server);
+    """,
+]
+
 [language.latex]
 filetypes = ["latex"]
 roots = [".git"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Breaking changes:
 - `lsp-show-message`, which handles `window/showMessage` requests from the server has been removed. See below for the replacement.
 
 Additions:
+- Default configuration for Julia (#502).
 - `lsp-show-message` has been replaced by four separate commands `lsp-show-message-{error,warning,info,log}`.
   The new default implementations log the given messages from the language server to the debug buffer. Important messages are shown in `%opt{toolsclient}`.
 - `lsp-code-actions` use the `menu` command to select an action interactively. The new command `lsp-show-code-actions` can be overridden to customize this behavior (#367).


kak-lsp.toml
```diff
@@ -273,3 +273,26 @@ args = ["--stdio"]
 filetypes = ["zig"]
 roots = ["build.zig"]
 command = "zls"
+
+[language.julia]
```
krobelus, line 277: > Requires LanguageServer, StaticLint, SymbolServer installed globally in julia. 

What about adding something like `# Please install Julia packages "LanguageServer", "StaticLint" and "SymbolServer"` here. I haven't tried what happens if they aren't; maybe the error messages from kak-lsp are good enough, but I think they are only visible in the log.

Actually, would it make sense to install them automatically?
Since we are already running `julia`, we could add a snippet that runs the `Pkg.add` statements for missing packages.
Could that cause problems?

krobelus (commented):
Looks great, thanks!

krobelus:
Thanks!

For some reason the server stopped working on my system. It works in a fresh docker container, but here I get the error below.
Deleting `~/.julia` and reinstalling the language server didn't help. I'm not Julia 1.6.2. Maybe you have an idea?

```
Aug 08 13:06:58.420 ERRO Language server error: [ Info: Couldn't retrieve cache file for Test.
[ Info: Couldn't retrieve cache file for Downloads.
[ Info: Couldn't retrieve cache file for Random.
[ Info: Couldn't retrieve cache file for Artifacts.
[ Info: Couldn't retrieve cache file for NetworkOptions.
[ Info: Couldn't retrieve cache file for Pkg.
[ Info: Couldn't retrieve cache file for UUIDs.
[ Info: Couldn't retrieve cache file for MozillaCACerts_jll.
[ Info: Couldn't retrieve cache file for LibCURL.
[ Info: Couldn't retrieve cache file for Tar.
ERROR: BoundsError: attempt to access 0-element[ Info: Couldn't retrieve cache file for TOML.
[ Info: Couldn't retrieve cache file for REPL.
[ Info: Couldn't retrieve cache file for Dates.
[ Info: Couldn't retrieve cache file for Unicode.
[ Info: Couldn't retrieve cache file for LibSSH2_jll.
[ Info: Couldn't retrieve cache file for Serialization.
[ Info: Couldn't retrieve cache file for Zlib_jll.
[ Info: Couldn't retrieve cache file for Base64.
[ Info: Couldn't retrieve cache file for ArgTools.
 Vector{[ Info: Couldn't retrieve cache file for Libdl.
Any} at index [1:12][ Info: Couldn't retrieve cache file for Logging.
[ Info: Couldn't retrieve cache file for MbedTLS_jll.

Stacktrace:
 [1] checkbounds
[ Info: Couldn't retrieve cache file for p7zip_jll.
   @ ./[ Info: Couldn't retrieve cache file for Printf.
[ Info: Couldn't retrieve cache file for nghttp2_jll.
[ Info: Couldn't retrieve cache file for InteractiveUtils.
[ Info: Couldn't retrieve cache file for SHA.
[ Info: Couldn't retrieve cache file for Sockets.
abstractarray.jl:616 [inlined]
 [2] getindex
   @ ./array.jl:807 [inlined]
 [3] request_julia_config(server::[ Info: Couldn't retrieve cache file for Mmap.
[ Info: Couldn't retrieve cache file for Markdown.
[ Info: Couldn't retrieve cache file for LibCURL_jll.
[ Info: Couldn't retrieve cache file for LibGit2.
LanguageServerInstance, conn::JSONRPC.JSONRPCEndpoint{Base.PipeEndpoint, Base.PipeEndpoint})
   @ LanguageServer ~/.julia/packages/LanguageServer/JrIEf/src/requests/workspace.jl:113
 [4] initialized_notification(params::LanguageServer.InitializedParams, server::LanguageServerInstance, conn::JSONRPC.JSONRPCEndpoint{Base.PipeEndpoint, Base.PipeEndpoint})
   @ LanguageServer ~/.julia/packages/LanguageServer/JrIEf/src/requests/init.jl:195
 [5] (::LanguageServer.var"#96#97"{typeof(LanguageServer.initialized_notification), LanguageServerInstance})(conn::JSONRPC.JSONRPCEndpoint{Base.PipeEndpoint, Base.PipeEndpoint}, params::LanguageServer.InitializedParams)
   @ LanguageServer ~/.julia/packages/LanguageServer/JrIEf/src/languageserverinstance.jl:252
 [6] dispatch_msg(x::JSONRPC.JSONRPCEndpoint{Base.PipeEndpoint, Base.PipeEndpoint}, dispatcher::JSONRPC.MsgDispatcher, msg::Dict{String, Any})
   @ JSONRPC ~/.julia/packages/JSONRPC/yu0G3/src/typed.jl:67
 [7] run(server::LanguageServerInstance)
   @ LanguageServer ~/.julia/packages/LanguageServer/JrIEf/src/languageserverinstance.jl:361
 [8] top-level scope
   @ none:9
, module: kak_lsp::language_server_transport:58
```

kak-lsp.toml
krobelus: (installing automatically is probably a bad idea)

I'm working on a patch that reports the stderr directly in Kakoune. Then we'll directly forward the error from julia which even tells the user how to install the package.

koehlerson:
sorry for the delay!

I think the problem with automatic installing is, that the package would recompile all the time, which would increase the start up time even more, but I guess, this could be handled by a simple if clause. Your error seems pretty weird, as if it has no permissions to read/write stuff. What does your julia REPL do if you insert the code snippet: 

```julia
        using LanguageServer;
        using Pkg;
        import StaticLint;
        import SymbolServer;
        import REPL;
        env_path = dirname(Pkg.Types.Context().env.project_file);
        server = LanguageServer.LanguageServerInstance(stdin, stdout, env_path, "");
        server.runlinter = true;
        run(server);
```
If it's consistent `run` should fail, but let's see

krobelus:
in the interactive console I get no bounds error but still the other messages.
Probably just something weird on my system.

After changing `~/.julia/packages/SymbolServer/rrlGM/src/utils.jl` by adding `@info link` somewhere in `get_file_from_cloud()`, I found that julia is trying to download something like

https://www.julia-vscode.org/symbolcache/store/v1/packages/L/LibSSH2_jll_29816b5a-b9ab-546f-933c-edad1886dfa8/vnothing_nothing.tar.gz

The code that tries to download even when the version is nothing seems wasteful, but it doesn't look wrong.

PR #507, opened 2021-08-08
Overhaul dynamic server configuration syntax

### LINKED ISSUE #234: Does this implement workspace/configuration?
Hey there, a user brought this to my attention, but I'm not 100% sure.
Are you implementing `workspace/configuration`? https://microsoft.github.io/language-server-protocol/specification#workspace_configuration

We got that idea due to this issue.
https://github.com/elm-tooling/elm-language-server/issues/154

### LINKED ISSUE #500: Server configuration syntax is too hard to use
https://github.com/kak-lsp/kak-lsp/issues/367#issuecomment-876435426 shows that it's difficult or impossible to correctly set server initialization options / configuration.
That comment is about `gopls`, which expects an object like `{"formatting.gopls":true}` in initialization / configuration.
It took me a long time to figure out how to do that with `kak-lsp`. It's obvious in hindsight, but there are too many things that can go wrong. Here are some examples:

1. this will send `{"formatting":{"gofumpt":true}}` (wrong, but `gopls` does not complain)
```toml
[language.go.initialization_options]
formatting.gofumpt = true
```

2. this will send `{"formatting.gofumpt":true}` (correct)
```toml
[language.go.initialization_options]
"formatting.gofumpt" = true
```

3. this will send `{"formatting:{"gofumpt":true}}` (wrong)
```kak
set-option -add global lsp_server_initialization_options formatting.gofumpt=true
```

There are multiple issues here
1. [x] Some servers expect structures like 1, others 2. The differences in the TOML between 1 and 2 are way too subtle, we don't want to have users deal with that.

	Will explain this special case in the README, maybe that's enough.

2. [x] With dynamic initialization options, it's currently not possible to please `gopls` here. Approach 3 used to work before 0e6fbd2 but that fixes some Rust language servers. We could allow to write something like `"formatting.gofumpt"=true` but again, that's too subtle.

	Fixed with

	```kak
	set global lsp_config %[
		[language.go.initialization_options]
		gopls."formatting.gofumpt" = true
	]
	```

2. [x] It's not obvious whether `lsp_server_initialization_options` wants a `gopls` (or `go`) prefix. It doesn't, but `lsp_server_configuration` does need the prefix, so this difference is again subtle and non-obvious.

	Explained in docs; we also emphasize that we almost always use initialization_options, which don't need the prefix, and it's now obvious that the "go" prefix is needed.

3. [x] Users might be surprised that `lsp_server_initialization_options` are sent to *all* language servers, not just one of some language.

	Fixed because `[language.<filetype>.initialization_options]` is only sent to the server that corresponds to `<filetype>`. 

4. [x] It took me a long time to figure out which configurations are valid. We should provide more error messages if we can.

Fixed by 8452186 ("Remove lsp-show-message in favor of individual functions for each log level")

5. [x] It's not obvious whether one should use `lsp_server_initialization_options` or `lsp_server_configuration` - it depends on what the language server supports. VSCode users don't have to make that decision. I think we should do the same.

	Better with `[language.<filetype>.initialization_options]`, which supersedes both.

6. [x] AFAIK, `lsp_server_configuration` can not be set in `kak-lsp.toml`

	Fixed now that we can set `[language.<filetype>.initialization_options]` in `%opt{lsp_config}`.

---

Here is a skeleton proposal that solves some problems. JSON strings are terrible to edit by hand, but maybe this is what it takes...

1. Move static server configuration to a single JSON string in `kak-lsp.toml`, to make it easy to compare with VSCode config.

```toml
server_configuration = """
{
  "gopls": {
    "formatting.gofumpt": true
   },
  "rust" {
    "features": ["somefeature"]
  }
}
"""
```

2. Consolidate dynamic server configuration in a single string (merging lsp_server_initialization_options and lsp_server_configuration). This string is another JSON dict, that can override (parts of) the static configuration.

```kak
declare-option str lsp_server_configuration
set-option -add global lsp_server_configuration %{
  {
    "gopls": {
      "formatting.gofumpt": false
     },
  }
}
```
(this should probably use the "rust" configuration from the static configuration)

This will make it trivial to load VSCode's per-workspace `settings.json`

---

Alternatively, we could do the same, but with TOML, then the kak-lsp.toml might not change

```toml
[configuration]
[configuration.gopls]
"formatting.gofumpt" = true
[server_configuration.rust]
features = ["somefeature"]
```

and the dynamic configuration

```kak
declare-option str lsp_server_configuration
set-option -add global lsp_server_configuration %{
  [configuration.gopls]
  "formatting.gofumpt" = false
}
```

krobelus:
Closes #500
Closes #234

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Bug fixes:
 - The suggested config for `rust-analyzer` was fixed for the case that `rustup` is installed but `rust-analyzer` is not installed via `rustup`.
 - Fix spurious cursor movement on `lsp-rename` and `lsp-rename-prompt` (#504).
 - Fix responses to `workspace/configuration` in case there are no initialization options set (#509).
+- When responding to `workspace/configuration`, kak-lsp has already used the configured initialization options, but now ignores the `section` parameter. Consequently, the section should not be included in the initialization options, which makes the options usable for other requests (#507).
 
 Deprecations:
 - `%opt{lsp_server_initialization_options}` and `%opt{lsp_server_configuration}` are deprecated in favor of setting `[language.<filetype>.initialization_options]` in `%opt{lsp_config}`(#500).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Additions:
 - `lsp-show-message` has been replaced by four separate commands `lsp-show-message-{error,warning,info,log}`.
   The new default implementations log the given messages from the language server to the debug buffer. Important messages are shown in `%opt{toolsclient}`.
 - `lsp-code-actions` use the `menu` command to select an action interactively. The new command `lsp-show-code-actions` can be overridden to customize this behavior (#367).
+- `%opt{lsp_config}` allows to set server configuration dynamically (#500).
 - New commands `lsp-{next,previous}-location` generalize `grep-next-match`, `lsp-next-match` and friends (#466).
 - New option `lsp_location_format` to tell `lsp-{next,previous}-location` which "<file>:<line>"-style to match (#466).
 
@@ -21,6 +22,7 @@ Bug fixes:
 - Fix responses to `workspace/configuration` in case there are no initialization options set (#509).
 
 Deprecations:
+- `%opt{lsp_server_initialization_options}` and `%opt{lsp_server_configuration}` are deprecated in favor of setting `[language.<filetype>.initialization_options]` in `%opt{lsp_config}`(#500).
 - `lsp-{goto,symbols}-{next,previous}-match` are deprecated in favor of `lsp-next-location *goto*` and similar (#466).
 
 ## 10.0.0 - 2021-06-03
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -227,6 +227,34 @@ If you are setting any server options via cli, do not forget to append them to
 
 Please let us know if you have any ideas about how to make the default config more sensible.
 
+==== Server-specific configuration
+
+Many servers accept configuration options that are not part of the LSP spec.
+The TOML table `[language.<filetype>.initialization_options]` can contain arbitrary values They will be sent to the server during server initialization, as well as in `workspace/didChangeConfiguration` whenever you change it via the `lsp_config` option. Additionally, it is sent as response to the server request `workspace/configuration`.
+
+If a language server's documentation says it wants a structure like this:
+
+[source=json]
+----
+{
+    "settings": {
+        "rust": {
+            "clippy_preference": "on"
+        },
+    }
+}
+----
+
++...you can do that in `kak-lsp.toml` or in the `lsp_config` option with an equivalent TOML string:
+
+Be aware of TOML serialization rules: for example, if you want to serialize `{"clippy.preference":"on"}`, use `"clippy.preference" = "on"`.
+
+[source=toml]
+----
+[language.rust.initialization_options]
+rust.clippy_preference = "on"
+----
+
 === Configuring Kakoune
 
 kak-lsp's Kakoune integration declares the following options:
@@ -247,29 +275,6 @@ this option.
 language server kak-lsp will ask Kakoune for this option value in the buffer which provoked start of
 the language server. If value is non-empty then it will override `initialization_options` set for
 the buffer's filetype in `kak-lsp.toml`. See below for more information about this option.
-
-The `lsp_server_configuration` and `lsp_server_initialization_options` options are unusual, since
-the language server wants deeply-nested JSON objects, which are hard to represent in Kakoune. If a
-language server's documentation says it wants a structure like this:
-
-[source=json]
-----
-{
-    "settings": {
-        "rust": {
-            "clippy_preference": "on"
-        }
-    }
-}
-----
-
-...you can achieve the same thing in Kakoune with:
-
-----
-set-option -add global lsp_server_configuration rust.clippy_preference="on"
-----
-
-That is, the keys of the `lsp_server_configuration` option are a `.`-delimited path of JSON objects. For implementation reasons, the values use TOML serialisation rules rather than JSON rules, but they're pretty much the same thing for strings, numbers and booleans, which are the most common configuration types.
 
 == Inlay hints for rust-analyzer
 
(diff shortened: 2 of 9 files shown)


sclu1034:
> Keep in mind that the workspace configuration always needs a section key

Did this change recently? As far as I remember, my initial implementation of copying the initialization options did work for the servers that I tested with.
So I'm wondering if this is something that all servers require now or if it is undefined by LSP and servers do different things.

Regardless, if the `[language.<filetype>.workspace_configuration]` really is just "`initialization_options`, but with a prefix" in most cases, would it be possible to add a way to specify something like "use `initialization_options`, but with prefix `gopls`"? Because right now it looks like it's going to require a lot of duplicating settings between the two sections.

krobelus:
In the current state of this PR, these two configurations are functionally equivalent - they make gopls do the same thing:

```toml
[language.go.workspace_configuration]
gopls."formatting.gofumpt" = true

[language.go.initialization_options]
"formatting.gofumpt" = true
```

Previously we required this to send the correct response to `workspace/configuration`

```toml
[language.go.initialization_options]
gopls."formatting.gofumpt" = true
```

However, since we also send this in `initialize`, this caused problems because gopls complains about "gopls" being an invalid key.
Gopls sends an error with `window/displayMessage`, which was only fixed recently.

---

> would it be possible to add a way to specify something like "use initialization_options, but with prefix gopls"?

Yeah actually I was gonna suggest the same (IIUC). This would mean that we'll write

```toml
[language.go.initialization_options]
"formatting.gofumpt" = true
```

and send this during `initialize` as before.
Now her'es the trick: if the language server requests `workspace/configuration` (with section=gopls) we just ignore the section, and return whatever is in `language.go.initialization_options`. This works at least for gopls, which accepts the same options in both `initialize` and `workspace/configuration` (though you sometimes still need to restart it).
I'll check a few other servers (pyls, rls, rust-analyzer, clangd), and if none of them errors we should be fine.

> So I'm wondering if this is something that all servers require now or if it is undefined by LSP and servers do different things.

LSP is generally really underspecified, but at least VSCode is forcing servers to support `workspace/configuration`, according to https://github.com/golang/go/issues/42814#issuecomment-895448535

krobelus:
Ok that won't work either because servers may request multiple options, like the [Julia language server does](https://github.com/julia-vscode/LanguageServer.jl/blob/aa8287aec44e61f858fa4f05b58a4ede7ba72a52/src/requests/workspace.jl#L85-L110)

This basically forces us to use a separate option if we want to be compatible with everything. That might look like

```toml
[workspace_configuration]
gopls."formatting.gofumpt" = true
"julia.format.comments" = false
```

It's a really messy situation - most other LSP clients write custom adapters for each server to make that nicer.
So we're kind of back at the n-times-m problem that LSP was meant to avoid.

I'm tempted to just declare `workspace/configuration` as unsupported, because I haven't seen a real use case except for servers that don't fully support `workspace/didChangeConfiguration` (or implement it by immediately sending `workspace/configuration` like gopls or Julia's LS).
We could just tell users to just restart servers on configuration change. However, from a quick look it looks like the Julia LS only reads settings from `workspace/configuration`, and not from initialization.

sclu1034:
> it looks like the Julia LS only reads settings from `workspace/configuration`, and not from initialization.

The same applies to https://github.com/sumneko/lua-language-server, which is why I implemented that request initially. And others as well, judging by #234.
My initial idea of adding an option that implements "use `initialization_options`, but with prefix `gopls`" would at least prevent the need for duplicating configuration between `initialization_options` and `workspace_configuration`.

However, as far as I remember VSCode, it only ever has one set of configuration options. So either they expect the extension developers to handle the split between initialization and configuration manually, or there actually is a way to do it automatically for all servers, and we just haven't found it yet.

krobelus:
>  So either they expect the extension developers to handle the split between initialization and configuration manually

yeah, I'm pretty sure

Of course we could use some heuristics to try to support current servers:

If we used a similar format to VSCode

```toml
[settings]
gopls."formatting.gofumpt" = true
"julia.format.comments" = false
"Lua.color.mode" = "Grammar"
rust-analyzer.cargo.features = ["some_feature_only_visible_to_rust_analyzer"]
rust.features = ["some_feature_only_visible_to_rls"]
```

then we could make that work with
- let $langid be the name of the language server, or the language ID as fallback
- the settings for $langid are all options that either
  - are wrapped in a $langid object (dropping the $langid)
    - like `gopls."formatting.gofumpt" = true` -> `"formatting.gofumpt" = true`
  - or, whose key starts with $langid (keeping the $langid prefix)
    - like `"julia.format.comments" = false`
    - case-insensitive, because Lua
- on initialize, send all settings for the given language server
- on workspace/configuration, send all settings for the given parameters

Of course that's super ugly, because it doesn't fit the current configuration
format. It would be much nicer if we could somehow convince everyone
to behave like gopls/rust-analyzer, who just have a single object
that can be sent in both initialization and `workspace/configuration`.

---

I guess we could compromise and add a "sections"-parameter to keep
initialization_options clean

```toml
[language.lua]
filetypes = ["lua"]
roots = [".git"]
command = "lua-language-server"
section = [] # The default - initialization_options are directly used as sections for workspace/configuration
[language.lua.initialization_options]
"Lua.color.mode" = "Grammar"

[language.go]
filetypes = ["go"]
roots = ["Gopkg.toml", "go.mod", ".git", ".hg"]
command = "gopls"
sections = ["gopls"] # When answering to workspace/configuration, only send section "gopls".
[language.go.initialization_options]
"formatting.gofumpt" = true
```

Of course that will fail again when a server mixes the two styles (I can't
think of one ATM).

sclu1034:
From a user perspective, I do like the idea of mirroring VSCode's style, where the user only needs one `settings` object.
And thanks to TOML, it can even be more orderly than VSCcode:

```toml
[settings.julia]
"format.comments" = true
"format.i_dont_know_other_options" = false

[settings.rust_analyzer]
"cargo.feature" = ["some_feature_only_visible_to_rust_analyzer"]
```

The `section` from `workspace/configuration` would then be an index into `settings`.

And for the servers that do need initialization options, we could turn things around and say `language.julia.section = "julia"` means send the content of `settings.julia` as `initialization_options`. And since these are unlikely to change over time, those can be added to the default `kak-lsp.toml` already.

I like that better, since it disconnects implementation details. I doubt many users really know what `initialization_options` or `workspace/configuration` are/mean, they just care about "where do I put settings?".

---

Do we know the extent of differences between LSP implementations? If it's only two or three different ways how they handle settings across all major servers, it might be feasible to implement all three and switch with a setting in `language.<id>`.

krobelus:
I've merged the %opt{lsp_config} part as 46e33e5~4..46e33e5

I've implemented something like your suggestion in #511. I've checked configuration options for all default language servers but tested only a few.

I went for `[language.rust.settings."rust-analyzer"]` over `[settings."rust-analyzer"]`.
The latter would require is to put all settings for all languages in a single table, separate from language configs.

> we could turn things around and say language.julia.section = "julia"

Right, this is a small price to pay - just one line of ugly LSP details per language. If we care about aesthetics we could also hardcode that part inside kak-lsp..

> I like that better, since it disconnects implementation details. I doubt many users really know what initialization_options or workspace/configuration are/mean, they just care about "where do I put settings?".

Yeah, that's a great point.

PR #511, opened 2021-08-25
Mimic VSCode server settings to support workspace/configuration

### LINKED ISSUE #234: Does this implement workspace/configuration?

krobelus:
"settings" replaces "initialization_options".
It can be set in kak-lsp.toml, or dynamically in %opt{lsp_config}.

"settings" uses the same structure as VSCode's settings.json. This
means we can properly implement workspace/configuration.

Initialization options and parameters for didChangeConfiguration are
computed by looking up a server-specific section. This seems to match
the expected behavior for all language servers I tried.

Includes configuration for all default language servers (largely
untested). These are default configurations according to the servers'
documentation. I'm not sure how to handle these in future;
- leaving them uncommented has the upside that we'll quickly find errors
- however, the downside is that we shadow the server's actual default
  configuration
- additionally these lists might go out of date quickly

I'm leaning towards only including (commented) example settings
for most servers. Users will get an idea of how it works and can
consult server docs for all available settings.
Perhaps we can include uncommented default settings for "important"
servers.

TODO maybe keep initialization_options around for one release cycle?

Closes #234, #507

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 ## Unreleased
 
 Breaking changes:
-- The configuration syntax for `semantic_tokens` has changed. See the updated `kak-lsp.toml` for an example (#488).
+- Two breaking changes to the configuration format - see the updated `kak-lsp.toml` for examples. 
+  - `semantic_tokens` syntax has changed (#488)
+  - `settings` replaces `initialization_options` for server-specific configuration  (#511)
 - Snippet support has been disabled by default, as a workaround for conflicts with Kakoune's built-in completion (#282).
 - `lsp-show-message`, which handles `window/showMessage` requests from the server has been removed. See below for the replacement.
 - Hidden commands `lsp-{next,previous}-match` were removed in favor of `lsp-{next,previous}-location` (#466).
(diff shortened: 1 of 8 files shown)


sclu1034 (commented):
Works fine for me with `rust-analyzer`. Based on its logs, both `initialization_options` and `workspace/configuration` seem to be sent correctly.

I would second keeping a few commented example settings to show the syntax, rather than a full set of uncommented default options. The latter would require both maintainers here and all users that copy the config to keep them updated individually.

However, for some servers, it might make sense to set a few specific settings for everyone. E.g. `kak-lsp` doesn't support `rust-analyzer`'s "Hover Actions", so we could disable that by default.

krobelus:
Thanks - I've trimmed down the config to include just one commented example setting for each language server.

I've also named the section parameter `settings_section` because server settings is the general term.

Let's see how this goes - I'm cautiously optimistic that we don't need more breaking changes for server configuration.
Seems like a good time to make a release soon. I guess we'll remove `lsp_server_initialization_options` and `lsp_server_settings` in a later release.

> E.g. kak-lsp doesn't support rust-analyzer's "Hover Actions", so we could disable that (rust-analyzer.hoverActions.enable = false) by default.

Ok I've disabled that

PR #515, opened 2021-09-07
Consolidate inlay diagnostics

raiguard:
Currently, inlay diagnostics are shown all in a row, one after the other. This makes them very difficult to digest when there are a lot of them.

Neovim's new built-in LSP client does inlay diagnostics in a way that I immediately loved, so I have implemented it here. The following changes were made:

- Added a five-space gap between the actual line and the diagnostics (this can be changed via the `lsp_inlay_diagnostic_gap` option)
- Added a symbol (`lsp_inlay_diagnostic_sign`) that is shown for each diagnostic on the line, and is colored accordingly
- Only the first line of the first diagnostic is printed

**Before:**

image

**After:**

image

I'm still fairly new to Rust, so I fully expect my code to be ripped apart and destroyed. It works just fine, but it turns what was a single pass into two passes, and makes a bunch of heap allocations that weren't there before. There's bound to be ways to improve it.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -70,6 +70,9 @@ declare-option -docstring "Character to signal an error in the gutter" str lsp_d
 declare-option -docstring "Character to signal a hint in the gutter" str lsp_diagnostic_line_hint_sign '-'
 declare-option -docstring "Character to signal an info in the gutter" str lsp_diagnostic_line_info_sign 'i'
 declare-option -docstring "Character to signal a warning in the gutter" str lsp_diagnostic_line_warning_sign '!'
+# Visual settings for inlay diagnostics
+declare-option -docstring "Character to represent a single inlay diagnostic of many on a line. May not contain '|'" str lsp_inlay_diagnostic_sign '■'
+declare-option -docstring "Character(s) to separate the actual line contents from the inlay diagnostics. May not contain '|'" str lsp_inlay_diagnostic_gap '     '
 # Another good default:
 # set-option global lsp_diagnostic_line_error_sign '▓'
 # set-option global lsp_diagnostic_line_warning_sign '▒'
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -5,7 +5,7 @@ use crate::util::*;
 use itertools::Itertools;
 use jsonrpc_core::Params;
 use lsp_types::*;
-use std::collections::HashSet;
+use std::collections::HashMap;
 
 pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
     let params: PublishDiagnosticsParams = params.parse().expect("Failed to parse params");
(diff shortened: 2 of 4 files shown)


sclu1034:
I'm not sure if we should add a separate indicator for every diagnostics entry on a line. `clangd` regularly produces 5-10 messages for a single error.
I think it might make more sense to treat the indicators as "there is at least one message of severity X". I.e. even if there are 5 error messages on a line, there would be only one red indicator.
If we do want to visualize the number of messages, [this](https://github.com/mawww/kakoune/issues/2727#issuecomment-853717947) might be a way to do it.

Additionally, if there are diagnostics for multiple severities on a line, I'd make sure that a message from the highest severity is displayed.
E.g. in one of your screenshots in #516, you have a line with both an error and a hint, but the hint's message is displayed.

raiguard:
Does `clangd` make ten _separate errors_, or does it have ten lines on a single error? Either way, that seems kinda crazy. All of the language servers I use have pretty concise errors, and I think being able to see the amount of each would be better in the long run. Worst case, I could add an option to choose your preference.

I agree with the diagnostic severity though, that's a good idea.

sclu1034:
Once you know how good modern compilers are with errors (`rustc` being _excellent_ in that regard), you wonder where those decades of development went in tools like `gcc` or `clang`.

E.g. these three warnings don't tell me anything about the fact that I need to `std::move` the value:
image

However, while trying to reproduce some errors to show, I did realize that clangd's messages actually contain long chains of "caused by X, caused by Y"-sections, which I might have mistaken for separate entries in the hover info box. But I remember info boxes filling the better part of my screen, so five entries on a single position still sounds plausible.

I guess whether the amount matters depends on preference/workflow.
For me, seeing the amount wouldn't change much. Either the visible error message gives me enough information to start editing, which invalidates all messages, or it doesn't, and I need to check the hover info. I don't think seeing a number of errors would make me always check all them before I start editing.

Making it configurable would definitely cover both options.

raiguard:
image

I made the change to show the most severe diagnostic text, but I'm not happy with the implementation. It feels ugly.

src/diagnostics.rs
```diff
@@ -57,35 +57,77 @@ pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
...
+            .entry(line_number)
+            .or_insert(LineDiagnostics {
+                range_end: x.range.end,
+                symbols: String::new(),
+                text: String::new(),
+                text_face: String::new(),
+                text_severity: None,
+            });
+
+        if let Some(severity) = x.severity {
```
raiguard, line 81: I am not happy with this code _at all_. This is what I would do in Lua, but it feels really weird in Rust. There has got to be a more idiomatic way to do these checks.

raiguard:
image

A thing I've noticed (this happens with the old diagnostics as well) is that the inlay diagnostics are printed before the inlay type hint at the end of a line. Is this something we can fix?

src/diagnostics.rs
raiguard: Remembered `map_or`, which made this much cleaner. I knew there was a better solution somewhere.

krobelus (approved):
This looks awesome! I haven't been using inlay diagnostics because they are still a bit buggy (or is it the replace-ranges highlighter?), but I'm eager to try them on again.

Showing the (first) most severe error sounds perfect.

> A thing I've noticed (this happens with the old diagnostics as well) is that the inlay diagnostics are printed before the inlay type hint at the end of a line. Is this something we can fix?

probably something about the order in which hooks are defined

another unrelated problem: as I type something like `let a = 123` I get inlay errors, because the statement is obviously incomplete; and my cursor is right of the inlay error, which is weird and distracting. Is there a way to disable it in insert mode?

krobelus:
I guess we can elide the computation of `diagnostic_ranges` if the highlighter is not enabled?
No need to do that now though.

rc/lsp.kak
```diff
@@ -61,13 +61,16 @@ declare-option -docstring "DEPRECATED, use %opt{lsp_config}. Configuration to se
 # Line flags for inline diagnostics.
 declare-option -docstring "Character to signal an error in the gutter" str lsp_diagnostic_line_error_sign '*'
 declare-option -docstring "Character to signal a warning in the gutter" str lsp_diagnostic_line_warning_sign '!'
+# Visual settings for inlay diagnostics
+declare-option -docstring "Character to represent a single inlay diagnostic of many on a line" str lsp_inlay_diagnostic_sign '■'
+declare-option -docstring "Character(s) to separate the actual line contents from the inlay diagnostics" str lsp_inlay_diagnostic_gap '     '
```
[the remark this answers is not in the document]
raiguard: The most errors on a single line I've seen so far has been five, and that was a specially crafted case. Usually there's no more than three.

src/types.rs
```diff
@@ -235,3 +235,11 @@ pub enum OffsetEncoding {
     #[serde(rename = "utf-16")]
     Utf16,
 }
+
+// An intermediate representation of the diagnostics on a line, for use with inlay diagnostics
+pub struct LineDiagnostics {
+    pub range_end: Position,
+    pub symbols: String,
+    pub text: String,
+    pub text_face: String,
```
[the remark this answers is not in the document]
raiguard: Using `&'static str` breaks the symbol generation because of `push_str`.

raiguard:
> another unrelated problem: as I type something like let a = 123 I get inlay errors, because the statement is obviously incomplete; and my cursor is right of the inlay error, which is weird and distracting. Is there a way to disable it in insert mode?

I actually have this in my kakrc:
```kakounescript
def inlay-diagnostics-enable %{
    lsp-inlay-diagnostics-enable window
    hook window -group inlay-diagnostics ModeChange (push|pop):.*:insert %{
        lsp-inlay-diagnostics-disable window
    }
    hook window -group inlay-diagnostics ModeChange (push|pop):insert:.* %{
        lsp-inlay-diagnostics-enable window
    }
}

def inlay-diagnostics-disable %{
    lsp-inlay-diagnostics-disable window
    remove-hooks window inlay-diagnostics
}
```

I then call `inlay-diagnostics-enable` either manually or in my config.

src/types.rs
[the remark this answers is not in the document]
krobelus: oh I meant just `text_face`, since it can only be one of a fixed set (could also make it an enum but no need)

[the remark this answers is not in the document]
raiguard: Ah, I see, my bad.

raiguard:
image

I'm noticing that the "print most severe diagnostic" idea should carry over to the actual inline diagnostics. For instance, this cyan underline should be red. However, that will take a lot more work to sort out and is out of scope for this PR.

sclu1034:
> inlay diagnostics [...] are still a bit buggy (or is it the replace-ranges highlighter?)
> my cursor is right of the inlay error

It's `replace-ranges`, which definitely isn't perfect yet. It can't be combined with the `wrap` highlighter at all, it can break Kakoune's "primary cursor should always be in view" paradigm and cursor movement around replaced ranges, especially around those that "replace nothing", is still a bit weird.

I also use a snippet similar to the above to hide both inlay diagnostics and inlay hints in edit mode.

src/diagnostics.rs
```diff
@@ -57,35 +57,69 @@ pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
...
+                    .unwrap_or_default()
+                    .replace("|", "\\|");
+                line_diagnostics.text_face = face;
+                line_diagnostics.text_severity = diagnostic.severity;
+            }
+        }
+
+        line_diagnostics
+            .symbols
+            .push_str(&format!("{{{}}}%opt[lsp_inlay_diagnostic_sign]", face))
```
sclu1034, line 98: The `if` block ignores everything that doesn't have a severity, thereby dropping the message, but the statement below adds an indicator for that ignored message regardless.

Maybe we should apply the same severity default as above:
```rust
let severity = diagnostic.severity.unwrap_or(DiagnosticSeverity::Warning);
```

src/diagnostics.rs
```diff
@@ -57,35 +57,58 @@ pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
...
-            let sep = if lines_with_errors.insert(line) {
-                ""
-            } else {
-                ", "
-            };
-            editor_quote(&format!(
-                "{}+0|{{{}}}{{\\}}{} {}",
+
+            format!(
+                "\"{}+0|%opt[lsp_inlay_diagnostic_gap]{} {{{}}}{{\\}}{}\"",
```
krobelus, line 114: We'll get an error if `%opt{lsp_inlay_diagnostic_gap}` contains a `|`.
I guess we don't care but maybe there's an easy fix.

raiguard: I added a note specifying that `|` isn't allowed.

raiguard:
image

Lol, this is the most errors I've seen in the new format thus far.

krobelus:
ugh that's a lot of space - maybe we should truncate the symbols after like 3?

raiguard:
I think it's fine, if you have 15 errors on a single line, that signifies that there is something _very wrong_ (or you have a merge conflict). You'll almost never see that many.

krobelus:
Ok I just tried it and it doesn't seem to bother me much for merge conflicts. We'll see

> Alternatively, I could add a lsp_inlay_diagnostic_max_symbols option (default 100) that will truncate any symbols beyond that count.

Yeah, we can try something like that in future, but let's see if we really need it

PR #516, opened 2021-09-08
Add hint and info diagnostics

raiguard:
It seemed odd to me that `kak-lsp` showed hints and infos as warnings instead of what they actually are, so I added hints and infos properly. Infos are colored blue by default, and hints are cyan (I wanted to make them grey, but there isn't a standard grey color).

I also took the opportunity to split the `LineFlagErrors` option into `LineFlag(Error|Warning|Info|Hint)`. The `flag-lines` highlighter now uses the `default` face, and `kak-lsp` adds the appropriate markup strings to color them.

The result is as thus:

image

(pictured with the improved inlay diagnostics from #515, and my [one-darker](https://github.com/raiguard/one.kak) color scheme)

The new diagnostic types work in the diagnostics buffer as well:

image

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -1162,13 +1162,13 @@ define-command lsp-find-error -params 0..2 -docstring "lsp-find-error [--previou
 Jump to the next or previous diagnostic error" %{
     evaluate-commands %sh{
         previous=false
-        errorCompare="DiagnosticError"
         if [ "$1" = "--previous" ]; then
             previous=true
             shift
         fi
+        includeWarnings=false
         if [ "$1" = "--include-warnings" ]; then
-            errorCompare="Diagnostic"
+            includeWarnings=true
         fi
         #expand quoting, stores option in $@
         eval set -- "${kak_quoted_opt_lsp_errors}"
@@ -1178,7 +1178,9 @@ Jump to the next or previous diagnostic error" %{
         prev=""
         selection=""
         for e in "$@"; do
-            if [ -z "${e##*${errorCompare}*}" ]; then # e contains errorCompare
+            if [ -z "${e##*DiagnosticError*}" ] || {
+                $includeWarnings && [ -z "${e##*DiagnosticWarning*}" ]
+            } then # e is an error or warning
                 e=${e%,*}
                 line=${e%.*}
                 column=${e#*.}
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -8,12 +8,19 @@ declare-option -docstring "Command with which lsp is run" str lsp_cmd "kak-lsp -
 
 # Faces used by inline diagnostics.
 set-face global DiagnosticError red
+set-face global DiagnosticHint default
+set-face global DiagnosticInfo default
 set-face global DiagnosticWarning yellow
 # Faces used by inlay diagnostics.
 set-face global InlayDiagnosticError DiagnosticError
+set-face global InlayDiagnosticHint DiagnosticHint
+set-face global InlayDiagnosticInfo DiagnosticInfo
 set-face global InlayDiagnosticWarning DiagnosticWarning
-# Line flags for errors and warnings both use this face.
-set-face global LineFlagErrors red
+# Faces used by line flags
+set-face global LineFlagError red
+set-face global LineFlagHint default
+set-face global LineFlagInfo default
+set-face global LineFlagWarning yellow
 # Face for highlighting references.
 set-face global Reference MatchingChar
 set-face global ReferenceBind +u@Reference
@@ -60,6 +67,8 @@ declare-option -docstring "DEPRECATED, use %opt{lsp_config}. Configuration to se
 declare-option -docstring "DEPRECATED, use %opt{lsp_config}. Configuration to send in initializationOptions of Initialize messages." str-to-str-map lsp_server_initialization_options
 # Line flags for inline diagnostics.
 declare-option -docstring "Character to signal an error in the gutter" str lsp_diagnostic_line_error_sign '*'
+declare-option -docstring "Character to signal a hint in the gutter" str lsp_diagnostic_line_hint_sign '-'
+declare-option -docstring "Character to signal an info in the gutter" str lsp_diagnostic_line_info_sign 'i'
 declare-option -docstring "Character to signal a warning in the gutter" str lsp_diagnostic_line_warning_sign '!'
 # Another good default:
 # set-option global lsp_diagnostic_line_error_sign '▓'
@@ -67,7 +76,7 @@ declare-option -docstring "Character to signal a warning in the gutter" str lsp_
 # This is used to render lsp-hover response.
 # By default it shows both hover info and diagnostics.
 declare-option -docstring "Format hover info" str lsp_show_hover_format 'printf ''%s\n\n%s'' "${lsp_info}" "${lsp_diagnostics}"'
-# If you want to see only hover info, try 
+# If you want to see only hover info, try
 # set-option global lsp_show_hover_format 'printf %s "${lsp_info}"'
 declare-option -docstring %{Defines location patterns for lsp-next-location and lsp-previous-location.
 Default locations look like "file:line[:column][:message]"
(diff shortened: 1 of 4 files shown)


src/diagnostics.rs
```diff
@@ -30,14 +30,18 @@ pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
                 lsp_range_to_kakoune(&x.range, &document.text, ctx.offset_encoding),
                 match x.severity {
                     Some(DiagnosticSeverity::Error) => "DiagnosticError",
+                    Some(DiagnosticSeverity::Hint) => "DiagnosticHint",
+                    Some(DiagnosticSeverity::Information) => "DiagnosticInfo",
                     _ => "DiagnosticWarning",
```
sclu1034, line 35: Since we're now implementing all possible values, the last arm should be explicit as well. That way, we'll get a compiler error if there ever was a change to this in an update of `lsp_types`.
```suggestion
                    Some(DiagnosticSeverity::Warning) => "DiagnosticWarning",
```

src/diagnostics.rs
```diff
@@ -47,11 +51,19 @@ pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
...
+                    }
+                    Some(DiagnosticSeverity::Hint) => {
+                        hint_count += 1;
+                        "{LineFlagHint}%opt[lsp_diagnostic_line_hint_sign]"
+                    }
+                    Some(DiagnosticSeverity::Information) => {
+                        info_count += 1;
+                        "{LineFlagInfo}%opt[lsp_diagnostic_line_info_sign]"
                     }
                     _ => {
```
sclu1034, line 64: Same as above.
```suggestion
                    Some(DiagnosticSeverity::Warning) => {
```

src/diagnostics.rs
```diff
@@ -63,6 +75,8 @@ pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
         .map(|x| {
             let face = match x.severity {
                 Some(DiagnosticSeverity::Error) => "InlayDiagnosticError",
+                Some(DiagnosticSeverity::Hint) => "InlayDiagnosticHint",
+                Some(DiagnosticSeverity::Information) => "InlayDiagnosticInfo",
                 _ => "InlayDiagnosticWarning",
```
sclu1034, line 80: Same as above.
```suggestion
                Some(DiagnosticSeverity::Warning) => "InlayDiagnosticWarning",
```

src/diagnostics.rs
```diff
@@ -136,6 +154,8 @@ pub fn editor_diagnostics(meta: EditorMeta, ctx: &mut Context) {
                         p.column,
                         match x.severity {
                             Some(DiagnosticSeverity::Error) => "error",
+                            Some(DiagnosticSeverity::Hint) => "hint",
+                            Some(DiagnosticSeverity::Information) => "info",
                             _ => "warning",
```
sclu1034, line 159: Same as above.
```suggestion
                            Some(DiagnosticSeverity::Warning) => "warning",
```

rc/lsp.kak
```diff
@@ -8,12 +8,19 @@ declare-option -docstring "Command with which lsp is run" str lsp_cmd "kak-lsp -
...
+set-face global InlayDiagnosticHint DiagnosticHint
+set-face global InlayDiagnosticInfo DiagnosticInfo
 set-face global InlayDiagnosticWarning DiagnosticWarning
-# Line flags for errors and warnings both use this face.
-set-face global LineFlagErrors red
+# Faces used by line flags
+set-face global LineFlagError red
+set-face global LineFlagHint cyan
+set-face global LineFlagInfo blue
+set-face global LineFlagWarning yellow
```
sclu1034, line 23: I'm not sure if the "info" and "hint" faces should be coloured by default. After all, Kakoune doesn't even have a `Warning` face.

Not everyone wants their editor to pop in every colour of the rainbow, and these severities aren't as ubiquitous as something like "error == red".

sclu1034 (changes requested):

src/diagnostics.rs
raiguard: I had kept it because one of the possible patterns is `None`, but you're right that it would just happily classify new diagnostic levels as warnings. I'll add an explicit case to convert `None` into warnings as well.

rc/lsp.kak
raiguard: I think avoiding ambiguity is probably more important than avoiding color vomit. LSP servers are going to throw out any of the four diagnostic levels, and being able to know which one it is is pretty important. I will change hints to a dim white though, so it'll decrease the color pop a bit.

src/diagnostics.rs
sclu1034: Oh, right. I overlooked the fact that it's an `Option<DiagnosticSeverity>`.

rc/lsp.kak
sclu1034: It's definitely important for errors and warnings. And uncoloured would make "info" and "hint" visually different from the other two already. But I don't think the difference between those two is big enough that more colours are necessary.
I'd be willing to bet that even most servers don't have a clear distinction whether a message goes into "info" or "hint".

raiguard: Fair enough, I'll change it then. I'm going to keep them colored in my personal colorscheme, but for defaults, you're probably correct.

raiguard: image

Done (with old inlay diagnostics)

krobelus: I agree that it's safer to go with the default face.
If we can find some colors that work with the default colorschemes for most languages, then we can of course add them.

blue/cyan would be nice but Python uses blue for keywords, and Lua uses cyan for function names. It's a bit confusing if (unrelated) gutter chars have the same color.

It looks like the lua language server's info/hint diagnostics are like warnings, so in that case yellow would be fine.

krobelus (commented):
Looks really great, thanks!

rc/lsp.kak
krobelus: using yellow (at least for the gutter text) would also have that advantage that it's obvious that the gutter text is not part of the file

rc/lsp.kak
```diff
@@ -61,13 +68,15 @@ declare-option -docstring "DEPRECATED, use %opt{lsp_config}. Configuration to se
 # Line flags for inline diagnostics.
 declare-option -docstring "Character to signal an error in the gutter" str lsp_diagnostic_line_error_sign '*'
 declare-option -docstring "Character to signal a warning in the gutter" str lsp_diagnostic_line_warning_sign '!'
+declare-option -docstring "Character to signal a hint in the gutter" str lsp_diagnostic_line_hint_sign '-'
+declare-option -docstring "Character to signal an info in the gutter" str lsp_diagnostic_line_info_sign 'i'
```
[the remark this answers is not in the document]
raiguard: I tried various "info" symbols, but they all looked bad. I think 99.99% of programmers who are using `kak-lsp` will probably have line numbers, so it shouldn't be much of a problem.

rc/lsp.kak
krobelus: Ok I somehow deleted my earlier commit because it appeard twice in GitHub..

Anyway, I just realized that the c_cpp filetype uses yellow for types, that kind of sucks..
I noticed rust-analyzer gives hints like "prefix variable with underscore" and so on, so yellow/warning wouldn't really here

rc/lsp.kak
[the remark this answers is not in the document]
krobelus: > I think 99.99% of programmers who are using kak-lsp will probably have line numbers

Hmm I almost never use line numbers (not sure what's the use outside pair programming).
I think we should always keep the default config working.
But probably the cases where it matters are scarce (haven't seen any besides this lua case) so we can try this for a while.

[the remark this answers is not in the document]
raiguard: Lol, that's what I get for saying "what are the chances".

I do think the cases where it'll be a problem are rare enough that we could leave it. I haven't found a symbol that "looks" better than the plain `i`.

sclu1034 (approved):

PR #519, opened 2021-09-12
Fix diagnostic jumping behaviour

Gskartwii:
Prior to this patch, attempting to jump to a file from the diagnostics buffer using an absolute path would result in an error such as: `File does not exist: /home/user/project//home/user/project/src/lib.rs`. The buggy behaviour is a result of interaction of numerical registers with the `s` key. This PR fixes the behaviour by immediately moving the `lsp_location_format` regex matches to `abcd` registers, where they can be safely stored. The original contents are, of course, restored after kak-lsp is done with the jump.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -1698,29 +1698,36 @@ define-command -hidden lsp-make-register-relative-to-root %{
             # Is it an absolute path?
             execute-keys s\A/.*<ret>
         } catch %{
-            set-register 1 "%opt{lsp_project_root}%reg{1}"
+            set-register a "%opt{lsp_project_root}%reg{a}"
         }
     }
 }
 
 define-command -hidden lsp-jump %{ # from grep.kak
-    evaluate-commands %{ # use evaluate-commands to ensure jumps are collapsed
+    evaluate-commands -save-regs abc %{ # use evaluate-commands to ensure jumps are collapsed
         try %{
             execute-keys "<a-x>s%opt{lsp_location_format}<ret>"
+            set-register a "%reg{1}"
+            set-register b "%reg{2}"
+            set-register c "%reg{3}"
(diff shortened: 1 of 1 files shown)


rc/lsp.kak
```diff
@@ -1698,29 +1698,36 @@ define-command -hidden lsp-make-register-relative-to-root %{
...
+    evaluate-commands -save-regs abc %{ # use evaluate-commands to ensure jumps are collapsed
         try %{
             execute-keys "<a-x>s%opt{lsp_location_format}<ret>"
+            set-register a "%reg{1}"
+            set-register b "%reg{2}"
+            set-register c "%reg{3}"
             lsp-make-register-relative-to-root
             set-option buffer grep_current_line %val{cursor_line}
-            evaluate-commands -try-client %opt{jumpclient} -verbatim -- edit -existing %reg{1} %reg{2} %reg{3}
+            evaluate-commands -try-client %opt{jumpclient} -verbatim -- edit -existing %reg{a} %reg{b} %reg{c}
```
krobelus, line 1715: since we quoted `%reg{3}` above: if there is no column, we now pass an empty string instead of nothing. Kakoune doesn't seem to mind, so it's fine.
We could probably add a `<space>` at line 1709, then there should be no need to quote?

krobelus (approved):
Wow, 
Thanks for the fix!

rc/lsp.kak
Gskartwii: Ah, that's a nice catch. What would the space do?

Thanks for merging!

krobelus: Space would remove all selections but the main one.
Currently, when there are multiple selections we always jump to the last one.
Though this case probably doesn't matter so I'll leave it for now.

PR #525, opened 2021-09-17
Use LineNumbers face for diagnostic flag-lines

### LINKED ISSUE #524: Diagnostic flag-lines shouldn't use default face
There is a noticeable white gap at left most side because kak-lsp's diagnostic flag-lines face doesn't match with colorscheme's face `LineNumbers`.

Screen Shot 2021-09-17 at 22 47 28

vbauerster:
I reckon diagnostic flag-line face should default to `LineNumbers` face to match any colorscheme.

Fixes #524


krobelus:
Great fix, thanks!
Probably other scripts that use flag-lines highlighters should do the same (clang/git/lint)

PR #529, opened 2021-09-20
Reset hover diagnostics face

### LINKED ISSUE #528: Diagnostics in hovers have their coloring spill onto the next ilne
image

As you can see, the warning's yellow coloring spills onto the hint line. I think it would make sense to use the `bullet` face for the bullets.

sclu1034:
Fixes #528.

### DIFF
--- a/src/language_features/hover.rs
+++ b/src/language_features/hover.rs
@@ -63,11 +63,12 @@ pub fn editor_hover(
 
                     if !x.message.is_empty() {
                         Some(format!(
-                            "• {{{}}}{}",
+                            "• {{{}}}{}{{{}}}",
                             face,
                             escape_brace(x.message.trim())
                                 // Indent line breaks to the same level as the bullet point
                                 .replace("\n", "\n  "),
+                            FACE_INFO_DEFAULT,
                         ))
                     } else {
                         None


sclu1034:
> I think it would make sense to use the `bullet` face for the bullets.

If we do want a non-default face for the bullet points, it should be a separate one. `bullet` is intended for buffer content, which usually has a different background colour than the info box, and `InfoBullet` is mostly intended for the rendered markup.

However, instead of allowing customization for just the face, we could also make the whole symbol configurable. If we did something like `config.diagnostics_list_symbol.replace("{i}", index)`, we could even allow for ordered lists.

krobelus:
Thanks for the fix, looks great.

> However, instead of allowing customization for just the face, we could also make the whole symbol configurable.

A config option like `lsp_info_diagnostic_bullet` or `lsp_hover_diagnostic_bullet` that defaults to `•` sounds reasonable. It could be an arbitrary markup string, though that's inconsistent with  `lsp_diagnostic_line_error_sign`+`LineFlagError` and friends.

In general, we should probably try to reduce the number of config options because they complicate things for the average user.
Of course everything should remain hackable (for example by overriding lsp-* commands), and all colors should be themeable.

raiguard:
@sclu1034 My apologies, I actually meant the `InfoBullet` face, which I already have configured to match my regular `bullet` face.

sclu1034:
@krobelus What are your thoughts on enabling GitHub Discussions for the repo? Instead of continuing the discussion in a related and merged PR, it might make sense to move discussion of the points you've mentioned there.

PR #531, opened 2021-09-21
Improve CI test workflow

sclu1034:
- Cache Cargo dependencies
- Cache Kakoune build
- Run different tests in parallel
- Improve reproducibility and stability

Partially [inspired by matklad](https://matklad.github.io/2021/09/04/fast-rust-builds.html).

You can find example logs here: https://github.com/sclu1034/kak-lsp/actions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,10 +3,8 @@ name: Test kak-lsp
 on: [push, pull_request]
 
 env:
-  CARGO_TERM_COLOR: always
-  # CI builds are closer to a from-scratch build anyways
-  # and incremental compilation hurts reproducibility
   CARGO_INCREMENTAL: 0
+  CARGO_TERM_COLOR: always
 
 jobs:
   unit-tests:
@@ -69,21 +67,22 @@ jobs:
           ${{ github.workspace }}/kakoune/src/.*.opt.o
 
     - name: Build Kakoune
-      if: steps.cache-kakoune.outputs.cache-hit != 'true'
       working-directory: ${{ github.workspace }}/kakoune/src
       run: |
-        if [ "${{ matrix.os }}" = "macos-latest" ]; then
-          n=$(sysctl -n hw.logicalcpu)
+        # The source files created by actions/checkout are newer than the
+        # output extracted from cache. So we need to touch these targets
+        # to tell Make that they are indeed up-to-date.
+        if [ "${{ steps.cache-kakoune.outputs.cache-hit }}" = "true" ]; then
+          make -t
         else
-          n=$(nproc)
+          if [ "${{ matrix.os }}" = "macos-latest" ]; then
+            n=$(sysctl -n hw.logicalcpu)
+          else
+            n=$(nproc)
+          fi
+
+          make -j$n
         fi
-
-        make -j$n
-
-    - name: Touch build targets
-      if: steps.cache-kakoune.outputs.cache-hit
-      working-directory: ${{ github.workspace }}/kakoune/src
-      run: make -t
 
     - name: Install Kakoune
       working-directory: ${{ github.workspace }}/kakoune/src
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,11 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest]
+        kakoune_version:
+          # Empty ref to use the repo's default branch
+          - ''
+          # Latest release
+          - 'v2021.08.28'
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -4,29 +4,103 @@ on: [push, pull_request]
...
+
+    - uses: Swatinem/rust-cache@v1.3.0
+
+    - name: Build kak-lsp
+      run: cargo test --no-run --locked --release
+
+    - name: Run unit tests
       run: cargo test --release
-    - name: Install system test dependencies
+
```
krobelus, line 27: I would call them system tests because they test exactly what the user runs - an actual instance of Kakoune with real language servers.
Though this really doesn't matter, no need to change it.

In future we could have integration tests that test the kak-lsp with mock LSP messages.

.github/workflows/test.yml
```diff
@@ -4,29 +4,103 @@ on: [push, pull_request]
...
+    - uses: Swatinem/rust-cache@v1.3.0
+
+    - name: Build kak-lsp
+      run: cargo build --locked --release
+
+    - name: Fetch Kakoune
+      uses: actions/checkout@v2
+      with:
+        repository: 'mawww/kakoune'
+        ref: 'v2021.08.28'
```
krobelus, line 45: Shouldn't we keep using HEAD to find issues before a Kakoune release? This could cause intermittent failures that are not kak-lsp's fault but the Kakoune master branch has been fairly stable.

.github/workflows/test.yml
```diff
@@ -4,29 +4,103 @@ on: [push, pull_request]
...
+    - name: Build Kakoune
+      if: steps.cache-kakoune.outputs.cache-hit != 'true'
+      working-directory: ${{ github.workspace }}/kakoune/src
       run: |
-        (cd && git clone https://github.com/mawww/kakoune)
-        (cd ~/kakoune && git rev-parse HEAD && make && sudo make install)
+        if [ "${{ matrix.os }}" = "macos-latest" ]; then
+          n=$(sysctl -n hw.logicalcpu)
+        else
+          n=$(nproc)
```
krobelus, line 73: this is clever although it probably doesn't do much. I think GitHub Actions allocates only 1 CPU.

.github/workflows/test.yml
```diff
@@ -4,29 +4,103 @@ on: [push, pull_request]
...
+      with:
+        key: ${{ runner.os }}-kakoune-${{ steps.kakoune-commit-id.outputs.id }}
+        path: |
+          ${{ github.workspace }}/kakoune/src/kak
+          ${{ github.workspace }}/kakoune/src/kak.opt
+          ${{ github.workspace }}/kakoune/src/.*.opt.d
+          ${{ github.workspace }}/kakoune/src/.*.opt.o
+
+    - name: Build Kakoune
+      if: steps.cache-kakoune.outputs.cache-hit != 'true'
```
krobelus, line 67:
```suggestion
      if: steps.cache-kakoune.outputs.cache-hit == 'false'
```

.github/workflows/test.yml
```diff
@@ -4,29 +4,103 @@ on: [push, pull_request]
 
 env:
   CARGO_TERM_COLOR: always
+  # CI builds are closer to a from-scratch build anyways
+  # and incremental compilation hurts reproducibility
+  CARGO_INCREMENTAL: 0
```
krobelus, line 9: I'd remove the comment; the commit message gives the reason.

Also please sort CARGO_INCREMENTAL before CARGO_TERM_COLOR

.github/workflows/test.yml
```diff
@@ -4,29 +4,103 @@ on: [push, pull_request]
...
+        if [ "${{ matrix.os }}" = "macos-latest" ]; then
+          n=$(sysctl -n hw.logicalcpu)
+        else
+          n=$(nproc)
+        fi
+
+        make -j$n
+
+    - name: Touch build targets
+      if: steps.cache-kakoune.outputs.cache-hit
```
krobelus, line 79: I think this condition is always true since the variable is just a javascript string.
It hints at why this is needed but it's also confusing. We could move this to the install step, like `make -t && sudo make install`.
However, it seems even cleaner to move this to the build step above:

```
if ${{ steps.cache-kakoune.outputs.cache-hit  }}; then
	make -t
else
	make -j$n
fi
```

krobelus (commented):
Looks nice! Last time I tried to add a cache to a Rust CI I actually made it slower, but this makes our CI faster overall.
I've left mostly cosmetic comments.

.github/workflows/test.yml
sclu1034: It's two (v)CPUs for the Ubuntu container and three for MacOS. Not sure if that's fixed and/or documented somewhere, but that's what this snippet currently evaluates to.

krobelus:
Thanks, merged as 1093749..0988032
The Kakoune cache is a great idea. If there is a PPA with nightlies (I didn't check) we could use that instead.
We should probably also parallelize the `test/run` script. Currently we have a bunch of sequential `sleep 10` in there..
I guess a downside is that we are no longer sharing the compiled kak-lsp across unit tests and system tests but that''s not so important, and can be solved in a different way

PR #534, opened 2021-09-27
Semantic token with controlFlow modifier

vbauerster:
Highlights ? operator with rust-analyzer.

### DIFF
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -344,6 +344,11 @@ token = "operator"
 face = "operator"
 
 [[semantic_tokens]]
+token = "operator"
+face = "keyword"
+modifiers = ["controlFlow"]
+
+[[semantic_tokens]]
 token = "string"
 face = "string"
 


kak-lsp.toml
```diff
@@ -343,6 +343,11 @@ face = "module"
 token = "operator"
 face = "operator"
 
+[[semantic_tokens]]
+token = "operator"
+face = "keyword"
+modifiers = ["controlFlow"]
```
krobelus, line 349: This highlights the `?` in blue with Kakoune's default theme. At least on gnome-terminal (GNOME light scheme) that's pretty similar to the default black text color. The old (yellow) was better visible for single-char operators that are surrounded by black text.

So with my knowledge I'd probably the "operator" face. Does it make sense to still add modifiers like this? Are they standardized across servers?

(I'm not sure which terminal to use as reference. Unfortunately xterm doesn't work because it looks pretty ugly with the default Kakoune theme already..)

vbauerster: > Are they standardized across servers?

Probably not, I think `controlFlow` modifier is specific to rust-analyzer.

> Does it make sense to still add modifiers like this?

Probably not again ) It wasn't obvious for me how modifiers work, so I though it would be helpful to include one more example with different modifier. I agree that this PR bears little value other than just a reference example, so it's ok to reject it if you decide so.

sclu1034: The protocol specification does include a number of tokens and modifiers: https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_semanticTokens. But they're basically just extracted from what TypeScript needs, and `controlFlow` is custom to rust-analyzer.

A different option to share examples for face mappings could be a wiki page or a discussions thread that encourages users to showcase their setup. That would also serve as a way to list what tokens/modifiers various servers support. Currently, you often need to dig around source code to find those.

PR #538, opened 2021-10-14
Show lightbulb in modeline to hint at available code actions

### LINKED ISSUE #475: Add highlighter for code actions?
It would be nice to have a highlighter (either in the margin or inline) to show when code actions are available, similar to the little lightbulbs in VSCode or Jetbrains IDEs.

krobelus:
Off by default because this wastes two columns.  Use
`:lsp-code-actions-enable` to activate.  When a lightbulb is shown,
use `:lsp-code-actions` (bound to `a` in the `lsp` user mode) to
select one to run via a menu.

Closes #475
Closes mawww/kakoune#4387

I've been testing with this file and rust-analyzer (make sure to uncomment the rust-analyzer part in the default config)

```rust
enum Test {
    Foo,
    Bar,
}
fn main() {
    let foo: Test = Test::Foo;
    match foo {
        Test::Foo => println!("foo"),
        _ => (),
    }
}
```

TODO:
- [x] How can we avoid wasting two columns?

### DIFF
--- a/test/rust-analyzer-inlay-code-actions.sh
+++ b/test/rust-analyzer-inlay-code-actions.sh
@@ -1,0 +1,46 @@
+#!/bin/sh
+
+# REQUIRES: command -v rust-analyzer
+
+. test/lib.sh
+
+cat > .config/kak-lsp/kak-lsp.toml << EOF
+[language.rust]
+filetypes = ["rust"]
+roots = ["Cargo.toml"]
+command = "rust-analyzer"
+EOF
+
+cat > main.rs << EOF
+enum Test {
+    Foo,
+    Bar,
+}
+fn main() {
+    let foo: Test = Test::Foo;
+    match foo {
+        Test::Foo => println!("foo"),
+        _ => (),
+    }
+}
+EOF
+
+cat >> .config/kak/kakrc << 'EOF'
+declare-option range-specs lsp_code_actions_hint
+add-highlighter global/lsp_code_actions_hint replace-ranges lsp_code_actions_hint
+define-command -override -hidden lsp-show-code-actions -params 1.. %{
+	evaluate-commands -draft %{
+		execute-keys <a-l><semicolon>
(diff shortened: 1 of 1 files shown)


Gskartwii:
> +1
> 
> Emacs shows this in modeline: [emacs-lsp/lsp-mode#1781](https://github.com/emacs-lsp/lsp-mode/pull/1781). It also shows the label of the fist action, which again is nice for discoverability (but might be considered too noisy).

I like this idea. It's also the reason why I put my line number view in the status line, instead of spending columns on it. 

@krobelus: Could kak-lsp expose information about the code actions in an `opt`, so that it can be used in the statusline, even when the side column view is off?

loovjo:
Wow, this looks great! I think the light bulb should be customizable, having it inside an option that the user could override would be great, so the user doesn't have to override the internal `:lsp-show-code-actions`-command.

sclu1034:
> This looks a bit dorky, is there something that looks better?

This depends on the font used. Twemoji's flat design tends to fit terminals better than Noto.
Alternatively, you could use the text mode variant selector:

```sh
echo -e "💡\ufe0e"
💡︎
```

dpc:
Did a `:plug-update` and it works here. Awesome! Thanks a lot!

Screenshot from 2021-10-19 13-19-05

matklad:
I am wondering about”unsupported progress” logs. We only emit progress notifications when the client actively advertises that it wants progress: 

https://github.com/rust-analyzer/rust-analyzer/blob/5051717856dd325027e003c6b64497a2b113ba07/crates/rust-analyzer/src/config.rs#L574

https://github.com/rust-analyzer/rust-analyzer/blob/5051717856dd325027e003c6b64497a2b113ba07/crates/rust-analyzer/src/lsp_utils.rs#L76

So it seems as if kak-lsp asks for progress, and than complaints that it can’t handle it?

PR #551, opened 2021-11-01
Support filterText in completions

### LINKED ISSUE #444: Julia unicodes
The LanguageServer of Julia has unicode completion support e.g. `\xi` becomes `ξ` (full list here https://docs.julialang.org/en/v1/manual/unicode-input/index.html). 

`kak-lsp` does know about this, but it can't handle the completion/substitution, see:
grafik

`\xi` is recognised but becomes:
grafik

krobelus:
This currently requires Kakoune built with https://github.com/mawww/kakoune/pull/4418.

Closes #444

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -208,6 +208,13 @@ define-command -hidden lsp-completion -docstring "Request completions for the ma
     lsp-did-change-and-then lsp-completion-request
 }
 
+declare-option -hidden bool lsp_have_kakoune_feature_filtertext
+declare-option -hidden completions lsp_have_kakoune_feature_filtertext_tmp
+try %{
+    set-option global lsp_have_kakoune_feature_filtertext_tmp 1.1@0 insert_text|filter_text|on_select|menu
+    set-option global lsp_have_kakoune_feature_filtertext true
+}
+
 define-command -hidden lsp-completion-request -docstring "Request completions for the main cursor position" %{
 try %{
     # Fail if preceding character is a whitespace (by default; the trigger could be customized).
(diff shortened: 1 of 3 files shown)


krobelus:
To test, compile Kakoune with the prerequisite https://github.com/mawww/kakoune/pull/4418.
So far I've done some light testing with ccls and Julia's language server.
This is relevant for servers that send `textEdit` in their completion responses, e.g. the log contains something like
`"textEdit":{"range":{"start":{"line":4,"character":2},"end":{"line":4,"character":4}},"newText":"foobar"}`

tadeokondrak:
From the Kakoune commit:

> Additionally, these completions will not be filtered or ranked, since there is no text to filter by.

Not having filtering or ranking makes autocompletion seems to make completion less useful.

Maybe completions could be `<text>|<select cmd>|<menu text>|<new option>` where the new option toggles whether Kakoune should insert the text - and the first parameter is still used for filtering. It would also make it possible to support LSP's `CompletionItem.filterText`.

krobelus:
You're right I forgot about that. I'll try to update soon with filtering. The missing filtering is really bad.
(I actually didn't realize that most (?) language servers these days send textEdits - at least `rust-analyzer` does.)

I just wasn't sure how to keep filtering working with the Unicode completions from Julia LS (which sends the thing below).
In hindsight, it seems trivial: we just need to use `label` as fallback for `CompletionItem.filterText`. Not sure how I missed that.

```json
      {
        "label": "\\xi",
        "kind": 11,
        "detail": "ξ",
        "documentation": "ξ",
        "textEdit": {
          "range": {
            "start": {
              "line": 6,
              "character": 0
            },
            "end": {
              "line": 6,
              "character": 2
            }
          },
          "newText": "ξ"
        }
      },
```

PR #555, opened 2021-11-04
add config for R language server

jcai849:
Using the standard https://github.com/REditorSupport/languageserver/
Tested, working as expected

### DIFF
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -234,6 +234,12 @@ settings_section = "_"
 # and https://github.com/palantir/python-language-server/blob/develop/vscode-client/package.json
 # pyls.configurationSources = ["flake8"]
 
+[language.r]
+filetypes = ["r"]
+roots = [".git", ".hg", "DESCRIPTION"]
+command = "R"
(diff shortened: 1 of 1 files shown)


kak-lsp.toml
```diff
@@ -234,6 +234,12 @@ settings_section = "_"
 # and https://github.com/palantir/python-language-server/blob/develop/vscode-client/package.json
 # pyls.configurationSources = ["flake8"]
 
+[language.r]
+filetypes = ["r"]
+roots = [".git", ".hg", "DESCRIPTION"]
```
krobelus, line 239: I guess we want DESCRIPTION first, otherwise if any parent directory has a `.git` folder, that directory will be used as root. Both orderings can go wrong of course.

We should probably change the root detection to check each glob at a directory before deferring to the parent directory.
Though the potential breakage may not be worth it.

jcai849: Yes, that's a good point. A breadth-first ordered search for globs may be better, but you're right - the breakage will be too significant. I have commited a change reflecting the recommendation.

PR #557, opened 2021-11-06
Add the ability to goto/get hover info of the next/prev symbols in a document

sidkshatriya:
This PR adds functionality to deal with the following Scenarios:

## Scenario 

You are deep within a huge function. You have forgotten the name of the function you are in or want to know what its parameters are. This PR provides the kakoune command `lsp-hover-previous-function`. After executing the command you get a modal with the hover info of the previous function _without your cursor actually going to the previous function_. 

If you press `g` after the modal appears you can actually goto to the function start. If you press any other key, the modal is dismissed and you continue editing like before. 

For completeness there is also `lsp-hover-next-function`, `lsp-hover-next-method`, `lsp-hover-previous-method`.

**More generically we have `lsp-hover-previous-symbol <symbol-kind>` and `lsp-hover-next-symbol <symbol-kind>`** . Example: `lsp-hover-next-symbol Constructor`, `lsp-hover-previous-symbol Variable` etc. 

If no `<symbol kind>` is provided, kakoune shows the hover info of the previous/next symbol of _any_ kind.

## Scenario

You are done editing the current function. You simply want to goto the next function start without having to hunt for it. Use `lsp-goto-next-function` and your cursor will goto the next function. This command is also useful for those situations when you just want to cycle through all the functions in a document.

For completeness there is `lsp-goto-previous-function`, `lsp-goto-next-method`, `lsp-goto-previous-method`.

**More generically we have `lsp-goto-previous-symbol <symbol-kind>` and `lsp-goto-next-symbol <symbol-kind>`**. Example: `lsp-goto-next-symbol Constructor`, `lsp-goto-previous-symbol Variable` etc.

If no `<symbol kind>` is provided, kakoune will goto the previous/next symbol of _any_ kind.

The following new shortcuts are proposed
```
map global lsp <[> '<esc>: lsp-hover-previous-symbol<ret>'   -docstring 'show hover for previous symbol'
map global lsp <]> '<esc>: lsp-hover-next-symbol<ret>'       -docstring 'show hover for next symbol'
map global lsp <{> '<esc>: lsp-goto-previous-symbol<ret>'    -docstring 'goto previous symbol'
map global lsp <}> '<esc>: lsp-goto-next-symbol<ret>'        -docstring 'goto next symbol'
map global lsp <9> '<esc>: lsp-hover-previous-function<ret>' -docstring 'show hover for previous function'
map global lsp <0> '<esc>: lsp-hover-next-function<ret>'     -docstring 'show hover for next function'
map global lsp <(> '<esc>: lsp-goto-previous-function<ret>'  -docstring 'goto previous function'
map global lsp <)> '<esc>: lsp-goto-next-function<ret>'      -docstring 'goto next function'
```


sidkshatriya:
**Comment on Testing**
I have tested this mainly with rust-analyzer successfully. 

I also was able to test it (less intensively though) but successfully on: gopls, elixir-ls, erlang-ls, ocamllsp, haskell language server, clangd and pyls

sidkshatriya:
P.S. Please test on your machine(s) and please let me know if the feature is working well for you!

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+    } else {
+        ctx.exec(
+            meta,
+            "eval %[
+                info -style modal 'Not found!\n\nPress any key to continue'
+                on-key %[info -style modal]
+            ]",
+        );
+    }
+}
```
krobelus, line 163: This is really amazing. I've briefly used VSCode, and I liked that it shows the name of the current function at the top of each buffer. So it's not a problem if you forget the name of the current function.

So this was a feature I've definitely been missing, and I like the hover UI a lot. I'll try it some more this week, to see if it stands the test of time. In general I don't use hover much, but this seems very useful.

I'll need some time to fully review this. I have already started writing lots of review comments.
If you don't have time to address them, let me know, then I can keep it to the important bits / take over.

I'm comfortable writing GitHub review comments, because I write them in Kakoune and post them using the GitHub API, but I understand that dealing with lots of comments may be difficult when using the GitHub web interface (Sadly I can't reply to threads using the API just yet). I'm happy to switch to something like email if people prefer that.

rc/lsp.kak
```diff
@@ -1388,8 +1534,16 @@ map global lsp n '<esc>: lsp-find-error<ret>'             -docstring 'find next
...
-map global lsp <&> '<esc>: lsp-highlight-references<ret>' -docstring 'lsp-highlight-references'
-map global lsp = '<esc>: lsp-range-formatting<ret>'       -docstring 'format selections'
+map global lsp <[> '<esc>: lsp-hover-previous-symbol<ret>'   -docstring 'show hover for previous symbol'
+map global lsp <]> '<esc>: lsp-hover-next-symbol<ret>'       -docstring 'show hover for next symbol'
+map global lsp <{> '<esc>: lsp-goto-previous-symbol<ret>'    -docstring 'goto previous symbol'
+map global lsp <}> '<esc>: lsp-goto-next-symbol<ret>'        -docstring 'goto next symbol'
+map global lsp <9> '<esc>: lsp-hover-previous-function<ret>' -docstring 'show hover for previous function'
+map global lsp <0> '<esc>: lsp-hover-next-function<ret>'     -docstring 'show hover for next function'
+map global lsp <(> '<esc>: lsp-goto-previous-function<ret>'  -docstring 'goto previous function'
+map global lsp <)> '<esc>: lsp-goto-next-function<ret>'      -docstring 'goto next function'
```
krobelus, line 1544: The mappings make sense. Nice.

rc/lsp.kak
```diff
@@ -228,18 +228,157 @@ define-command lsp-hover -docstring "Request hover info for the main cursor posi
...
+[params]
+position.line = %d
+position.column = %d
+style = "%s"
+context = "%s"
+' "${kak_session}" "${kak_client}" "${kak_buffile}" "${kak_opt_filetype}" "${kak_timestamp}" ${kak_cursor_line} ${kak_cursor_column} "$1" "$2" | eval ${kak_opt_lsp_cmd} --request) > /dev/null 2>&1 < /dev/null & }
+}
+
+declare-option -hidden str lsp_symbol_kind_completion %{
+  symbol_kinds="
```
krobelus, line 248: we use four space indentation here.
I have an editorconfig file lying around, I'll add it soon.

rc/lsp.kak
```diff
@@ -228,18 +228,157 @@ define-command lsp-hover -docstring "Request hover info for the main cursor posi
...
+  File Module Namespace Package Class Method Property Field Constructor Enum Interface Function
+  Variable Constant String Number Boolean Array Object Key Null EnumMember Struct Event Operator
+  TypeParameter
+  "
+  for symbol_kind in ${symbol_kinds}; do
+    printf '%s\n' "${symbol_kind}"
```
krobelus, line 254: The loop is actually not necessary - `printf '%s\n' ${symbol_kinds}` is enough

rc/lsp.kak
```diff
@@ -228,18 +228,157 @@ define-command lsp-hover -docstring "Request hover info for the main cursor posi
...
+  done
+}
+
+define-command lsp-goto-previous-symbol -params 0..1 -shell-script-candidates %opt{lsp_symbol_kind_completion} \
+    -docstring "lsp-goto-previous-symbol [<symbol-kind>]: Goto to previous document symbol of type <symbol-kind>. If <symbol-kind> is not given, it means _any_ symbol" %{
```
krobelus, line 259:
```suggestion
    -docstring "lsp-goto-previous-symbol [<symbol-kind>]: Goto the previous document symbol of type <symbol-kind>. If <symbol-kind> is not given, it means _any_ symbol" %{
```

same below

src/language_features/hover.rs
```diff
@@ -109,17 +111,36 @@ pub fn editor_hover(
...
+        anchor_or_style,
         contents.replace("§", "§§"),
         diagnostics.replace("§", "§§")
     );
 
+    // Wrap if we're using a HoverModal
+    if let Some(hover_modal) = maybe_hover_modal {
+        command = format!(
+            "eval %§
+                 {}
```
krobelus, line 136: The extra indentation in continuation lines will look weird in the log, so I think it's better to use

```rust
            "eval %§\n    {}\n    {}\n§",
```

not sure if we use this style of indentation elsewhere but why not I guess.

Same thing happens a few more times in this patch. I wonder if it's worth looking into a general solution. Sadly I haven't found a conventional way to do this in Rust.

Java, out of all languages gets multiline string literals right - unlike most other languages they automatically remove common indentation in strings like

```java
	String x = """
		123
			456
		789
	""";
```

This will evaluate to the same as `"123\n\t456\789\n"` (not sure about leading/trailing whitespace, but that's not the point).
Maybe we should add a function that does that, should be very simple.

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
+use crate::{
```
krobelus, line 1: not sure about creating a new file.

We have a couple of other home-grown requests like `stop`, `apply-workspace-edit`, `apply-text-edits` which should be called `kak-lsp/stop` etc. (but I guess it's not worth to change those).

The new filename implies that `kak-lsp/next-or-previous-symbol` is somehow special but, as you state, the implementation is very similar to the kak-lsp request `textDocument/documentSymbol`.
Both start with requesting the LSP request `textDocument/documentSymbol` from the language server, and the do their custom, kak-lsp-specific postprocessing.

This confusion is because the existing kak-lsp request `textDocument/documentSymbol` borrows the name from the corresponding LSP request.
This is probably still a reasonable thing, because it's an obvious name (makes logs easy to read) and keeps us close to the protocol.

So I think we should move these additions to `src/language_features/document_symbol.rs`, which is a much better name because it's less generic. This also makes it easier to share some of the logic (for example we could extract a function to share between `text_document_document_symbol` and `document_next_or_prev_symbol_request`).

Maybe we should also rename `kak-lsp/next-or-previous-symbol` into something that contains `textDocument/documentSymbol`,
so it's obvious what the underlying LSP request is.
`kak-lsp/next-or-previous/textDocument/documentSymbol`? Reads a bit awkward

src/controller.rs
```diff
@@ -259,6 +259,9 @@ fn dispatch_editor_request(request: EditorRequest, mut ctx: &mut Context) {
         request::DocumentSymbolRequest::METHOD => {
             document_symbol::text_document_document_symbol(meta, &mut ctx);
         }
+        "kak-lsp/next-or-previous-symbol" => {
+            kak_lsp_custom::document_next_or_prev_symbol_request(meta, params, &mut ctx);
```
krobelus, line 263: I believe that the names of these commands should correspond to the method name 1:1 (I didn't check if we have exceptions).
For example, `text_document_document_symbol` corresponds to `textDocument/documentSymbol`.

The consistent name here would be `kak_lsp_next_or_previous_symbol`. Maybe we should make exceptions for kak_lsp-prefixed ones..

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+    editor_params: EditorParams,
+    result: Option<DocumentSymbolResponse>,
+    ctx: &mut Context,
+) {
+    let params = NextOrPrevSymbolParams::deserialize(editor_params).unwrap();
+    let hover = params.hover;
+    let maybe_details = match result {
+        Some(DocumentSymbolResponse::Flat(mut result)) => {
+            if result.is_empty() {
+                return;
```
krobelus, line 56: these are redundant but why not I guess

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+            }
+            // First let's sort the results so we can find next and previous properly
+            // This step does _not_ happen in `document_symbol::editor_document_symbol()`.
+            //
+            // Some language servers return symbol locations in unsorted order
+            // or non-ascending order.
+            result.sort_by(|a, b| a.location.range.start.cmp(&b.location.range.start));
```
krobelus, line 63: The docs say

> When applicable, unstable sorting is preferred because it is generally faster

do we expect symbols that start at the same location? Seems hard to find out

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+/// a) Instructions to move the cursor to the next/previous symbol.
+/// b) Instructions to show hover information (without actually moving the
+///    visible user cursor) _of_ the next/previous symbol
+fn next_or_prev_response(
+    meta: EditorMeta,
+    ctx: &mut Context,
+    maybe_details: Option<(String, KakounePosition, String, SymbolKind)>,
+    hover: bool,
+) {
+    if let Some((filename, symbol_position, name, kind)) = maybe_details {
```
krobelus, line 96: I'd use an early return here, to reduce indentation and make it easier to read, especially for error handling.
I usually use this pattern

```rust
    let (filename, symbol_position, name, kind) = match maybe_details {
        Some((filename, symbol_position, name, kind)) => (filename, symbol_position, name, kind),
        None => {
            ctx.exec(...);
            return;
        }
    }
```

Same for the `if hover` below - IMO the easy case should come first, like `if !hover { ...; return; }`.

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+                        },
+                        result,
+                        ctx,
+                    )
+                },
+            );
+        } else {
+            let location = format!(
+                "%§{}§ {} {}",
+                filename.replace("§", "§§"),
```
krobelus, line 147: we usually use `editor_quote()` for simple case like this.
I think the intention behind using `§` is that we don't get so many nested single quotes in the logs.

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+                        uri: Url::from_file_path(&meta.buffile).unwrap(),
+                    },
+                    position: get_lsp_position(&meta.buffile, &symbol_position, ctx).unwrap(),
+                },
+                work_done_progress_params: Default::default(),
+            };
+
+            // This context is shown at the top of the modal
+            let context = format!(
+                                "line {}:{}:{{+b@KindAndName}}{:?} {}{{KindAndName}} (Press `g` to goto this position. Press any other key to continue)",
```
krobelus, line 111: The docstring of Kakoune's `add-highlighter` uses single quotes, which fits better than backticks IMO. Or maybe no quotes at all?

```suggestion
                                "line {}:{}:{{+b@KindAndName}}{:?} {}{{KindAndName}} (Press 'g' to goto this position. Press any other key to continue)",
```

rc/lsp.kak
```diff
@@ -228,18 +228,157 @@ define-command lsp-hover -docstring "Request hover info for the main cursor posi
...
+[params]
+position.line      = %d
+position.column    = %d
+symbolKind = "%s"
+searchNext = true
+' "${kak_session}" "${kak_client}" "${kak_buffile}" "${kak_opt_filetype}" "${kak_timestamp}" ${kak_cursor_line} ${kak_cursor_column} "$1" | eval ${kak_opt_lsp_cmd} --request) > /dev/null 2>&1 < /dev/null & }
+}
+
+# Frequently used, provided for convenience
+define-command lsp-goto-next-function -params 0 \
```
krobelus, line 303: I'm not sure if it's worth to provide these convenience methods.
Without it, users will find the more powerful `lsp-goto-next-symbol` faster, and for bindings it doesn't matter.

krobelus, line 303: I don't think we use the redundant `-params 0` bit anywhere.

rc/lsp.kak
```diff
@@ -999,10 +1138,17 @@ define-command -hidden lsp-show-hover -params 3 -docstring %{
 
     content=$(printf %s "$content" | sed s/\'/\'\'/g)
 
-    case $kak_opt_lsp_hover_anchor in
-        true) printf "info -markup -anchor %%arg{1} -- '%s'" "$content";;
-        *)    printf "info -markup -- '%s'" "$content";;
+    case $1 in
+        modal) printf "info -markup -style modal -- '%s'" "$content";;
+        above) printf "info -markup -style above -- '%s'" "$content";;
```
krobelus, line 1143: only the "modal" style is currently used, so I'd remove the other cases.
(also "below" doesn't seem to work).

If we allow them we'd write it as
`above|below|menu|modal) printf "info -markup -style $1 -- '%s'" "$content";;`

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+                                exec {}g{}lh
+                                ';
+                     else echo 'info -style modal';
+                     fi
+                 ]]",
+                symbol_position.line, symbol_position.column
+            );
+
+            // Make a HoverRequest to the backend language server and then call `hover::editor_hover()`
+            ctx.call::<HoverRequest, _>(
```
krobelus, line 129: In general, I'm not a big fan of having literal function names in comments, because it makes `:grep` and `:lsp-rename` worse. I'd rather describe in prose what should happen.

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+                },
+            );
+        } else {
+            let location = format!(
+                "%§{}§ {} {}",
+                filename.replace("§", "§§"),
+                symbol_position.line,
+                symbol_position.column
+            );
+            let command = format!("edit! -existing {}", location);
```
krobelus, line 151: for the unlikely case a filename starts with `-`

```suggestion
            let command = format!("edit! -existing -- {}", location);
```

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+                                "line {}:{}:{{+b@KindAndName}}{:?} {}{{KindAndName}} (Press `g` to goto this position. Press any other key to continue)",
+                                symbol_position.line, symbol_position.column, kind, escape_brace(&name)
+                            );
+
+            // This kak script is appended after the kakoune `lsp-show-hover`
+            let do_after = format!(
+                "on-key %[eval %sh[
+                     if [ \"$kak_key\" = \"g\" ];
+                     then echo 'info -style modal
+                                exec {}g{}lh
```
krobelus, line 120: One tiny thing that feels a bit weird is that we'll jump to the last character of target identifiers.
Maybe it's more intuitve to end up at the last-plus-one character, or at the first character? I'm not sure.
Though I guess you added the `h` for a reason, so I'm probably missing something here.

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+            let context = format!(
+                                "line {}:{}:{{+b@KindAndName}}{:?} {}{{KindAndName}} (Press `g` to goto this position. Press any other key to continue)",
+                                symbol_position.line, symbol_position.column, kind, escape_brace(&name)
+                            );
+
+            // This kak script is appended after the kakoune `lsp-show-hover`
+            let do_after = format!(
+                "on-key %[eval %sh[
+                     if [ \"$kak_key\" = \"g\" ];
+                     then echo 'info -style modal
```
krobelus, line 119: I guess it's better to extract out the hiding of the modal menu

```kak
on-key %[
    info -style modal
    evaluate-commands %sh[
        if [ \"$kak_key\" = \"g\" ]; then echo exec {}g{}lh; fi
    ]
]
```

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+            meta,
+            "eval %[
+                info -style modal 'Not found!\n\nPress any key to continue'
+                on-key %[info -style modal]
+            ]",
+        );
+    }
+}
+
+/// Does the currently being inspected symbol's location "exceed" the
```
krobelus, line 165:
```suggestion
/// Does the current symbol's location "exceed" the
```

(or "symbol at cursor")

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+    expected_kind: &str,
+) -> bool {
+    // expected kind is the symbol for which the user is searching for
+    // e.g. `Function`, `Constructor` etc.
+    if !expected_kind.is_empty() && format!("{:?}", symbol_kind) != expected_kind {
+        return false;
+    }
+
+    // If searching forwards, the first element that has a greater line/column
+    // combination
```
krobelus, line 181: please reflow to 100 columns (which should be the rustfmt default). Again, I'll add the editorconfig soon

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
...
+    None
+}
+
+fn get_symbol_hover_pos(
+    filename: &String,
+    ctx: &Context,
+    location: &Location,
+    name: &String,
+) -> KakounePosition {
+    // We need to handle some language servers specially
```
krobelus, line 259: If I open our `src/controller.rs` using rust-analyzer, scroll to the bottom and
run `lsp-goto-previous-function`, then the cursor ends up inside the docstring,
in a line that reads

	/// by the incoming request. `ensure_did_open` tries to sneak in `textDocument/didOpen` request for

This is because we use `find_name_in_file` to "correct" the position that the language server gives us - the start of the doc comment. We want to the "meat", the "fn ensure_did_open". 
So `find_name_in_file` is a workaround for missing information in LSP. I guess it's fine as long as it doesn't make realistic cases worse. We can probably ignore the case of `ensure_did_open` since it's rare? I'm not sure, I usually work with more concise  doc comments, if any.

So I guess `location.range.start` is not good enough. I guess, because of doc comments, as mentioned above.
Does `selection_range` work? Ideally we could drop language-specific workarounds.

rc/lsp.kak
```diff
@@ -228,18 +228,157 @@ define-command lsp-hover -docstring "Request hover info for the main cursor posi
...
+  File Module Namespace Package Class Method Property Field Constructor Enum Interface Function
+  Variable Constant String Number Boolean Array Object Key Null EnumMember Struct Event Operator
+  TypeParameter
+  "
+  for symbol_kind in ${symbol_kinds}; do
+    printf '%s\n' "${symbol_kind}"
+  done
+}
+
+define-command lsp-goto-previous-symbol -params 0..1 -shell-script-candidates %opt{lsp_symbol_kind_completion} \
```
krobelus, line 258: Maybe we should drop the "goto" bit from these, to be consistent with `lsp-next-location`. I deprecated `lsp-goto-next-match` because it's not as flexible.

rc/lsp.kak
```diff
@@ -228,18 +228,157 @@ define-command lsp-hover -docstring "Request hover info for the main cursor posi
...
+
+define-command -hidden lsp-hover-previous-symbol-request -params 0..1 -shell-script-candidates %opt{lsp_symbol_kind_completion} \
+    -docstring "lsp-hover-previous-symbol-request [<symbol-kind>]: Show hover of the previous symbol of type <symbol-kind> in the document. If <symbol-kind> is not given, it means _any_ symbol. The editor does _not_ actually navigate to that symbol. This command is useful to peek at the symbol hover information" %{
+    nop %sh{ (printf '
+session   = "%s"
+client    = "%s"
+buffile   = "%s"
+filetype  = "%s"
+version   = %d
+method    = "kak-lsp/next-or-previous-symbol"
```
krobelus, line 339: We should probably just have one hidden `lsp-next-or-previous-symbol <next> <hover> <symbol>`.
We could even make autocompletion work with `-shell-script-completion` but since it's a hidden command it doesn't matter.

src/util.rs
```diff
@@ -228,3 +230,74 @@ pub fn short_file_path<'a>(target: &'a str, current_dir: &str) -> &'a str {
...
+    if name.is_empty() {
+        return Some(start_from);
+    }
+    let contents = get_file_contents(filename, ctx).unwrap();
+    let line_offset: usize = contents.line_to_char(start_from.line.try_into().unwrap());
+    let within_line_offset: usize = start_from.character.try_into().unwrap();
+    let char_offset = line_offset + within_line_offset;
+    let mut it = name.chars();
+    let mut maybe_found = None;
+    for (num, c) in contents.chars_at(char_offset).enumerate() {
```
krobelus, line 257: I don't understand everything here, but this looks like it could be done with a simple regex search.
`treat_name_as_symbol` is effectively unused right now (since it's always true), so we should remove it.
With that, we can use a `\b` word boundary instead of checking if the next character is alphanumeric,
that will protect against `f_totally_not_f`.

src/util.rs
```diff
@@ -228,3 +230,74 @@ pub fn short_file_path<'a>(target: &'a str, current_dir: &str) -> &'a str {
         .and_then(|p| p.to_str())
         .unwrap_or(target)
 }
+
+/// Given a starting Location, try to find `name` in a file and report its Position
```
krobelus, line 234: I don't think we should add more to `src/util.rs`, because it's such an opaque name. Rather add these where needed and *maybe* extract them when we want to reuse them.

src/language_features/kak_lsp_custom.rs
sidkshatriya: @krobelus Feel free to add as many review comments as you need to! 

I'll try to address as many as I can this week. For any significant ones that remain, we can discuss what to do about them -- thanks for your offer of taking them up, if necessary.

sidkshatriya:
@krobelus 
I haven't forgotten about this. I will try to get to this soon -- really. I would like to see this get into master.

Have you used this feature BTW this week? Has it been reliable?

krobelus:
@sidkshatriya sure sounds good.

Actually the only *interesting* question that came up in my review is whether we can use `selection_range` over `range` to avoid the language-specific workarounds. The rest are mainly style nitpicks, we can fix the easy ones and ignore others.

(also we should update the documentation once we have finalized the commands/keybindings, but that can be done later)

> Have you used this feature BTW this week? Has it been reliable? 

not as much as much as I'd like, I guess mostly because I still need to change habits: I was using `m` to skip function bodies, but this is more powerful (because I don't need to awkwardly navigate to the `{` first). I haven't found any problems so far. I wanted to do a quick internal demo but didn't get around to, I'll try to do that next week ;)

src/language_features/hover.rs
krobelus: Looks like the [indoc!](https://crates.io/crates/indoc) macro would solve this. Any concerns with adding that dependency?

sidkshatriya:
> selection_range over range to avoid the language-specific workarounds.

`selection_range` would be great.

But most LSPs invoked by `kak-lsp` seem to return `struct SymbolInformation` on the `textDocument/documentSymbol` request only. 

Now `struct SymbolInformation` in kak-lsp only has `range`, it does not have `selection_range` (unlike `struct DocumentSymbol`). This is what the reference [docs](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#symbolInformation) indicate also.

How can we start using `selection_range` ? Is there some setting on the kak-lsp side that will start allowing LSPs to send us their results as `DocumentSymbol`?

src/language_features/hover.rs
krobelus: Done with 0cc20e9 (Use indoc for multiline string literals) - (sorry for the force-push), so this can be

```diff
diff --git a/src/language_features/hover.rs b/src/language_features/hover.rs
index 149f9b6..3ccf100 100644
--- a/src/language_features/hover.rs
+++ b/src/language_features/hover.rs
@@ -2,6 +2,7 @@ use crate::context::*;
 use crate::markup::*;
 use crate::types::*;
 use crate::util::*;
+use indoc::formatdoc;
 use itertools::Itertools;
 use lsp_types::request::*;
 use lsp_types::*;
@@ -131,8 +132,8 @@ pub fn editor_hover(
 
     // Wrap if we're using a HoverModal
     if let Some(hover_modal) = maybe_hover_modal {
-        command = format!(
-            "eval %§
+        command = formatdoc!(
+            "evaluate-commands %§
                  {}
                  {}
              §",
diff --git a/src/language_features/kak_lsp_custom.rs b/src/language_features/kak_lsp_custom.rs
index dc27f59..df77ba8 100644
--- a/src/language_features/kak_lsp_custom.rs
+++ b/src/language_features/kak_lsp_custom.rs
@@ -1,6 +1,7 @@
 use crate::{
     context::*, language_features::hover::editor_hover, markup::escape_brace, types::*, util::*,
 };
+use indoc::{formatdoc, indoc};
 use lsp_types::{request::*, *};
 use serde::Deserialize;
 use url::Url;
@@ -113,16 +114,17 @@ fn next_or_prev_response(
                             );
 
             // This kak script is appended after the kakoune `lsp-show-hover`
-            let do_after = format!(
-                "on-key %[eval %sh[
-                     if [ \"$kak_key\" = \"g\" ];
-                     then echo 'info -style modal
-                                exec {}g{}lh
-                                ';
-                     else echo 'info -style modal';
-                     fi
-                 ]]",
-                symbol_position.line, symbol_position.column
+            let do_after = formatdoc!(
+                "on-key %[
+                     info -style modal
+                     evaluate-commands %sh[
+                         if [ \"$kak_key\" = \"g\" ]; then
+                             echo 'exec {}g{}lh'
+                         fi
+                     ]
+                 ]",
+                symbol_position.line,
+                symbol_position.column
             );
 
             // Make a HoverRequest to the backend language server and then call `hover::editor_hover()`
@@ -154,10 +156,14 @@ fn next_or_prev_response(
     } else {
         ctx.exec(
             meta,
-            "eval %[
-                info -style modal 'Not found!\n\nPress any key to continue'
-                on-key %[info -style modal]
-            ]",
+            indoc!(
+                "evaluate-commands %[
+                     info -style modal 'Not found!
+
+                     Press any key to continue'
+                     on-key %[info -style modal]
+                 ]"
+            ),
         );
     }
 }
```

src/language_features/kak_lsp_custom.rs
krobelus: > But most LSPs invoked by `kak-lsp` seem to return `struct SymbolInformation` on the `textDocument/documentSymbol` request only. 
		
thanks for this information, makes sense

I guess the workarounds are okay because they only apply to servers that send SymbolInformation, and failures should be rare.
It might be nice to allow users to add their own custom position-computation logic in kakscript, but that can be done later.

> How can we start using `selection_range` ? Is there some setting on the kak-lsp side that will start allowing LSPs to send us their results as `DocumentSymbol`?

right, I'll try to check now. I mostly care about gopls, clangd and rls/rust-analyzer, I'll check those

Reading gopls sources got me there quickly

```go
	// TODO: Remove this once the lsp deprecates SymbolInformation.
	symbols := make([]interface{}, len(docSymbols))
	for i, s := range docSymbols {
		if snapshot.View().Options().HierarchicalDocumentSymbolSupport {
			symbols[i] = s
			continue
		}
		// If the client does not support hierarchical document symbols, then
		// we need to be backwards compatible for now and return SymbolInformation.
		symbols[i] = protocol.SymbolInformation{...}
```

so it's just

```
diff --git a/src/general.rs b/src/general.rs
index fbd26eb..f5a742c 100644
--- a/src/general.rs
+++ b/src/general.rs
@@ -163,7 +163,7 @@ pub fn initialize(root_path: &str, meta: EditorMeta, ctx: &mut Context) {
                 document_symbol: Some(DocumentSymbolClientCapabilities {
                     dynamic_registration: Some(false),
                     symbol_kind: None,
-                    hierarchical_document_symbol_support: None,
+                    hierarchical_document_symbol_support: Some(true),
                     tag_support: None,
                 }),
                 formatting: Some(DynamicRegistrationClientCapabilities {
```

then it magically works. Very nice that we finally notice this, now we can also tweak the standard document `lsp-document-symbol` to maybe use the selection_range.

src/language_features/kak_lsp_custom.rs
```diff
@@ -0,0 +1,318 @@
+use crate::{
+    context::*, language_features::hover::editor_hover, markup::escape_brace, types::*, util::*,
```
krobelus, line 2: probably better to stick to the one-line-per `use` statement when importing our own modules,
so they are easier to grep. I'm still trying to figure when to use start imports

src/language_features/kak_lsp_custom.rs
sidkshatriya: The requirement to sort was originally triggered by the `ocamllsp` which was returning its results for document symbol in descending line order instead of ascending order which caused the next / previous logic to get messed up.

So then I just implemented the easiest way to sort things. I didn't think about stable / unstable sorting at that point. 

In general, there are quite a few inefficiencies in the kakoune <-> kak-lsp pipeline with shells being invoked from kakoune quite indiscriminately. In my view we should not worry about stable / unstable sort distinction as everything is in-memory and generally about ~100-200 items. 

You are free to mess around with this in the future, of course, to how you see fit. Lets leave as is for now?

rc/lsp.kak
sidkshatriya: Done

rc/lsp.kak
sidkshatriya: Removed

rc/lsp.kak
sidkshatriya: Removed other cases

src/util.rs
sidkshatriya: I don't know how we can do a regex search on Ropey without rendering the whole document to a `String` (which I've tried to avoid...)

src/language_features/kak_lsp_custom.rs
sidkshatriya: The `if result.is_empty()` check is consistent with code in `document_symbol::editor_document_symbol()`

Though I don't think it is redundant because an external LSP is returning back our results and we should generally be defensive about the kind of data it returns.

src/language_features/kak_lsp_custom.rs
sidkshatriya: Moved the code into `document_symbol.rs`.

I've left the function names the same for now. When the dust settles you can rename the function names to what you prefer

src/controller.rs
sidkshatriya: I leave the renaming of the functions to what you feel appropriate. Kept the name the same for now...

src/language_features/kak_lsp_custom.rs
sidkshatriya: Done

src/util.rs
sidkshatriya: Done. Moved into `document_symbol.rs` (which also now contains everything from `kak_lsp_custom.rs`

rc/lsp.kak
sidkshatriya: Done. Dropped `goto`.

src/language_features/kak_lsp_custom.rs
sidkshatriya: The way the imports are formatted should not be human edited. I think rustfmt has some common options and we should let rustfmt handle import styles.

See https://rust-lang.github.io/rustfmt

e.g. `imports_layout`, `imports_indent` etc.

sidkshatriya:
@krobelus I've dealt with a significant chunk of your review comments. Thanks for the time and energy you spent on it!

I will try to deal with the rest of the review items in the next couple of days. 

In the meanwhile if you get the time:
- There are some review comments that I've marked as resolved. You can un-resolve them in case you have any further comments
- There are some review comments that I've responded back on which I've asked a question/expressed a different point of view. We can discuss them further or you can mark them as resolved.

src/language_features/kak_lsp_custom.rs
sidkshatriya: P.S. I guess you are just suggesting we use `sort_unstable_by()` right? Its a trivial change, however, I don't have strong views on this. If you feel you want to use the unstable version, you can definitely change in future.

src/language_features/kak_lsp_custom.rs
sclu1034: The only _stable_ Rustfmt option in that regard is `reorder_imports = "yes"/"no"`. Granularity/grouping is still left untouched and has to be done manually (or semi-manually with something like rust-analyzer's (un)merge imports actions).

src/util.rs
krobelus: Ok that makes sense (I hadn't checked in detail).
Actually, since we now ask for DocumentSymbol, do we still need the workarounds for SymbolInformation?
Is there any relevant server that still only returns the deprecated SymbolInformation?
If not, then I'd just go with directly returning what the server gives us, to make everything more predictable.

src/language_features/kak_lsp_custom.rs
krobelus: <details>

It's not only performance, but about adding extra assumptions for the reader.
If the code follows the simple rule "use unstable sort unless stable sort is required" then a `.sort_unstable()` communicates that we don't get equal elements (or we don't care about their order).
For the same reason I'd rather use `x.unwrap()` than `if let Some(x) = x` if `x` is guaranteed to be filled.

</details>

Ok nevermind, it looks like we actually *can* get two symbols with identical location, using good old C macros.

```c
#define CLEVER_MACRO int x, y;

CLEVER_MACRO;
```

clangd gives me

	test.c:3:1:Variable x
	test.c:3:1:Variable y

so the stable sort is actually important, because otherwise the order might change randomly even if clangd is consistent

src/util.rs
sidkshatriya: I don't think we can assume that all LSPs will return `DocumentSymbol`. Let me check the LSPs for which I had added workarounds. If they are now returning `DocumentSymbols` I will remove workarounds for them.

rc/lsp.kak
sidkshatriya: I tried doing it the way you suggested. Each completion option needs to be separated by a newline so the `for` loop does seem necessary.

We probably can avoid the `for` via some smart bash-ism though.

rc/lsp.kak
sidkshatriya: I actually like these conveniences and would request they remain for now. I'm OK if you want to remove them in the future. This feature is useful in language servers where you get symbols for variables and it quickly becomes tedious to goto them instead of functions which you're mainly interested in.

Of course, you could do `lsp-next-symbol Function` but that's a bit more verbose.

rc/lsp.kak
sidkshatriya: This was a good suggestion and I have created `lsp-next-or-previous-symbol` and everything is funneled through it.

sidkshatriya:
@krobelus I've had another round of additions. If you get the chance, please have a look!

You have done the heavy lifting for the `indoc!` macro feature -- it would be best if you pushed up code for that to this branch...

That leaves some other minor rust review items. I will deal with them once you have pushed up `indoc!` related changes...

rc/lsp.kak
krobelus: You wrote `printf '%s\n' "${symbol_kinds}"` but it should be `printf '%s\n' ${symbol_kinds}`.
(We need to word-split somewhere, whether in the for-loop or in the args to printf. I just find the implicit looping quite elegant)

sidkshatriya: Nice!

src/language_features/kak_lsp_custom.rs
krobelus: my bad, I missed that we only return in the empty case, sorry

sidkshatriya:
@krobelus Many servers are now sending us `DocumentSymbol` instead of `SymbolInformation`. However both our (current, on master) `textDocument/documentSymbol` features and the next/prev features are kinda "broken".

That is because `DocumentSymbol` is a nested JSON. If you open kakoune's `client.cc` file and try to see document symbol you will see what I mean. We can only see the Namespace Kakoune as the solitary symbol. The child symbols occur in the `children` key. `children` can be inside `children` and so on (think tree).

The fix would be to flatten the JSON. I plan to look at this perhaps tomorrow to see if we can fix this quickly. Just wanted to give you a heads up in case people complained.

src/util.rs
sidkshatriya: I removed the special case for elixir-ls as it is now sending DocumentSymbol. Erlang still sends SymbolInformation so special case for that remains.

However at the current moment `textDocument/documentSymbol` is somewhat broken on this branch and master. See https://github.com/kak-lsp/kak-lsp/pull/557#issuecomment-969154614

krobelus:
thanks, I have fixed the child handling of lsp-document-symbol> Same for the new commands, see the `pr-557` branch.
I have also fixed goto if the cwd is different from the project root.
Also I've fixed argument handling and `lsp-next-symbol` and <s>butchered the changes</s> made some cosmetic changes.
Only missing docs, then this should be ready.

I wonder if it's a good idea to define a shared trait between `DocumentSymbol` and `SymbolInformation`.. like the one for call hierarchy.

rc/lsp.kak
krobelus: Ok I guess it's useful to give the function/method variants more exposure since they are the 80% here.

> I'm OK if you want to remove them in the future.

Well I'm not gonna randomly remove features.
But we can decide later, of course.

sidkshatriya:
@krobelus Thanks for the various additions, changes and cleanups you did on d7e9ef445e3723c1930708ab0ff2e2392ce844e9 on https://github.com/kak-lsp/kak-lsp/tree/pr-557 

However, I am a bit uncomfortable that the commit message lists me as the sole author. 

I think it would be only fair if you added the following to the commit message:

```
Co-authored-by: Johannes Altmanninger <aclopte@gmail.com>
```

_Can we do this? I would not be comfortable otherwise :-)_

---

At first glace I don't think the problem of `children` has been resolved for `DocumentSymbol`. You are flattening the list only 1 deep. The JSON could be arbitrarily deep.

sidkshatriya:
* I now notice the recursive calls for children in both `fn next_or_prev_document_symbol_details()` and `format_document_symbol()` -- so this should work properly for nested `children` -- Great!
* As mentioned previously, I would request that the PR 557 commit message have `Co-authored-by: Johannes Altmanninger <aclopte@gmail.com>` as you have made contributions to the code. _Please do not merge commit to `master` without that_!

krobelus:
I'll merge with some more small more changes. Let me know if you have some feedback, I have reasons for everything but didn't write everything down.

I just realized that the goto feature should probably be provided as a Kakoune text object.
Then people can user familiar bindings like `[X`, `{X`, `<a-i>X` and `<a-a>X` where X is our key for symbol/function/method.
Of course that's something for the future.
Let's find chars that are not used by the core or https://github.com/Delapouite/kakoune-text-objects

Anyway, the current changes are already great, it's really much nicer to use than stock Kakoune commands, especially when functions have long parameter lists and docstrings.

> You are deep within a huge function. You have forgotten the name of the function you are in or want to know what its parameters are.

Maybe it's more obvious to only provide the jump functionality (the user can use `<c-o>` to jump back).
Anyway, that's something to try for the future.

sidkshatriya:
> Maybe it's more obvious to only provide the jump functionality (the user can use <c-o> to jump back).

The problem with a pure jump approach is the shift in context/loss of mental concentration as the cursor moves around. Without moving my cursor I want to know which function I'm in so this hover feature is useful to me.

(It would be natural that I would be attached to the hover feature :-) )

> Also we should probably provide a way to go to the next Fucntion OR Method.

Good point -- I had thought about it too but this PR was already getting a bit unwieldy. This is easily added in the future.

> Then people can user familiar bindings like [X, {X, <a-i>X and <a-a>X where X is our key for symbol/function/method.

Yes this is a good idea.

PR #558, opened 2021-11-08
Shorter Semantic Tokens Syntax

topisani:
The Syntax is:
```toml
"token+mod1+mod2" = "face"
```

This makes semantic tokens a lot easier to configure, and should have the same expressivity. I know the syntax has already changed at least once, but i feel this improvement is large enough to justify the hassle of upgrading

It also allows for entries without the token field, for when faces should be selected only based on modifiers.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -349,15 +349,12 @@ The syntax for such an entry is
 
 [source,toml]
 ----
-[[semantic_tokens]]
-token = "variable"
-face = "const_variable_declaration"
-modifiers = ["constant", "declaration"]
+"[token][+modifier1[+modifier2]...]" = "face"
 ----
 
-where `token` is the token's name as reported by the language server (see `lsp-capabilities`), `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config) and `modifiers` is an array of modifier names (also reported by the language server). `modifiers` may be omitted, but `token` and `face` are required.
+where `token` is the token's name as reported by the language server (see `lsp-capabilities`), `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config) and `modifiers` is an array of modifier names (also reported by the language server). `modifiers` or `token` may be omitted. An entry with no token will match any token (but should still have modifiers)
 
-You may create any arbitrary number of definitions with permutations between the token names and modifiers reported by the server. For an entry to match a token, all the entry's modifiers must exist on the token. However, the token may have additional modifiers not assigned in the config entry. +
+You may create any arbitrary number of definitions with permutations between the token names and modifiers reported by the server. For an entry to match a token, all the entry's modifiers must exist on the token. However, the token may have additional modifiers not assigned in the config entry.
 `kak-lsp` will find the most specific matching configuration to apply, where specificity is defined as the number of matching modifiers. If multiple matching entries have the same number of modifiers, the one that was defined last in the configuration wins.
 
 *Example:*
@@ -366,19 +363,10 @@ Assuming the following configuration,
 
 [source,toml]
 ----
-[[semantic_tokens]]
-token = "variable"
-face = "const_variable_declaration"
-modifiers = ["constant", "declaration"]
-
-[[semantic_tokens]]
-token = "variable"
-face = "const_variable"
-modifiers = ["constant"]
-
-[[semantic_tokens]]
-token = "variable"
-face = "variable"
+[semantic_tokens]
+"variable+constant+declaration" = "const_variable_declaration"
+"variable+constant" = "const_variable"
+"variable" = "variable"
 ----
 
 `kak-lsp` will perform these mappings:
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -352,7 +352,13 @@ The syntax for such an entry is
 "[token][+modifier1[+modifier2]...]" = "face"
 ----
 
-where `token` is the token's name as reported by the language server (see `lsp-capabilities`), `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config) and `modifiers` is an array of modifier names (also reported by the language server). `modifiers` or `token` may be omitted. An entry with no token will match any token (but should still have modifiers)
+where `token` is the token's name as reported by the language server (see `lsp-capabilities`), `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config) and each `modifier` is a modifier name (also reported by the language server).
+
+If the `token` or all `modifier`s are omitted, they will act as "match anything" for their respective category.
+Example:
+- `"variable" = "my_face"` will match any `variable` token, regardless of modifiers
+- `"+readonly" = "my_face"` will match any token that's flagged as `readonly`
+- `"" = "my_face"` will match everything. You should avoid this, as it will result in huge, inefficient highlighter ranges. Use a generic `fill` highlighter from Kakoune instead.
 
 You may create any arbitrary number of definitions with permutations between the token names and modifiers reported by the server. For an entry to match a token, all the entry's modifiers must exist on the token. However, the token may have additional modifiers not assigned in the config entry.
 `kak-lsp` will find the most specific matching configuration to apply, where specificity is defined as the number of matching modifiers. If multiple matching entries have the same number of modifiers, the one that was defined last in the configuration wins.
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,5 +1,6 @@
 use jsonrpc_core::{Call, Output, Params};
 use lsp_types::{DiagnosticSeverity, Position, Range, SemanticTokenModifier};
+use serde::de::{MapAccess, SeqAccess, Visitor};
 use serde::{de::Error as SerdeError, Deserialize, Deserializer, Serialize};
 use serde_json::Value;
 use std::borrow::Cow;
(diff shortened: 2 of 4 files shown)


README.asciidoc
```diff
@@ -316,15 +316,12 @@ The syntax for such an entry is
 
 [source,toml]
 ----
-[[semantic_tokens]]
-token = "variable"
-face = "const_variable_declaration"
-modifiers = ["constant", "declaration"]
+"[token][+modifier1[+modifier2]...]" = "face"
```
krobelus, line 319: Looks reasonable. This is a nice improvement, and validation is still fairly easy.

Regarding the breakage, we probably should have left these commented in `kak-lsp.toml`, or maybe extend dynamic configuration options, so that most (casual) users don't need a `kak-lsp.toml`.

The only thing I'm not sure about is that given `token+modifier1+modifier2 = face`, the `token` does not really stand out.
I considered both

- `token+modifier1,modifier2 = face`
- `token+modifier1|modifier2 = face`

but since precedence is not obvious, your version may be better, not sure.
The `+` is a good choice because it's also used for Kakoune attributes.

topisani: I get the syntax concern, but I think the fact that it matches kakoune attributes syntax, combined with the simplicity, makes it a good option.

Regarding dynamic configuration, I haven't used it personally, and don't know a lot about it. But it might make sense to do something special for this as well

src/types.rs
```diff
@@ -85,7 +85,16 @@ fn deserialize_semantic_tokens<'de, D>(
 where
     D: Deserializer<'de>,
 {
-    Vec::deserialize(deserializer).map_err(|e| {
```
krobelus, line 88: we could include the old deserializer as fallback

I get that we do want to tell users about the new syntax, but I suspect that most people don't even use semantic tokens, so they don't really care. Of course we make other improvements to `kak-lsp.toml`, so it's good practice to start fresh from the default config once in a while. Pressing the issue is not necessary, but we didn't have any complaints last time, so I doubt it matters. LGTM either way.

README.asciidoc
```diff
@@ -316,15 +316,12 @@ The syntax for such an entry is
...
 ----
 
-where `token` is the token's name as reported by the language server (see `lsp-capabilities`), `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config) and `modifiers` is an array of modifier names (also reported by the language server). `modifiers` may be omitted, but `token` and `face` are required.
+where `token` is the token's name as reported by the language server (see `lsp-capabilities`), `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config) and `modifiers` is an array of modifier names (also reported by the language server). `modifiers` or `token` may be omitted. An entry with no token will match any token (but should still have modifiers)
```
sclu1034, line 322: The part about the modifiers array needs to be changed as well, and with more components being optional, examples might help to clarify functionality:

```suggestion
where `token` is the token's name as reported by the language server (see `lsp-capabilities`), `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config) and each `modifier` is a modifier name (also reported by the language server).

If the `token` or all `modifier`s are omitted, they will act as "match anything" for their respective category.
Example:
- `"variable" = "my_face"` will match any `variable` token, regardless of modifiers
- `"+readonly" = "my_face"` will match any token that's flagged as `readonly`
- `"" = "my_face"` will match everything. You should avoid this, as it will result in huge, inefficient highlighter ranges. Use a generic `fill` highlighter from Kakoune instead.
```

sclu1034 (changes requested):

src/types.rs
sclu1034: Since the change is just a "cosmetic" syntax change, I think it would be fine to keep both.
Especially since "expressive names vs. concise definitions" is usually about personal preference, and we do have the ability to offer both.

krobelus: > `"" = "my_face"` will match everything.

you mean every token?
Or the entire buffer? In that case we should probably reject it.

sclu1034: Every token individually.
Rejecting it would probably make sense either way, since there really is no benefit to it.

topisani: I agree, rejecting the empty string makes sense

README.asciidoc
```diff
@@ -333,19 +330,10 @@ Assuming the following configuration,
...
-face = "const_variable"
-modifiers = ["constant"]
-
-[[semantic_tokens]]
-token = "variable"
-face = "variable"
+[semantic_tokens]
+"variable+constant+declaration" = "const_variable_declaration"
+"variable+constant" = "const_variable"
+"variable" = "variable"
```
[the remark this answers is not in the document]
topisani: I would be fine with this syntax as well. I think the short syntax is important to make any sort of styling feasible. Especially for something like rust-analyzer, which provides a lot of token and modifier types.

topisani:
I adopted the syntax proposed by @sclu1034, which is not as sexy, but preserves the  order, which can be important. 
```rust
semantic_tokens = [
  [ "variable+constant+declaration", "const_variable_declaration" ],
  [ "variable+constant", "const_variable" ],
  [ "variable", "variable" ],
]
```

I also now accept the old syntax as well, so we dont break any configs, and people who want the more explicit syntax can keep that.

krobelus:
Thanks for taking the initiative! The deserialization looks solid.

I started looking into this and realized we can already use [TOML inline tables](https://github.com/toml-lang/toml/blob/master/toml.md#user-content-inline-table) today.
Do you think this is concise enough?

```toml
semantic_tokens = [
    { token = "comment", modifiers = ["documentation"], face = "documentation" },
    { token = "comment", face = "comment" },
]
```

sclu1034:
While this does make the configuration _shorter_ by a few characters, I don't think the latest iterations will be _simpler/easier_ for the average user.
Dropping the named indices of tables already loses a certain kind of "self documentation" and when everything is a single string, TOML syntax highlighting can't help distinguish individual fields anymore.

And while we could allow all of these syntax variations at once, that would likely lead to confusion as well and will make code maintenance harder.

I think the current implementation is a good enough compromise.

krobelus:
I agree that when [setting >30 faces](https://github.com/topisani/dotfiles/blob/master/config/kak-lsp/kak-lsp.toml#L180-L213), it's much nicer if they only need one line each, with a high signal-to-noise ratio.

> ```toml
> semantic_tokens = [
>   "comment+documentation=documentation",
>   "comment=comment"
> ]
> ```
> While this  gets even further out of toml, i think it is more ergonomic than the nested arrays syntax, its as easy to parse, and still makes a lot of sense to kakoune users imo. 

yeah that looks better than the nested arrays

Alternatively we could allow a raw string instead of a TOML structure:

	semantic_tokens = """
	comment + documentation = documentation
	# Comments should be allowed.
	comment = comment
	function = function
	keyword = keyword
	namespace = module
	operator = operator
	string = string
	type = type
	variable + readonly = default+d
	variable + constant = default+d
	variable = variable
	"""

not sure which one is better.

> Dropping the named indices of tables already loses a certain kind of "self documentation"

We'll have plenty documentation and examples in comments above, so I think that's okay.
It might take a minute to get used to the syntax but that's probably worth it.

> TOML syntax highlighting can't help distinguish individual fields anymore.

The + is indeed a bit subtle but it's also used in Kakoune colorschemes. Maybe adding a space helps?

> And while we could allow all of these syntax variations at once, that would likely lead to confusion as well and will make code maintenance harder.

True but I think it's more important to make this file easy to read and write by hand.

Is it possible that a rust-analyzer semantic tokens configuration still has "redundancy"? Like, would it be nicer to generate some of it with a script? In that case we might want to think about a kakscript solution.
I haven't really used semantic tokens much myself.

krobelus:
I've opened https://github.com/kak-lsp/kak-lsp/pull/587 which implements the inline table variant.
I don't use semantic tokens myself so I'm not sure.

> especially when lines become long enough to need wrapping, the table syntax quickly gets out of hand.

How long is too long here and does it happen often? In my inline table patch the longest line has 97 characters:

    {token="variable", face="const_variable_declaration", modifiers=["constant", "declaration"]},

I think that's still acceptable but that's a matter of preference.
The array-of-strings syntax would be fine me as well. It's a bit shorter but also a bit less obvious.

sclu1034:
I think most of the "OG" dev communities are still very loud about keeping 80 columns max width, but 100 (rustfmt default) and 140 columns make second and third place these days.

Your snippet does fit 100 columns, but only until we add subjective whitespace formatting, e.g.
```toml
semantic_tokens = [
    { token = "variable", face = "const_variable_declaration", modifiers = ["constant", "declaration"] },
]
```
That's 105 columns.

However, in this case the `face` name is quite long, and multiple modifiers will likely not be common. So I think it would be fine if those few cases would need to be broken into separate lines.

```toml
semantic_tokens = [
    { token = "variable", face = "variable" },
    {
		token = "variable",
		face = "const_variable_declaration",
		modifiers = ["constant", "declaration"],
	},
]
```

So I'm fine with #587.

krobelus:
> {
>	token = "variable",
>	face = "const_variable_declaration",
>	modifiers = ["constant", "declaration"],
> },

Sadly this is not allowed in TOML. Inline tables must be in one line (unless the values contain newlines).
Probably not a big deal anyway.

PR #563, opened 2021-11-17
Cleanup: remove some unused/redundant code related to #557

sidkshatriya:
Cleanup: remove some unused/redundant code related to #557

- We don't ever use `HoverModalParams` from the editor so we should just remove the ability to pass in parameters to kak-lsp. We never need to _use_ the serialization as of now. 
- `style` is no longer used
- `context` comes with `struct HoverModalParams` so it can be removed
- We never need to serialize/deserialize the struct `HoverModal`

If we every find a use for these things they can be added again.


krobelus:
thanks

Issue #514, opened 2021-09-04
Show hover information in a scratch buffer


krobelus:
Currently, `:lsp-hover` uses the info box. That works well for many simple cases, however, it is quite limited. There are some issues with the info box:

- open files/links from hover: #409
- hover info can be too large (without setting `lsp_hover_max_lines`): #75, #257
- let hover info stick around while writing function call arguments: #387
- info box (and completion popup) should have highlighting: #395, #476

## Proposal
- Add a `*hover*` buffer, which will always hold the latest hover information. This should solve all issues above.
- `lsp-hover` could open the `*hover*` buffer in `%opt{docsclient}`. If no docsclient is defined, use the infobox and/or a temporary client.

The main UI problem here is that this is usually meant to be a popup window - it shouldn't be visible all the time. Inside tmux we can use temporary splits or the new tmux popup windows, but in other windowing systems we have less control over temporary clients.

So the recommended workflow for non-tmux would be to always have the docsclient visible.
This might waste screen estate, but the buffer can be reused as toolsclient. Alternatively, users can switch-to/delete the `*hover*` buffer.

PR #564, opened 2021-11-18
Feature: Allow to show hover info in scratch buffer

sidkshatriya:
info_in_hoverclient

Sometimes the hover information sent to kakoune by kak-lsp is either
very long or has rich markdown (or both). The hover info is shown in an
info box which can be very difficult to read.

Add a new command `lsp-hover-in-hover-client` which will open the hover
information in a new client. You also get the benefit of syntax
highlighting as the codeblocks are enclosed within three ticks
(with the language name) and kak understands that.

The user can continue issuing `lsp-hover-in-hover-client` commands
and they will appear in the `hoverclient`. The focus will not
shift from the window in which the user was typing in.

This feature can be very visually appealing if you use kakoune
with something like tmux however it will work properly even without it.

If a `hoverclient` does not exist it will be created via `new`.

**Try this out**

If you are using rust-analyzer then place your cursor over any
of the characters of `fn` or any of the characters of `let` and
issue the above command (**keyboard shortcut is _lsp_ H**).

You will see great documentation for `fn` / `let`.

**Why not just use goto definition?**

Many languages have the documentation for a function above it.
But you cannot goto definition in all situations. For instance, you
cannot goto definition on the `let` statement in Rust.

This feature can be especially useful for [ocaml lsp server](https://github.com/ocaml/ocaml-lsp) to be able to
see long signatures. In my testing other LSPs that give you
a lot of hover info become more usable.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -230,8 +230,18 @@ define-command lsp-hover -docstring "Request hover info for the main cursor posi
     lsp-did-change-and-then lsp-hover-request
 }
 
-define-command -hidden lsp-hover-request -docstring "Request hover info for the main cursor position" %{
-    nop %sh{ (printf '
+define-command lsp-hover-in-hover-client -docstring "Request hover info for the main cursor position in a separate client" %{
+    lsp-did-change-and-then "lsp-hover-request in_hover_client"
+}
+
+define-command -hidden lsp-hover-request -params 0..1 -docstring "Request hover info for the main cursor position" %{
+    nop %sh{
+        maybe_hover_fifo=""
+        if [ "$1" = "in_hover_client" ]; then
+            maybe_hover_fifo="hoverFifo = \"${kak_opt_lsp_hover_fifo}\""
+        fi
+
+        (printf '
 session   = "%s"
 client    = "%s"
 buffile   = "%s"
@@ -239,9 +249,10 @@ filetype  = "%s"
 version   = %d
 method    = "textDocument/hover"
 [params]
+%s
 position.line = %d
 position.column = %d
-' "${kak_session}" "${kak_client}" "${kak_buffile}" "${kak_opt_filetype}" "${kak_timestamp}" ${kak_cursor_line} ${kak_cursor_column} | eval ${kak_opt_lsp_cmd} --request) > /dev/null 2>&1 < /dev/null & }
+' "${kak_session}" "${kak_client}" "${kak_buffile}" "${kak_opt_filetype}" "${kak_timestamp}" "$maybe_hover_fifo" ${kak_cursor_line} ${kak_cursor_column} | eval ${kak_opt_lsp_cmd} --request) > /dev/null 2>&1 < /dev/null & }
 }
 
 declare-option -hidden str lsp_symbol_kind_completion %{
@@ -1526,6 +1537,7 @@ map global lsp d '<esc>: lsp-definition<ret>'             -docstring 'go to defi
 map global lsp e '<esc>: lsp-diagnostics<ret>'            -docstring 'list project errors, info, hints and warnings'
 map global lsp f '<esc>: lsp-formatting<ret>'             -docstring 'format buffer'
 map global lsp h '<esc>: lsp-hover<ret>'                  -docstring 'show info for current position'
+map global lsp H '<esc>: lsp-hover-in-hover-client<ret>'   -docstring 'show info for current position in separate client'
 map global lsp i '<esc>: lsp-implementation<ret>'         -docstring 'go to implementation'
 map global lsp j '<esc>: lsp-outgoing-calls<ret>'         -docstring 'list outgoing call for function at cursor'
 map global lsp k '<esc>: lsp-incoming-calls<ret>'         -docstring 'list incoming call for function at cursor'
@@ -1943,4 +1955,13 @@ define-command lsp-goto-next-match -docstring 'DEPRECATED: use lsp-next-location
 
 define-command lsp-goto-previous-match -docstring 'DEPRECATED: use lsp-previous-location. Jump to the previous goto match' %{
     lsp-previous-location '*goto*'
+}
+
+declare-option -hidden str lsp_hover_fifo
+
+set-option global lsp_hover_fifo %sh{
+    tmpdir=$(mktemp -q -d -t 'lsp-hover-client.XXXXXX' 2>/dev/null || mktemp -q -d)
+    output="$tmpdir/fifo"
+    mkfifo "$output"
+    echo "$output"
 }
--- a/src/language_features/document_symbol.rs
+++ b/src/language_features/document_symbol.rs
@@ -236,8 +236,7 @@ fn editor_next_or_prev_for_details(
         work_done_progress_params: Default::default(),
     };
 
-    // This context is shown at the top of the modal.
-    let context = format!(
+    let modal_heading = format!(
         "line {}:{}:{{+b@KindAndName}}{:?} {}{{KindAndName}} \
             (Press 'g' to goto this position. Press any other key to continue)",
         symbol_position.line,
@@ -263,7 +262,10 @@ fn editor_next_or_prev_for_details(
     ctx.call::<HoverRequest, _>(meta, req_params, move |ctx: &mut Context, meta, result| {
         editor_hover(
             meta,
-            Some(HoverModal { context, do_after }),
+            HoverType::Modal {
+                modal_heading,
+                do_after,
+            },
             PositionParams {
                 position: symbol_position,
             },
--- a/src/language_features/hover.rs
+++ b/src/language_features/hover.rs
@@ -1,3 +1,5 @@
+use std::fs;
+
 use crate::context::*;
 use crate::markup::*;
 use crate::position::*;
@@ -9,8 +11,17 @@ use lsp_types::*;
 use serde::Deserialize;
 use url::Url;
 
-pub fn text_document_hover(meta: EditorMeta, editor_params: EditorParams, ctx: &mut Context) {
-    let params = PositionParams::deserialize(editor_params.clone()).unwrap();
+pub fn text_document_hover(meta: EditorMeta, params: EditorParams, ctx: &mut Context) {
+    let maybe_hover_fifo = HoverDetails::deserialize(params.clone())
+        .unwrap()
+        .hover_fifo;
+
+    let hover_type = match maybe_hover_fifo {
+        Some(fifo) => HoverType::InfoInHoverClient { fifo },
+        None => HoverType::Normal,
+    };
+
+    let params = PositionParams::deserialize(params).unwrap();
     let req_params = HoverParams {
         text_document_position_params: TextDocumentPositionParams {
             text_document: TextDocumentIdentifier {
@@ -21,13 +32,13 @@ pub fn text_document_hover(meta: EditorMeta, editor_params: EditorParams, ctx: &
         work_done_progress_params: Default::default(),
     };
     ctx.call::<HoverRequest, _>(meta, req_params, move |ctx: &mut Context, meta, result| {
-        editor_hover(meta, None, params, result, ctx)
+        editor_hover(meta, hover_type, params, result, ctx)
     });
 }
 
 pub fn editor_hover(
     meta: EditorMeta,
-    maybe_hover_modal: Option<HoverModal>,
+    hover_type: HoverType,
     params: PositionParams,
     result: Option<Hover>,
     ctx: &mut Context,
@@ -90,7 +101,7 @@ pub fn editor_hover(
         .get(&ctx.language_id)
         .and_then(|l| l.workaround_server_sends_plaintext_labeled_as_markdown)
         .unwrap_or(false);
-    let mut contents = match result {
+    let contents = match result {
         None => "".to_string(),
         Some(result) => match result.contents {
             HoverContents::Scalar(contents) => {
@@ -105,43 +116,112 @@ pub fn editor_hover(
                     FACE_INFO_RULE, FACE_INFO_DEFAULT
                 )),
             HoverContents::Markup(contents) => match contents.kind {
-                MarkupKind::Markdown => markdown_to_kakoune_markup(contents.value, force_plaintext),
+                MarkupKind::Markdown => {
+                    if let HoverType::InfoInHoverClient { .. } = hover_type {
+                        contents.value
+                    } else {
+                        markdown_to_kakoune_markup(contents.value, force_plaintext)
+                    }
+                }
                 MarkupKind::PlainText => contents.value,
             },
         },
     };
 
-    if contents.is_empty() && diagnostics.is_empty() && maybe_hover_modal.is_none() {
-        return;
-    }
+    match hover_type {
+        HoverType::Normal => {
+            if contents.is_empty() && diagnostics.is_empty() {
+                return;
+            }
 
-    let anchor_or_style = if let Some(hover_modal) = maybe_hover_modal.as_ref() {
-        contents = format!("{}\n---\n{}", hover_modal.context, contents);
-        "modal".to_string()
-    } else {
-        format!("{}", params.position)
+            let command = format!(
+                "lsp-show-hover {} %§{}§ %§{}§",
+                params.position,
+                contents.replace("§", "§§"),
+                diagnostics.replace("§", "§§"),
+            );
+            ctx.exec(meta, command);
+        }
+        HoverType::Modal {
+            modal_heading,
+            do_after,
+        } => {
+            show_hover_modal(meta, ctx, modal_heading, do_after, contents, diagnostics);
+        }
+        HoverType::InfoInHoverClient { fifo } => {
+            show_hover_in_hover_client(meta, ctx, fifo, contents);
+        }
     };
+}
 
-    let mut command = format!(
-        "lsp-show-hover {} %§{}§ %§{}§",
-        anchor_or_style,
+fn show_hover_modal(
+    meta: EditorMeta,
+    ctx: &Context,
+    modal_heading: String,
+    do_after: String,
+    contents: String,
+    diagnostics: String,
+) {
+    let contents = format!("{}\n---\n{}", modal_heading, contents);
+    let command = format!(
+        "lsp-show-hover modal %§{}§ %§{}§",
         contents.replace("§", "§§"),
-        diagnostics.replace("§", "§§")
+        diagnostics.replace("§", "§§"),
     );
-
-    // Wrap if we're using a HoverModal
-    if let Some(hover_modal) = maybe_hover_modal {
-        command = formatdoc!(
-            "eval %§
+    let wrapped_command = formatdoc!(
+        "eval %§
                  {}
                  {}
              §",
-            command.replace("§", "§§"),
-            hover_modal.do_after.replace("§", "§§")
-        );
+        command.replace("§", "§§"),
+        do_after.replace("§", "§§")
+    );
+    ctx.exec(meta, wrapped_command);
+}
+
+fn show_hover_in_hover_client(
+    meta: EditorMeta,
+    ctx: &Context,
+    hover_fifo: String,
+    contents: String,
+) {
+    if contents.is_empty() {
+        return;
     }
 
+    let handle = std::thread::spawn(move || {
+        // Note that we don't show diagnostics in the hover client
+        fs::write(hover_fifo, contents.as_bytes()).unwrap();
+    });
+
+    let command = formatdoc!(
+        "
+        %[
+            edit! -existing -readonly -fifo %opt[lsp_hover_fifo] *hover*
+            set buffer=*hover* filetype markdown
+        ]"
+    );
+
+    let client = meta.client.clone().unwrap_or_default();
+    let command = formatdoc!(
+        "
+        try %[
+            eval -client hoverclient {}
+        ] catch %[
+            new %[
+                rename-client hoverclient
+                eval {}
+                addhl -override window/wrap wrap
+                focus {}
+            ]
+        ]",
+        command,
+        command,
+        client
+    );
+
     ctx.exec(meta, command);
+    handle.join().unwrap();
 }
 
 trait PlainText {
--- a/src/types.rs
+++ b/src/types.rs
@@ -156,6 +156,12 @@ pub struct PositionParams {
     pub position: KakounePosition,
 }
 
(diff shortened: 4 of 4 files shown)


sidkshatriya:
It builds fine in rust nightly. Seems like rust stable has a problem. Checking...

sidkshatriya:
Fixed. Builds on stable.

src/language_features/hover.rs
```diff
@@ -105,43 +116,112 @@ pub fn editor_hover(
...
+            edit! -existing -readonly -fifo %opt[lsp_hover_fifo] *hover*
+            set buffer=*hover* filetype markdown
+        ]"
+    );
+
+    let client = meta.client.clone().unwrap_or_default();
+    let command = formatdoc!(
+        "
+        try %[
+            eval -client hoverclient {}
```
krobelus, line 209: This is a great improvement, which solves lots of long-standing issues.
A dedicated buffer is much more powerful than the info box.

In my quick planning over at https://github.com/kak-lsp/kak-lsp/issues/514 I suggested to use `%opt{docsclient}`, which is already used by `:doc`, `:man` and `:git`. However, since we want to (also) create this new client, your solution seems better.

I personally don't use `%opt{docsclient}` but I can imagine that someone would want this to always spawn in their docsclient.
Maybe it's worth adding a second attempt here: 

```kak
        try %[
            eval -client hoverclient {}
	] catch %[
            eval -client "%opt{docsclient}" {}
	] catch %[
	    new ...
	]
```

Not completely sure yet. We can also decide later.

One alternative is to have `lsp-hover` accept an optional client name as argument.
The only disadvantage I see is that it's less discoverable (you'd have to type `:lsp-hover` and read the docstring in the info box).

rc/lsp.kak
```diff
@@ -1944,3 +1956,12 @@ define-command lsp-goto-next-match -docstring 'DEPRECATED: use lsp-next-location
 define-command lsp-goto-previous-match -docstring 'DEPRECATED: use lsp-previous-location. Jump to the previous goto match' %{
     lsp-previous-location '*goto*'
 }
+
+declare-option -hidden str lsp_hover_fifo
+
+set-option global lsp_hover_fifo %sh{
+    tmpdir=$(mktemp -q -d -t 'lsp-hover-client.XXXXXX' 2>/dev/null || mktemp -q -d)
+    output="$tmpdir/fifo"
+    mkfifo "$output"
```
krobelus, line 1965: We should probably create this lazily (the first time the user requests the hover-in-in-hoverclient)

Also we should clean up the tempdir. Maybe in `lsp-exit-editor-session`/`lsp-stop` which are run on our KakEnd hook -- they also run `remove-hooks global lsp`.

Other commands like `lsp-replace-selection` clean up their temp files immediately.
Though the persistent FIFO seems fine.

src/language_features/document_symbol.rs
```diff
@@ -236,8 +236,7 @@ fn editor_next_or_prev_for_details(
         work_done_progress_params: Default::default(),
     };
 
-    // This context is shown at the top of the modal.
-    let context = format!(
+    let modal_heading = format!(
         "line {}:{}:{{+b@KindAndName}}{:?} {}{{KindAndName}} \
```
krobelus, line 240: BTW instead of KindAndName we could use one of our established faces like `InfoHeader`

src/language_features/hover.rs
```diff
@@ -1,3 +1,5 @@
+use std::fs;
+
```
krobelus, line 2: Separating stdlib from other imports with a blank line seems unconventional for this project. Should wed do that?

src/language_features/hover.rs
```diff
@@ -105,43 +116,112 @@ pub fn editor_hover(
...
+        diagnostics.replace("§", "§§"),
     );
-
-    // Wrap if we're using a HoverModal
-    if let Some(hover_modal) = maybe_hover_modal {
-        command = formatdoc!(
-            "eval %§
+    let wrapped_command = formatdoc!(
+        "eval %§
                  {}
```
krobelus, line 173: the brace should be four spaces right of the "e" from eval

src/language_features/hover.rs
```diff
@@ -105,43 +116,112 @@ pub fn editor_hover(
...
-        diagnostics.replace("§", "§§")
+        diagnostics.replace("§", "§§"),
     );
-
-    // Wrap if we're using a HoverModal
-    if let Some(hover_modal) = maybe_hover_modal {
-        command = formatdoc!(
-            "eval %§
+    let wrapped_command = formatdoc!(
+        "eval %§
```
krobelus, line 172: BTW I think we should eventually move towards using long names only ("evaluate-commands"), just like in kak scripts

src/language_features/hover.rs
```diff
@@ -105,43 +116,112 @@ pub fn editor_hover(
...
     }
 
+    let handle = std::thread::spawn(move || {
+        // Note that we don't show diagnostics in the hover client
+        fs::write(hover_fifo, contents.as_bytes()).unwrap();
+    });
+
+    let command = formatdoc!(
+        "
+        %[
```
krobelus, line 199: I think we should keep the quote on the same line as the start of the content, that's much less wasteful for our current uses.
Then the next line needs +1 indentation, like

```
"%[
     edit! ...
 1234 
````

File: src/language_features/hover.rs [hunk 8]
```diff
@@ -105,43 +116,112 @@ pub fn editor_hover(
                     FACE_INFO_RULE, FACE_INFO_DEFAULT
                 )),
             HoverContents::Markup(contents) => match contents.kind {
-                MarkupKind::Markdown => markdown_to_kakoune_markup(contents.value, force_plaintext),
+                MarkupKind::Markdown => {
+                    if let HoverType::InfoInHoverClient { .. } = hover_type {
+                        contents.value
```
krobelus, line 121: yep, we need not interpret Markdown when we directly render it to a filetype=markdown file.

We should probably do the same in the other cases, where we get `MarkedString` objects

- `MarkedString::String` we can just pass on verbatim.
- `MarkedString::LanguageString` we should interpret as like this markdown snippet

    ```${language}
    ${value}
    ```

src/language_features/hover.rs
```diff
@@ -105,43 +116,112 @@ pub fn editor_hover(
...
+    let handle = std::thread::spawn(move || {
+        // Note that we don't show diagnostics in the hover client
+        fs::write(hover_fifo, contents.as_bytes()).unwrap();
+    });
+
+    let command = formatdoc!(
+        "
+        %[
+            edit! -existing -readonly -fifo %opt[lsp_hover_fifo] *hover*
+            set buffer=*hover* filetype markdown
```
krobelus, line 201: let's make this conditional on the server actually sending us Markdown.
If we get MarkupKind::PlainText, or if `force_plaintext` is true (\*cough* pyls \*cough*) then we should set the filetype to the empty string.

src/language_features/hover.rs
```diff
@@ -105,43 +116,112 @@ pub fn editor_hover(
...
+    ctx: &Context,
+    hover_fifo: String,
+    contents: String,
+) {
+    if contents.is_empty() {
+        return;
     }
 
+    let handle = std::thread::spawn(move || {
+        // Note that we don't show diagnostics in the hover client
```
krobelus, line 193: Not showing diagnostics seems a bit arbitrary, users are supposed to set `lsp_show_hover_format`.
But I guess we can change that later, if anyone complains.

src/language_features/hover.rs
```diff
@@ -105,43 +116,112 @@ pub fn editor_hover(
...
+        fs::write(hover_fifo, contents.as_bytes()).unwrap();
+    });
+
+    let command = formatdoc!(
+        "
+        %[
+            edit! -existing -readonly -fifo %opt[lsp_hover_fifo] *hover*
```
krobelus, line 200: not a fan of readonly buffers -- it's very nice to be able to edit buffers like `*grep*` -- and I don't see an advantage

src/language_features/hover.rs
```diff
@@ -105,43 +116,112 @@ pub fn editor_hover(
...
+    hover_fifo: String,
+    contents: String,
+) {
+    if contents.is_empty() {
+        return;
     }
 
+    let handle = std::thread::spawn(move || {
+        // Note that we don't show diagnostics in the hover client
+        fs::write(hover_fifo, contents.as_bytes()).unwrap();
```
krobelus, line 194: can we log an error if this write goes wrong? I'd need to try that, not super important of course. Maybe when we join the thread..

src/language_features/hover.rs
```diff
@@ -105,43 +116,112 @@ pub fn editor_hover(
...
+    });
+
+    let command = formatdoc!(
```
krobelus, line 197: if there are no arguments to format, I guess indoc! is preferred over formatdoc!

*However*, since this command is used twice with different indentation, I'd just write it in one line, to avoid weird-looking logs
Something like

```rust
    let command = "%[ \
            edit! -existing -readonly -fifo %opt[lsp_hover_fifo] *hover*; \
            set buffer=*hover* filetype markdown; \
            ]";
```

src/language_features/hover.rs
```diff
@@ -105,43 +116,112 @@ pub fn editor_hover(
...
+    let client = meta.client.clone().unwrap_or_default();
+    let command = formatdoc!(
+        "
+        try %[
+            eval -client hoverclient {}
+        ] catch %[
+            new %[
+                rename-client hoverclient
+                eval {}
+                addhl -override window/wrap wrap
```
krobelus, line 214:
```suggestion
                add-highlighter -override window/wrap wrap -word
```

src/types.rs
```diff
@@ -228,10 +234,16 @@ pub struct KakounePosition {
     pub column: u32, // in bytes, not chars!!!
 }
 
-#[derive(Clone, Debug)]
-pub struct HoverModal {
-    pub context: String,
-    pub do_after: String,
+#[derive(PartialEq)]
+pub enum HoverType {
```
krobelus, line 238: I wonder if there is a better name, like HoverUI, HoverTarget or HoverDestination.
Though these all sound awkward.

rc/lsp.kak
```diff
@@ -1526,6 +1537,7 @@ map global lsp d '<esc>: lsp-definition<ret>'             -docstring 'go to defi
 map global lsp e '<esc>: lsp-diagnostics<ret>'            -docstring 'list project errors, info, hints and warnings'
 map global lsp f '<esc>: lsp-formatting<ret>'             -docstring 'format buffer'
 map global lsp h '<esc>: lsp-hover<ret>'                  -docstring 'show info for current position'
+map global lsp H '<esc>: lsp-hover-in-hover-client<ret>'   -docstring 'show info for current position in separate client'
```
krobelus, line 1540: should add a line for this in README (in the "Either way you get:" list, next to `lsp-hover`)

(also alignment is off - I'll try to fix it for the entire block at some point)

src/language_features/hover.rs
```diff
@@ -105,43 +116,112 @@ pub fn editor_hover(
...
+                focus {}
```
krobelus, line 215: The implementation is interesting. I'd have expected this logic in the `lsp-show-hover` callback, like the info box / modal logic.

rc/lsp.kak
```diff
@@ -230,18 +230,29 @@ define-command lsp-hover -docstring "Request hover info for the main cursor posi
     lsp-did-change-and-then lsp-hover-request
 }
 
-define-command -hidden lsp-hover-request -docstring "Request hover info for the main cursor position" %{
-    nop %sh{ (printf '
+define-command lsp-hover-in-hover-client -docstring "Request hover info for the main cursor position in a separate client" %{
```
krobelus, line 233: The name repeats "hover", which isn't so nice.
Also "client" should be something like "buffer" (as opposed to the info box), also because it's a valid use case to have the `*hover*` buffer appear in the current client.

That's another thing that we might want to tweak:
by hardcoding the client name, it seems difficult to use the current client.
Granted, we can use `:rename-client hoverclient`, but that will fail if some other client is already called `hoverclient` from a previous call.
This makes the `docsclient` approach seem more attractive, because then we can use `set global docsclient %val{client}; lsp-hover-in-hover-client`

Anyway, if we don't change the interface, I'd say

```suggestion
define-command lsp-hover-buffer -docstring "Show hover info for the main cursor position in a scratch buffer" %{
```

sidkshatriya: > Granted, we can use :rename-client hoverclient, but that will fail if some other client is already called hoverclient from a previous call.

Is that possible? The renaming of the client is only happening in `catch` statement. And we landed up in the catch only because the `hoverclient` did not exist...

krobelus: I mean that the user can rename any client to `hoverclient` if they want the `*hover*` buffer to show up there, but an option would be more flexible

krobelus: Regarding the other discussion about `%opt{docsclient}` - I think we should just accept an optional argument, and in the default mapping call `:lsp-hover-buffer %opt{docsclient}`.
Then default to `hoverclient` if docsclient is not set.

(Of course I can do this extension - I just like to get the commits into a "mostly done" state before merging so the repo history is easy to read)

src/language_features/hover.rs
krobelus, line 209: one thing that came up when testing this with x11 windowing, is that the hoverclient might be in the background.
I considered running `focus` to make it visible, and then another `focus` to go back to the original client but that seems very hacky.
So let's just leave this responsibility to the user.

rc/lsp.kak
sidkshatriya: @krobelus 

> (Of course I can do this extension - I just like to get the commits into a "mostly done" state before merging so the repo history is easy to read)

I should have time in the next few days to deal with the review comments you have made. Thanks again for your useful and thoughtful review comments.

- I'll push up my changes in a very small number of commits on this branch, force pushing if necessary on this branch
- You are free to make further adjustments/improvements as _separate_ commits on the same branch (you should be able to push to this branch)
- We can then merge this branch when you're happy

Personally I would prefer that you don't `--amend` my commits to make any adjustments. Amending does keep the history simple if but confuses me because I may see a change and then not be sure if that is something you added in or I added and then I have to refer back to the original git history.

The consolidated feature code can always be viewed in the merge commit so we always have the ability to see what the code for the whole feature is.

Does this sound reasonably? We can also discuss the git workflow on email if you want...

sidkshatriya: P.S. And the merge from this branch should ideally be a normal merge rather than a squash merge -- that keeps things clear about what changes I made and what changes you made -- again we can discuss on email if needed...

krobelus: > Personally I would prefer that you don't `--amend` my commits to make any adjustments. Amending does keep the history simple if but confuses me because I may see a change and then not be sure if that is something you added in or I added and then I have to refer back to the original git history.

Cleaning up history may be a slight inconvenience to the author and first reviewer, but it helps *everyone* else, including future-you and future-me.
People who read the history later will arguably prefer a condensed history, and probably don't care about our old mistakes.

There are tools like git-range-diff that make it easier to see how history was rewritten.

I can read the history of a repository like https://github.com/git/git and quickly know what is actually going on because there is very little noise in the history (because they use a proper review process).
Of course we are much much smaller, so a clean history is not essential but it's still much nicer.

I must admit that I couldn't wait to merge the next/prev-symbol PR, so I ended up doing the cleanups myself.
Hence the confusion -- sorry -- I should have waited for your changes, and communicated how we should structure them.
I'm happy to review until we have something we're both happy with.

However, if review changes are added in separate commits with not-so-useful commit messages, that just makes the final history worse than if I just did the cleanups myself. (Also for example, the back-and-forth changes to `lsp_symbol_kind_completion` should go away because they cancel each other out.)

If you strongly prefer never rewriting your commits, that's okay too, then I'll probably just apply most review comments myself later.
I just don't see a good reason to treat old commits from pending PRs as if they are set in stone. For example, if a question comes up in code review that should be answered by the commit message, why not change the message.

> The consolidated feature code can always be viewed in the merge commit so we always have the ability to see what the code for the whole feature is.

In theory yes, but in practice git-blame will still point to a single commit, that would be more useful if the commit was a logical unit.

> P.S. And the merge from this branch should ideally be a normal merge rather than a squash merge -- that keeps things clear about what changes I made and what changes you made

I'm still not sure if it's better to have merge commits or not, but they seem fine.

src/language_features/hover.rs
krobelus: I'm starting to think that the default should be to display the `*hover*` buffer in the current client like `:doc`, `:grep` and friends do (and if docsclient is set, use that instead). The current client is the lowest common denominator, it will work alright everywhere, also when tmux is not available (even with tmux, it depends on the window layout whether the new split fits in well -- it could be too small).

On top of that, probably keep the current behavior if a client name is given as argument,

rc/lsp.kak
sclu1034: The problem here is that changing existing commits always feels to me like pulling the rug from under your contributors.
If I've been working on review comments or new stuff for a PR for a day or two until I consider it ready for pushing, `! [rejected]` is not fun to see. Especially when accompanied by a bunch of diffs that require me to change my mental model of the code I've been working with.

A clean history on merge does make sense, but, in my opinion, it should either be discussed in comments until all parties are on the same page or be postponed until the PR author considers their work complete and ready for merge/hand-off.
I always start out my PRs with "Allow edits" unchecked for that reason.

sidkshatriya: @sclu1034 @krobelus @ul
> The problem here is that changing existing commits always feels to me like pulling the rug from under your contributors.

I agree with you and I am happy that you have independently arrived at this conclusion without you and me knowing each other. 

It seems `--amend`ing commits of others and changing code and git commit comments seems to be a regular occurrence on this project.

In fact my singular commit on this branch has been again modified in "successor" #570. See how different https://github.com/kak-lsp/kak-lsp/pull/570/commits/a648be06bb6e78087d7ff57e621525fc8af997ff is to my original commit https://github.com/kak-lsp/kak-lsp/commit/15fabf5a2ec0ed1478ee245513b0684c60be1b37

This is even though I have made an explicit request in private communication that I would prefer that @krobelus is free to build upon my commit in _subsequent_ commits and make any changes that are deemed necessary. I just want my _original_ commit to be preserved.

The same story happened in https://github.com/kak-lsp/kak-lsp/pull/557 -- My commit was modified significantly and I have already objected to that in private email communication. At that point I was not aware this approach of modifying other people's code in this project was common and regular. Puzzled, thinking this was a "one-off" I even was happy to have Johannes as a Co-Author. See here https://github.com/kak-lsp/kak-lsp/pull/557#issuecomment-969280030 . Yet even that request was not accepted and the commit was finally merged as https://github.com/kak-lsp/kak-lsp/commit/369ce314ce4f1ff269f13c2616a4dfbcbeb733d0 with the commentary changed and comment "
Patch-tweaked-by: Johannes Altmanninger <aclopte@gmail.com>". I certainly don't think that Johannes, tweaked my patch. There were a significant amount of changes made.

This is a pattern that is being repeated. All this is being done in the name of "clean commit history". What use is "clean commit history" if your contributors are unhappy? I would assume that "clean commit history" would be secondary.

These are my requests:
- My commits are not `--amend` without discussing with me. Which means that in the default case, please leave them as is. I am happy for your to combine them together also or ask me to combine them at any point in time.
- **If the maintainer or anybody wants to improve or fix my code, they are free to do them on _separate commits_**. If I take too long in incorporating the review comments, they can discuss with me on next steps. Failing prompt resolution, I am happy for my commits to be used as a base to incorporate additional features, and fixes in _separate_ commits. My objection is _only_ about `--amend`ing my commit, changing my commit commentary and most importantly the code that I wrote.
- Avoid giving fiat-acomplis to contributors. Discuss what you want to do before you do it. If you want to merge, please go ahead and merge. But if you are merging with `--amend`-ing the commits, at least ask the original author first?
  
I don't think I'm asking very much. 

It is not my intention to slow down the pace of this project. All I'm saying is don't `--amend` my code without discussing. Do whatever you want on top of my commits in separate commits though. I'm cool with that.

krobelus: > A clean history on merge does make sense, but, in my opinion, it should either be discussed in comments until all parties are on the same page or be postponed until the PR author considers their work complete and ready for merge/hand-off.

Agreed. As I said earlier, I'm happy to discuss every change I'd like to make.

We can still do that for this PR (dropping #570), I'm not sure if @sidkshatriya is up for that?

krobelus: > Failing prompt resolution, I am happy for my commits to be used as a base to incorporate additional features, and fixes in _separate_ commits.

That doesn't feel right to me. If there is something wrong in a commit (and the author is no longer available) I'd rather fix the commit than creating a fixup commit.

sidkshatriya: @krobelus I feel we have reached an impasse here. I am no longer motivated to spent my time and energy on this project because you refuse to simply use my commit but insist on on `--amend`ing it. 

I actually was intending to work on the review points but you took the initiative on #570 and incorporated the review comments yourself into the work I originally did. I comment you for taking the initiative. But to say that you refuse to build upon this commit unless you have the ability to `--amend` it is absolutely breathtaking to me.

**If there is something wrong in a commit, you have another commit to fix it. When a merge happens, your fixes will be merged along with it when the branch is merged. You are aware of this.**

This commit compiles, is functional. The feature works. Yes, it can do with improvement and changes in approach. It has flaws but I'm not aware of any fatal ones. But to say that it must not exist as _one_ of the many commits on a feature branch that you have full ability to add on before a merge is quite surprising to me. 

The experience has been a net negative for me. At this point I can continue debating with you or cut my losses. I suggest you do what you feel like without my approval or consent.

Or you can take a small pause. Don't take my word work it. Please consult with other contributors  (e.g. @ul  -- the originator of the project or @sclu1034 who happened to add an opinion) and see what they suggest. You can still chose to take your own path in the end!

sidkshatriya: P.S. @krobelus In the interest of resolving this issue (because I just don't have the time to spend more time on a contribution I purely did out of love of a project)...

**Proposal**
Please make the minimum changes you deem acceptable to this commit to be considered merge-able. You have already done the work in #570 where you have addressed many of your own review comments. 

You can even commit as the primary author of this commit. I'm happy to be credited as the Co-author of this feature in the commit message. And then we can call it a day. kak-lsp becomes better and we can move on.

This would address your concern about having a commit on the master branch which does not have any flaws. Can we do this?

krobelus: Thanks for the contribution ;)

PR #568, opened 2021-11-21
add racket lsp config

shenlebantongying:
It works as intended.

image

The `info.rkt` file is equivalent of `
CMakeLists.txt` or `Cargo.toml`. It indicates the root of projects.
https://docs.racket-lang.org/raco/info_rkt.html

### DIFF
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -240,6 +240,12 @@ roots = ["DESCRIPTION", ".git", ".hg"]
 command = "R"
 args = ["--slave", "-e", "languageserver::run()"]
 
+[language.racket]
+filetypes = ["racket"]
+roots = ["info.rkt"]
+command = "racket"
+args = ["-l", "racket-langserver"]
+
 [language.reason]
 filetypes = ["reason"]
 roots = ["package.json", "Makefile", ".git", ".hg"]


krobelus:
Sweet thanks!
I wonder if completion of identifiers with weird characters like # and * also works, this could depend on how smart the server is.

shenlebantongying:
Hmmm, is this a recent bug or something?

I cannot autocomplete `->` like [`string->list`](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string-~3elist%29%29)

the lsp can find the keyword, but if i type beyond the `-` or `>`, the choices are gone (reset from beginning).

image

krobelus:
The option `lsp_completion_fragment_start` tells kak-lsp at which position to request completions.
For example if you type `foo#bar`, then by default kak-lsp will ask for completions at the position of `b`, because by default only word characters are included. If that's not enough for your server, you can use something like

    set-option global lsp_completion_fragment_start %{execute-keys <esc><a-h>s\$?[\w>#*+-]+.\z<ret>}

to include other characters besides the `\w`. (probably theres a better defintion using `[^]`
Maybe we can improve the default.

shenlebantongying:
I think the problem is that we don't reuse the `extra_word_chars`, like lisp:

The special chars are predefined by the rc files

https://github.com/mawww/kakoune/blob/c507863a00289fe4b210ecc4fef5554de4c87bf4/rc/filetype/lisp.kak#L19

shenlebantongying:
Maybe this is a problem on Kak's side, rather than here?

krobelus:
The fact that `\w` ignores extra_word_chars is intentional. Kakoune's `:doc regex` states

> * \w which matches word characters A-Z, a-z, 0-9 and underscore (ignoring the extra_word_chars option).

Making the behavior of a regex depend on such an option would be bad because it makes search less predictable.

But you are right, we should probably honor that option here.
This should work (I have only looked at the logs, and not tested with a real language server).

```diff
diff --git a/rc/lsp.kak b/rc/lsp.kak
index 41d8f58..3aa8df6 100644
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -37,7 +37,21 @@ declare-option -docstring "Completion request is sent only when this expression
 # Kakoune requires completions to point fragment start rather than cursor position.
 # This variable provides a way to customise how fragment start is detected.
 # By default, it tracks back to the first punctuation or whitespace.
-declare-option -docstring "Select from cursor to the start of the term being completed" str lsp_completion_fragment_start %{execute-keys <esc><a-h>s\$?\w+.\z<ret>}
+declare-option -docstring "Select from cursor to the start of the term being completed" str lsp_completion_fragment_start %{execute-keys "<esc><a-h>s\$?[\w%opt{lsp_extra_word_chars}]+.\z<ret>"}
+declare-option str lsp_extra_word_chars
+hook global WinSetOption extra_word_chars=.* %{
+    set-option window lsp_extra_word_chars %sh{
+        eval set -- $kak_quoted_opt_extra_word_chars
+        for char; do
+            case "$char" in
+                (-) printf '\\-' ;;
+                (\) printf '\\\\' ;;
+                (]) printf '\\]' ;;
+                (*) printf %s "$char" ;;
+            esac
+        done
+    }
+}
 # If hover in insert mode is enabled then request is made only when this expression doesn't fail and
 # for position at which it moves cursor; by default, it ensures that cursor is after opening parens
 # and then moves cursor to opening parens to request hover info for current function; note that it
```

Instead of applying the diff you can also just add the
`declare-option str lsp_extra_word_chars`
and
`hook global WinSetOption extra_word_chars=.* %{ ... }`
bits to your configuration, plus of course

```
set-option global lsp_completion_fragment_start %{execute-keys "<esc><a-h>s\$?[\w%opt{lsp_extra_word_chars}]+.\z<ret>"}
```

krobelus:
my example was missing a backslash.
Anyway, fixed the extra_word_chars option in master now with d9a8eb7 (Honor extra_word_chars option when finding start of token to complete, 2021-11-22)
someone should add the racket filetype to Kakoune ;)
We can reuse existing lisp highlighters, like `rc/filetype/clojure.kak` does

shenlebantongying:
Thanks for your gorgeous help!

However, there is a minor problem, the completion will be crippled if the `extra_word_chars` contains those chars (they are crippled in different way :sweat_drops: :)

- `<` The completion candidates will stay the same for whatever.
- '=' The `lsp_extra_word_chars` will be updated, but will not respect those extra chars altogether.

`clojure.kak` contains both of them:

https://github.com/mawww/kakoune/blob/c507863a00289fe4b210ecc4fef5554de4c87bf4/rc/filetype/clojure.kak#L187

krobelus:
I've fixed the problem with < in 92db557d32dd66cb182f560a1ff74753b79fa309.
With that fix, I can't find a problem with =, are there any other chars necessary to reproduce?

shenlebantongying:
I am so sorry, it seems I cannot reproduce it too. I must change something else.......

krobelus:
maybe it happened because it also had `<`

PR #569, opened 2021-11-23
Order of checking exceeds() matters if we are searching forward or backward

sidkshatriya:
Currently _previous_ symbol goto/hover is not working properly for documents with nested children.

E.g. Open up kakoune source `src/input_handler.cc` line 227 (Use kakoune `08d2a8ba63f9f0fbabf6a6221531afc91698fc20`). (I assume that you are using clangd) 

**Actual**
Hover of the previous symbol i.e. `<lsp> [` will show you `Namespace Kakoune`

**Expected**
It should be `Method on_disabled`

This PR fixes this issue.

### DIFF
--- a/src/language_features/document_symbol.rs
+++ b/src/language_features/document_symbol.rs
@@ -317,13 +317,13 @@ fn next_or_prev_symbol_details<T: Symbol<T> + 'static>(
     for symbol in it {
         let kind = symbol.kind();
         let mut filename_path = PathBuf::default();
-        let filename = symbol_filename(meta, &symbol, &mut filename_path);
+        let filename = symbol_filename(meta, &symbol, &mut filename_path).to_string();
 
         let mut symbol_position = symbol.range().start;
         if TypeId::of::<T>() == TypeId::of::<SymbolInformation>() {
             symbol_position = find_identifier_in_file(
                 ctx,
-                filename,
+                &filename,
                 symbol_position,
                 unadorned_name(ctx, symbol.name()),
             )
@@ -332,24 +332,42 @@ fn next_or_prev_symbol_details<T: Symbol<T> + 'static>(
         let symbol_position =
             get_kakoune_position_with_fallback(&meta.buffile, symbol_position, ctx);
 
-        if exceeds(
-            symbol_position,
-            kind,
-            params.position,
-            params.search_next,
-            &params.symbol_kind,
-        ) {
-            return Some((
-                filename.to_string(),
+        let symbol_name = symbol.name().to_owned();
+
+        // Order of checking the parent symbol for exceeds() matters.
+        // If we're searching forward then check if the parent symbol exceeds() the current
+        // symbol first. If we're searching backwards then check if any of children symbols
+        // exceeds() the current symbol first
+
+        // Deals with the case we're searching forwards
+        if params.search_next
+            && exceeds(
                 symbol_position,
-                symbol.name().to_owned(),
                 kind,
-            ));
+                params.position,
+                params.search_next,
+                &params.symbol_kind,
+            )
+        {
+            return Some((filename, symbol_position, symbol_name, kind));
         }
 
         let from_children = next_or_prev_symbol_details(symbol.children(), params, meta, ctx);
         if from_children.is_some() {
             return from_children;
+        }
+
+        // Deals with the case we're searching backwards
+        if !params.search_next
+            && exceeds(
+                symbol_position,
+                kind,
+                params.position,
+                params.search_next,
+                &params.symbol_kind,
+            )
+        {
+            return Some((filename, symbol_position, symbol_name, kind));
         }
     }
 


src/language_features/document_symbol.rs
```diff
@@ -317,13 +317,13 @@ fn next_or_prev_symbol_details<T: Symbol<T> + 'static>(
     for symbol in it {
         let kind = symbol.kind();
```
krobelus, line 318: Great fix, thanks!
Since we use the start position as a sorting key, there's a natural asymmetry between searching forward and backward (I believe parents never have lower starting position than their children).

While the old behavior could in theory be useful, it was inconsistent.
This fix means we never skip any children which seems the best behavior (we don't wanna create something as confusing as GNU info ;).

I'll gladly merge as-is but I'll leave a few comments first.

You have a nice reproducer that explains the problem, it would be nice to put that in the commit message (not only on github)

src/language_features/document_symbol.rs
```diff
@@ -332,25 +332,43 @@ fn next_or_prev_symbol_details<T: Symbol<T> + 'static>(
...
-            &params.symbol_kind,
-        ) {
-            return Some((
-                filename.to_string(),
+        let symbol_name = symbol.name().to_owned();
+
+        // Order of checking the parent symbol for exceeds() matters.
+        // If we're searching forward then check if the parent symbol exceeds() the current
+        // symbol first. If we're searching backwards then check if any of children symbols
+        // exceeds() the current symbol first
```
krobelus, line 340: These three sentences are kind of redundant because they only repeat what the code is doing.
What would be much more interesting is *why* there is a difference between forward and backward search.

here is my attempt

```rust
// Assume that children always have a starting position higher than (or equal to) their
// parents.  This means that when searching for the node with the next-higher position
// (anywhere in the tree) we need to check the parent first.  Conversely, when looking
// for the node with the next-lower position, we need to check children first, because
// they have a higher position than their parents.
```

Maybe that's a bit too wordy. It's probably not necessary to spend too much on explaining that.
The key thing is the assumption about parent/child positions, everything else can be derived from it.

src/language_features/document_symbol.rs
```diff
@@ -332,25 +332,43 @@ fn next_or_prev_symbol_details<T: Symbol<T> + 'static>(
...
         }
 
         let from_children = next_or_prev_symbol_details(symbol.children(), params, meta, ctx);
         if from_children.is_some() {
             return from_children;
         }
+
+        // Deals with the case we're searching backwards
+        if !params.search_next
+            && exceeds(
```
krobelus, line 362: you now duplicate the lengthy call to exceeds().
My first idea was to compute it in one place and share that.

```rust
let beyond_cursor = if exceeds(...) {
    Some((filename, symbol_position, symbol_name, kind))
} else {
    None
};

if params.search_next && beyond_cursor.is_some() {
    return beyond_cursor;
}

if let Some(child_beyond_cursor) =
    next_or_prev_symbol_details(symbol.children(), params, symbol_kinds, meta, ctx)
{
    return Some(child_beyond_cursor);
}

if !params.search_next && beyond_cursor.is_some() {
    return beyond_cursor;
}
```

but actually, it seems better to just remove exceeds() altogether.
Beyond checking the symbol kind (which is not obvious from its name), it does exactly 

```rust
    if search_next {
        symbol_position > cursor_position
    } else {
        symbol_position < cursor_position
    }
```

which means it definitely no longer pulls its weight if the callers *already* branch on `search_next`.
So the whole thing should be simplified to something like

```rust
let want_symbol = symbol_kind.is_empty() || format!("{:?}", kind) == symbol_kind;

if params.search_next && want_symbol && symbol_position > cursor_position {
    return Some((filename, symbol_position, symbol_name, kind));
}

if let Some(from_children) =
    next_or_prev_symbol_details(symbol.children(), params, symbol_kinds, meta, ctx)
{
    return Some(from_children);
}

if params.search_next && want_symbol && symbol_position < cursor_position {
    return Some((filename, symbol_position, symbol_name, kind));
}
```

(having removed this overzealous abstraction, I also see it as strange that we use < and > instead of something total like  < and >= but I haven't found problems)

src/language_features/document_symbol.rs
sidkshatriya: @krobelus 
> You have a nice reproducer that explains the problem, it would be nice to put that in the commit message (not only on github)

Done! force pushed commit with reproducer in commit message

> I'll gladly merge as-is but I'll leave a few comments first.

Yes, I would greatly appreciate that. Yes definitely improve the code in subsequent commits as you see fit -- you've already done the ground work in your review comments...

>  also see it as strange that we use < and > instead of something total like < and >= but I haven't found problems)

The omission of the `=` in `>=` is deliberate. IIRC, this was to ensure that we went forward/backward and didn't get "stuck". Basically next/previous always yielded next/previous instead of the same symbol.

src/language_features/document_symbol.rs
sidkshatriya: Removal of exceeds sound great!

PR #570, opened 2021-11-28
Show hover information in a scratch buffer v2

krobelus:
This is built on top of https://github.com/kak-lsp/kak-lsp/pull/564,
with review comments address, and some extra features.

The most important behavior change is that the new binding (lsp H)
by default shows the `*hover*` buffer in the current client.
I'm not super sure about that yet, but it seems like the smallest
common denominator.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -130,7 +130,8 @@ Either way you get:
 
 * completions
 * `lsp-definition` command to go to definition, mapped to `gd` by default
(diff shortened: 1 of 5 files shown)


krobelus:
I elaborated most of my review changes in 20f28ca and squashed & cleaned up review artifacts in 23d2c6b

gnull:
If this patch is still on the table, I'd be interested in testing it. How do I do that? And what's the latest version?

PR #571, opened 2021-11-30
Allow whitespace in gutter signs

0phoff:
By surrounding the different line-specs with single quotes, we enable the use of whitespace for the `lsp_diagnostic_line_error_sign`, `lsp_diagnostic_line_hint_sign`, `lsp_diagnostic_line_info_sign`, `lsp_diagnostic_line_warning_sign` options.

This is necessary to allow for some breathing room between the lsp-error and other gutters (eg. line-numbers).
As an example use-case: I use nerd-font icons as my symbols, but they are wider than an actual character and thus I need to add a whitespace character afterwards. Otherwise the icon overlaps the line-numbers

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -52,7 +52,7 @@ pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
         .iter()
         .map(|x| {
             format!(
-                "{}|{}",
+                "'{}|{}'",
                 x.range.start.line + 1,
                 match x.severity {
                     Some(DiagnosticSeverity::ERROR) => {
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -151,7 +151,7 @@ pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
          set buffer lsp_diagnostic_info_count {}; \
          set buffer lsp_diagnostic_warning_count {}; \
          set buffer lsp_errors {} {}; \
-         eval \"set buffer lsp_error_lines {} {} '0| '\"; \
+         eval \"set buffer lsp_error_lines {} {} '0|%opt[lsp_diagnostic_line_error_sign]'\"; \
          set buffer lsp_diagnostics {} {}",
         error_count,
         hint_count,
--- a/test/clangd-diagnostic-gutter.sh
+++ b/test/clangd-diagnostic-gutter.sh
@@ -20,3 +20,9 @@ test_tmux_kak_start main.c
 test_tmux capture-pane -p | sed 2q
 # CHECK: {{W }}void main(int argc, char** argv) {}
 # CHECK: {{ X}}syntax error
+
+test_tmux send-keys %:comment-line Enter
+test_sleep
+test_tmux capture-pane -p | sed 2q
+# CHECK: {{  }}// void main(int argc, char** argv) {}
+# CHECK: {{  }}// syntax error
--- a/test/clangd-diagnostic-gutter.sh
+++ b/test/clangd-diagnostic-gutter.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+
+# REQUIRES: command -v clangd
+
+. test/lib.sh
+
+cat >> .config/kak/kakrc << EOF
+set-option global lsp_diagnostic_line_error_sign   ' X'
+set-option global lsp_diagnostic_line_hint_sign    '¿ '
+set-option global lsp_diagnostic_line_info_sign    '¡ '
+set-option global lsp_diagnostic_line_warning_sign 'W '
+EOF
+
+cat > main.c << EOF
+void main(int argc, char** argv) {}
+syntax error
+EOF
+
+test_tmux_kak_start main.c
+test_tmux capture-pane -p | sed 2q
+# CHECK: {{W }}void main(int argc, char** argv) {}
+# CHECK: {{ X}}syntax error


0phoff:
Note to the maintainers, I am not sure whether this needs some unit testing, but I don't understand how your tests work and thus might need some guidance here... :innocent:

src/diagnostics.rs
```diff
@@ -28,7 +28,7 @@ pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
         .rev()
         .map(|x| {
             format!(
-                "{}|{}",
+                "'{}|{}'",
```
krobelus, line 31: Yes would be great to have a test because this is non-obvious (why would we need quoting if we always get something like `1.1,2.2|DiagnosticError` here?)

Something like

```sh
#!/bin/sh

# REQUIRES: command -v clangd

. test/lib.sh

cat >> .config/kak/kakrc << EOF
set-option global lsp_diagnostic_line_error_sign ' '
EOF

cat > main.c << EOF
syntax error
EOF

test_tmux_kak_start main.c
test_tmux capture-pane -p | grep syntax.error
# CHECK: {{ }}syntax error
# TODO add some other line-flags highlighter, so we can see the fix
```

Then run that file (`test/clangd-diagnostic-gutter.sh`) to see the output and
`test/run test/clangd-diagnostic-gutter.sh` to check the output.

The tests are a bit underdocumented but the nice thing is that you can add `test_tmux attach` and see what's going on

0phoff: Oops, because I made a stupid mistake and edited the wrong brackets.
I first made this fix in my local `.config/kak/plugins/kak-lsp`, and then quickly forked and applied the same change to submit a PR.
Apparently, I edited the wrong line though (changed 31 instead of 55).

I will add the unit test later this week!

0phoff:
I added a unit test to check whether whitespaces are working correctly for the gutter signs (checking whitespace before and after).
I then checked the unit test before and after making my modification to the code-base, to make sure my fix is indeed working properly!

### Before Change
```
 void main(int argc, char** argv) {}
 syntax error
~
~
~
~
                                           main.c 1:1  1 sel - client0@[session]
```

<details>
<summary>test error output</summary>

```
Failure in test/clangd-diagnostic-gutter.sh:

  The CHECK on line 19 wants:
    {{W  }}void main(int argc, char** argv) {}

  which failed to match line stdout:1:
     void main(int argc, char** argv) {}

  Context:
     void main(int argc, char** argv) {} <= does not match CHECK '{{W  }}void main(int argc, char** argv) {}' on line 19
     syntax error <= does not match CHECK '{{ X }}syntax error' on line 20
    ~
    ~
    ~

  when running command:
    /bin/sh test/clangd-diagnostic-gutter.sh
```

</details>


### After Change
```
W  void main(int argc, char** argv) {}
 X syntax error
~
~
~
~
                                           main.c 1:1  1 sel - client0@[session]
```

No output from `test/run`, which means it was successful.

src/diagnostics.rs
0phoff: I added quotes to the correct line now and amended my commit.

0phoff:
@krobelus, using my branch, I noticed some annoying behaviour which might need fixing before merging this in.

When there are no errors, you set `lsp_error_lines` to `0 '0| '` (or at least that is what I think).
However, as my lsp gutter symbols take more width, this means the gutter constantly shifts from a character width of 1 when there are no errors and a width of n characters when there are errors
This is basically the same issue you had, when you enabled auto hiding of the lsp gutter.

We can fix this by computing the biggest number of characters from `lsp_diagnostic_line_error_sign` etal.
I will look into this when I have some free time :)

src/diagnostics.rs
krobelus: > I first made this fix in my local `.config/kak/plugins/kak-lsp`, and then quickly forked and applied the same change to submit a PR.

BTW I think you could have directly used that repo, you just need to set the remote to `git@github.com:user/repo` I usually use `hub fork`, which does that automatically I think.

Thanks for the test, looks really nice!

src/diagnostics.rs
```diff
@@ -52,7 +52,7 @@ pub fn publish_diagnostics(params: Params, ctx: &mut Context) {
         .iter()
         .map(|x| {
             format!(
-                "{}|{}",
+                "'{}|{}'",
```
krobelus, line 55: > When there are no errors, you set `lsp_error_lines` to `0 '0| '`

Good catch; I've added another commit that simply uses `%opt{lsp_diagnostic_line_error_sign}` as the default width.

I think that's good enough for now because users probably want to configure all 4 symbols to the same width, to avoid the same glitch.

It's a bit hacky, but so is this quoting -- it means that `'` needs to be doubled if users want that in the gutter.
Neither is a problem, so I'd merge like this and perhaps in future implement this on the Rust side.

> I use nerd-font icons as my symbols, but they are wider than an actual character and thus I need to add a whitespace character afterwards. Otherwise the icon overlaps the line-numbers

Sounds like Kakoune (which uses the system's wcwidth()) disagrees on the width of the character with your terminal.
This is typical of macOS systems because those ship old versions of wcwidth(), and as a result terminals opt to bundle newer ones..
We could also do that -- using something like https://github.com/ridiculousfish/widecharwidth -- but that would probably make things worse.

test/clangd-diagnostic-gutter.sh
```diff
@@ -0,0 +1,28 @@
...
+syntax error
+EOF
+
+test_tmux_kak_start main.c
+test_tmux capture-pane -p | sed 2q
+# CHECK: {{W }}void main(int argc, char** argv) {}
+# CHECK: {{ X}}syntax error
+
+test_tmux send-keys %:comment-line Enter
+test_sleep
```
krobelus, line 25: I dropped this line, to make the test more resilient to unrelated changes (there's a double space in our modeline that I'd like to get rid of eventually).
Also "unit" test is probably the wrong term, it's more of an end-to-end system test because it tests interaction, plus it even depends on external components (clangd) outside our control (I plan to replace it with a fake language servers eventually).

PR #573, opened 2021-12-05
Add support for custom texlab commands

PJungkamp:
New commands:
 - `texlab-build` - request `textDocument/build` from texlab
 - `texlab-forward-search` - request `textDocument/forwardSearch` from texlab

I implemented the commands based on the language_features modules of
rust-analyzer and ccls.

I can't confirm whether these changes are correct, since the master branch of kak-lsp currently crashes for me with texlab.

I need to file an issue for that and investigate.

I just configured kak and kak-lsp today and well in love with kakoune's simple, yet powerful featureset.
Therefore I am neither particulary experienced with kakoune, nor with this codebase.
Any help is appreciated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -423,6 +423,7 @@ dependencies = [
  "serde",
  "serde_derive",
  "serde_json",
+ "serde_repr",
  "slog",
  "slog-scope",
  "sloggers",
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -1127,6 +1127,45 @@ method    = "rust-analyzer/inlayHints"
 ' "${kak_session}" "${kak_buffile}" "${kak_opt_filetype}" "${kak_timestamp}" | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }
 
+# texlab extensions
+
+define-command texlab-forward-search -docstring "Request SyncTeX Forward Search for current line from the texlab language server
+
+This will focus the current line in your PDF viewer, starting one if necessary.
+To configure the PDF viewer, use texlab's options 'forwardSearch.executable' and 'forwardSearch.args'." %{
+    lsp-did-change-and-then texlab-forward-search-request
+}
+
+define-command -hidden texlab-forward-search-request %{
+    nop %sh{ (printf '
+session   = "%s"
+client    = "%s"
+buffile   = "%s"
+filetype  = "%s"
+version   = %d
+method    = "textDocument/forwardSearch"
+[params.position]
+line      = %d
+column    = %d
+' "${kak_session}" "${kak_client}" "${kak_buffile}" "${kak_opt_filetype}" "${kak_timestamp}" ${kak_cursor_line} ${kak_cursor_column} | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
+}
+
+define-command texlab-build -docstring "Ask the texlab language server to build the LaTeX document" %{
+    lsp-did-change-and-then texlab-build-request
+}
+
+define-command -hidden texlab-build-request %{
+    nop %sh{ (printf '
+session   = "%s"
+client    = "%s"
+buffile   = "%s"
+filetype  = "%s"
+version   = %d
+method    = "textDocument/build"
+[params]
+' "${kak_session}" "${kak_client}" "${kak_buffile}" "${kak_opt_filetype}" "${kak_timestamp}" | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
+}
+
 # semantic tokens
 
 define-command lsp-semantic-tokens -docstring "lsp-semantic-tokens: Request semantic tokens" %{
--- a/src/language_features/texlab.rs
+++ b/src/language_features/texlab.rs
@@ -1,0 +1,107 @@
+use crate::context::Context;
+use crate::position::get_lsp_position;
+use crate::types::{EditorMeta, EditorParams};
+use crate::PositionParams;
+use lsp_types::request::Request;
+use lsp_types::TextDocumentIdentifier;
+use lsp_types::TextDocumentPositionParams;
+use lsp_types::Url;
+use serde::{Deserialize, Serialize};
+use serde_repr::{Deserialize_repr, Serialize_repr};
+use std::fmt;
+
+pub enum ForwardSearch {}
+
+impl Request for ForwardSearch {
+    type Params = TextDocumentPositionParams;
+    type Result = ForwardSearchResult;
+    const METHOD: &'static str = "textDocument/forwardSearch";
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ForwardSearchResult {
+    status: ForwardSearchStatus,
+}
+
+#[derive(Serialize_repr, Deserialize_repr, Debug)]
+#[repr(i32)]
+pub enum ForwardSearchStatus {
+    Success = 0,
+    Error = 1,
+    Failure = 2,
+    Unconfigured = 3,
+}
+
+impl fmt::Display for ForwardSearchResult {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Forward Search {:?} (texlab)", self.status)
+    }
+}
+
+pub fn forward_search(meta: EditorMeta, params: EditorParams, ctx: &mut Context) {
+    let params = PositionParams::deserialize(params).unwrap();
+    let req_params = TextDocumentPositionParams {
+        text_document: TextDocumentIdentifier {
+            uri: Url::from_file_path(&meta.buffile).unwrap(),
+        },
+        position: get_lsp_position(&meta.buffile, &params.position, ctx).unwrap(),
+    };
+    ctx.call::<ForwardSearch, _>(meta, req_params, move |ctx, meta, response| {
+        forward_search_response(meta, response, ctx)
+    });
+}
+
+pub fn forward_search_response(meta: EditorMeta, result: ForwardSearchResult, ctx: &mut Context) {
+    let command = format!("echo {}", result);
+    ctx.exec(meta, command);
+}
+
+pub enum Build {}
+
+impl Request for Build {
+    type Params = BuildTextDocumentParams;
+    type Result = BuildResult;
+    const METHOD: &'static str = "textDocument/build";
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct BuildTextDocumentParams {
+    text_document: TextDocumentIdentifier,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct BuildResult {
+    status: BuildStatus,
+}
+
+#[derive(Serialize_repr, Deserialize_repr, Debug)]
+#[repr(i32)]
+pub enum BuildStatus {
+    Success = 0,
+    Error = 1,
+    Failure = 2,
+    Cancelled = 3,
+}
+
+impl fmt::Display for BuildResult {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Build {:?} (texlab)", self.status)
+    }
+}
+
+pub fn build(meta: EditorMeta, _params: EditorParams, ctx: &mut Context) {
+    let req_params = BuildTextDocumentParams {
+        text_document: TextDocumentIdentifier {
+            uri: Url::from_file_path(&meta.buffile).unwrap(),
+        },
+    };
+    ctx.call::<Build, _>(meta, req_params, move |ctx, meta, response| {
+        build_response(meta, response, ctx)
+    });
+}
+
+pub fn build_response(meta: EditorMeta, result: BuildResult, ctx: &mut Context) {
+    let command = format!("echo {}", result);
+    ctx.exec(meta, command);
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ ropey = "1.2.0"
 serde = "1.0.112"
 serde_derive = "1.0.112"
 serde_json = "1.0.55"
+serde_repr = "0.1.7"
 slog = { version = "2.5.2", features = ["release_max_level_debug"] }
 slog-scope = "4.3.0"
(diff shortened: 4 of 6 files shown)


src/language_features/texlab.rs
```diff
@@ -0,0 +1,95 @@
...
+            uri: Url::from_file_path(&meta.buffile).unwrap(),
+        },
+    };
+    ctx.call::<Build, _>(meta, req_params, move |ctx, meta, response| {
+        build_response(meta, response, ctx)
+    });
+}
+
+pub fn build_response(meta: EditorMeta, result: BuildResult, ctx: &mut Context) {
+    let command = format!("echo {:?}", result);
```
PJungkamp, line 93: This command is executed when texlab indicates a build result.

Is the echo of the Build Statue here correct?

src/language_features/texlab.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct BuildResult {
+    status: BuildStatus,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub enum BuildStatus {
```
PJungkamp, line 74: This is plain wrong, this should derive `Serialize_repr`  and `Deserialize_repr` with `#[repr(i32)]`.

PJungkamp:
The texlab commands seem to be working. But I'd like to get feedback in kakoune (maybe in the modeline) about the return status of texlab-build / texlab-forward-search. See the review above.
Seems like the echo I added in `build_response` does not suffice.

krobelus:
> I can't confirm whether these changes are correct, since the master branch of kak-lsp currently crashes for me with texlab.

uh-oh, let's fix that first. Do you have a reproducer or a log?

src/language_features/texlab.rs
krobelus: The problem with `ctx.exec(meta, "echo something")` is that it runs without a (Kakoune) client in the context.
It's equivalent to running `echo echo something | kak -p my-session`.

Luckily, since we know the Kakoune client who made the request. So something like (untested)

```rust
let command = format!("evaluate-commands -client {} %{ echo the-status }", meta.client.unwrap());
ctx.exec(meta, command);
```

should work. I also like to use `echo -debug` for debugging (that one does not need a client context).

src/language_features/texlab.rs
```diff
@@ -0,0 +1,96 @@
+use crate::context::Context;
+use crate::position::get_lsp_position;
+use crate::types::{EditorMeta, EditorParams};
+use crate::PositionParams;
+use lsp_types::request::Request;
+use lsp_types::TextDocumentIdentifier;
+use lsp_types::TextDocumentPositionParams;
```
krobelus, line 7: please squash such commits, see https://git-rebase.io/ for a guide.
I think other readers don't care about your mistakes ;) and I don't mind force-pushes on ephemeral branches

Cargo.lock
```diff
@@ -423,6 +423,7 @@ dependencies = [
  "serde",
  "serde_derive",
  "serde_json",
+ "serde_repr",
```
krobelus, line 426: I don't think we are supposed to edit this file by hand, `cargo build` will take care of that

PJungkamp: That change was made by cargo build. But since I initially didn't push the updated `Cargo.lock` the Pipeline failed on `cargo build --locked`, since the Lockfile was out of date.

src/language_features/texlab.rs
PJungkamp: I will sqash all commits once this is satisfactory. This isn't a huge change, so I think a single final commit should suffice.

PJungkamp:
> > I can't confirm whether these changes are correct, since the master branch of kak-lsp currently crashes for me with texlab.
> 
> uh-oh, let's fix that first. Do you have a reproducer or a log?

Had my path messed up... (fish_add_path acts on all shells so I had the recent kak-lsp with the old lsp.kak running, which then failed due to `have_kakoune_feature_filtertext` missing)
My bad.

PJungkamp:
As far as I can tell this is working beautifully. I'm now ready to ditch my `nvim` :D.

src/language_features/texlab.rs
PJungkamp: The Problem was that I did not pass the client to the meta when calling build in `lsp.kak`. `kak-lsp` uses `eval -client <client> <cmd>` automatically when invoking `ctx.exec(meta,command)` when the meta specifies a client, so a change in lsp.kak was sufficient.

krobelus:
I didn't try it yet but this looks really great!

I love the `make` command that ships with Kakoune because it let's me jump to error locations (if the compiler has a gcc-like `file:line: error` output). Sadly getting this location info from latex compilers seems a bit harder (or maybe not?).

BTW I once wrote a command for synctex support (both directions) at https://gist.github.com/krobelus/b064c3cfae55b635908a59ac9255ace1
I think it worked fine but I never really used it. I guess that texlab-forward-search obsoletes some of that.

> Had my path messed up... (fish_add_path acts on all shells so I had the recent kak-lsp with the old lsp.kak running, which then failed due to `have_kakoune_feature_filtertext` missing)

Right, this is a moderately-common user error. I've been meaning to add a dedicated error message for such a version mismatch.
I assume you've been using plug.kak? Otherwise you'd probably be using `eval %sh{kak-lsp --kakoune -s $kak_session}` (which sources the equivalent of rc/lsp.kak) then there would never be a mismatch.

(And yeah, by default fish_add_path sets a universal $fish_user_paths, so it even affects running shells which is both good and bad)

> As far as I can tell this is working beautifully. I'm now ready to ditch my `nvim` :D.

Hehe nice. I haven't looked back. Some Kakoune features are a bit bare-bones but everything feels much simpler, especially scripting.
I recommend reading Kakoune's `doc/` and some of `rc/`.

src/language_features/texlab.rs
krobelus: Right, it's always a good idea to check the `*debug*` buffer for error messages

src/language_features/texlab.rs
```diff
@@ -0,0 +1,109 @@
...
+pub enum ForwardSearch {}
+
+impl Request for ForwardSearch {
+    type Params = TextDocumentPositionParams;
+    type Result = ForwardSearchResult;
+    const METHOD: &'static str = "textDocument/forwardSearch";
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
```
krobelus, line 22: Nitpick :microscope: I think we usually don't use the camelCase annotation, unless there is a field that needs it.
So only for BuildTextDocumentParams, which has `text_document`, but not others.

krobelus:
I've tried this and it works like a charm. I've also made some smaller documentation changes, maybe this is better? I emphasized that texlab is a language server, for users who might not know that.

Should we add a working default configuration? I'm thinking of

```diff
diff --git a/kak-lsp.toml b/kak-lsp.toml
index a8fa8b2..e2f4237 100644
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -191,6 +191,8 @@ settings_section = "texlab"
 [language.latex.settings.texlab]
 # See https://github.com/latex-lsp/texlab/blob/master/src/options.rs
 # bibtexFormatter = "texlab"
+forwardSearch.executable = "zathura"
+forwardSearch.args = ["--synctex-forward", "%l:1:%f", "%p"]
 
 [language.lua]
 filetypes = ["lua"]
```

krobelus:
I also looked into adding backwards synchronization. In zathura you can Control+Click to jump to the origin of a line.

This is an interesting problem, since we need to use `forwardSearch.args` to pass at least the Kakoune session name to zathura, otherwise it won't know how to talk to us.
We can make this happen with dynamic configuration options, like this:

```kak
define-command texlab-forward-search-with-backwards-sync \
    -docstring "Request SyncTeX Forward Search for current line in zathura

This will focus the current line in zathura.
Control+Click inside zathura will make the Kakoune client jump to the origin of that line." %{
    # TODO Avoid clobbering the global config.
    set-option global lsp_config "
    [language.latex.settings.texlab]
    forwardSearch.executable = 'zathura'
    forwardSearch.args = [
        '%%p',
        '--synctex-forward',
        '%%l:1:%%f',
        '--synctex-editor-command',
        ""sh -c 'echo evaluate-commands -verbatim -client %val{client} -- edit -- %%{input} %%{line} | kak -p %val{session}'"",
    ]
    "
    texlab-forward-search
}
```

BTW it took me a long time to figure out Control+Click. https://github.com/latex-lsp/texlab/blob/master/docs/previewing.md says it's Alt+Click. Is that wrong?

rc/lsp.kak
```diff
@@ -1127,6 +1127,42 @@ method    = "rust-analyzer/inlayHints"
 ' "${kak_session}" "${kak_buffile}" "${kak_opt_filetype}" "${kak_timestamp}" | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }
 
+# texlab extensions
+
+define-command texlab-forward-search -docstring "texlab-forward-search: Execute SyncTeX Forward Search for current line (texlab)" %{
```
krobelus, line 1132: I believe we have an unwritten convention that we use a `-docstring "texlab-forward-search:` synopsis only if there are any parameters. Wil change.

krobelus:
Merging now since I want to include this in today's release, but let me know any improvements. Thanks!

krobelus:
@cosimodpersia this looks like a nice feature if you're using texlab ;)

PJungkamp:
Sorry for not responding. My last few days were a little busy. I'm using kak-lsp with texlab now for writing my bachelors thesis.

I should have posted my configuration.... oops.
I'm going to share some of my experiences with kak-lsp and zathura here.

I'm also using forward and backward synchronization with zathura and zathuras shortcut was always <Ctrl+LMB>.

Your backwards synchronization command passed to zathura is *exactly* the one I'm using. 👍 

But I also wrap the forward synchronization into an `sh -c` which then forks and daemonizes zathura. Reason for that is that zathura tries to find an already open window with the required document. But if it does not find one, the searching zathura process opens a window and blocks the calling thread of texlab.
When you then exit kak, the texlab process will daemonize and make future sessions for the same file impossible, as long as you don't manually close zathura.
Since my zathura has been forked of it does not block texlab.

Another problem when a zathura window from a previous session exists is that the `--synctex-editor-command` is only applied on the first forward search so a subsequent kak session with diffent id can no longer bavkwards sync until zathura is manually closed.

I circumvented these problems by having one kak session name for each shell (not a `kak -d -s ...` daemon, just an environment variable and an alias).

A little request. Could you try calling `lsp-formatting-sync` from kak with texlab? I have no idea how it is suposed to work, but for me only the non-sync version actually works. I'm using kak and texlab now the 4th day and I am not to familiar with kakoune's hooks yet.

PJungkamp:
My `kak-lsp.toml` for texlab, after fixing aforementioned oddities:
```toml
[language.latex]
filetypes = ["latex"]
roots = [".git", ".hg",".latexmain"]
command = "texlab"
settings_section = "texlab"
[language.latex.settings.texlab]
latexindent.modifyLineBreaks = true
build.executable = "latexmk"
# compile with synctex, duh
build.args = ["-synctex=1","-interaction=nonstopmode","%f"]
# build could also be done by an Post Write Hook
build.onSave = true
# here using sh instead of zathura directly
forwardSearch.executable = "sh"
forwardSearch.args = [ "-c", """\
    zathura \
    --synctex-forward %l:1:%f \
    --synctex-editor-command 'sh -c "echo eval -try-client %opt{jumpclient} edit %{input} %{line} | kak -p $kak_session"' \
    %p & \
    """ ]
```
I'm using a setup with latexmk redirecting warnings and errors straight to my `toolsclient` while zathura always syncs the `jumpclient`. 
I had the backwards synchronization in the kakrc (like you posted) but wasn't quite satisfied with that.

Thank you for your awesome work! Adding those LSP commands was really straight forward. I actually did that the day I first tried kakoune, editing kak-lsp while still on neovim. ;)

krobelus:
Thanks for the config! I've been using Kakoune for years and I didn't realize that $kak_session was available, which means we can actually make this work in the TOML config.
I've tweaked it a bit, let me know if you think this is a usable default config: https://github.com/kak-lsp/kak-lsp/pull/577

> build.args = ["-synctex=1","-interaction=nonstopmode","%f"]

any reason for leaving out the "pdf"? Otherwise this is the default, so let's stick to that if we can

> A little request. Could you try calling `lsp-formatting-sync` from kak with texlab? I have no idea how it is suposed to work, but for me only the non-sync version actually works. I'm using kak and texlab now the 4th day and I am not to familiar with kakoune's hooks yet.

yeah that's just a bug in kak-lsp (exposed since 4ff7180 (Mimic VSCode server settings to support workspace/configuration, 2021-08-25))
because we use a fifo to implement the synchronicity, and texlab requests configuration on every sync (which is weird but okay I guess) and we write the config response to the fifo, so the actual formatting response has no reader, so it blocks

I briefly tried to fix it by not using the fifo for the config response, but it didn't work, I'd need to take a closer look.

(of course, for interactive editing, the async version is good enough)

> Another problem when a zathura window from a previous session exists is that the `--synctex-editor-command` is only applied on the first forward search so a subsequent kak session with diffent id can no longer bavkwards sync until zathura is manually closed.

Hmm I had expected that each Kakoune session should only interact with its own zathura window.
(I usually name my sessions after `basename $(git rev-parse --show-toplevel)`, so one session per project. Some people use one session for almost everything.)

But somehow the forward search works across sessions. It even works with two different sessions open at the same time.
Okay for now I guess.

> But if it does not find one, the searching zathura process opens a window and blocks the calling thread of texlab.

I removed the & and couldn't reproduce the problem. Let me know if I read this wrong.
This is what I did:
- use the config from #577
- run `kak myfile.tex`
- run forward search, it works
- quit kak (quitting the last window quits the session)
- observe that the texlab server also stops
- run `kak myfile.tex` again, and forward search works again (it finds the same zathura session)

so I'm not sure in which scenario we need backgrounding

krobelus:
> A little request. Could you try calling lsp-formatting-sync from kak with texlab? I have no idea how it is suposed to work, but for me only the non-sync version actually works. I'm using kak and texlab now the 4th day and I am not to familiar with kakoune's hooks yet.

Fixed now with a12831a (Cache workspace/configuration response instead of asking editor, 2022-01-01), thanks for reporting!

PR #574, opened 2021-12-06
Implement textDocument/selectionRange

### LINKED ISSUE #288: textDocument/selectionRange support
`ra_lsp_server` uses it: https://rust-analyzer.github.io/manual.html#extend-selection

krobelus:
This adds support for selecting interesting ranges around the cursors.
Usually the ranges correspond to AST nodes.  Add a lsp-selection-range
user mode no quickly navigate the ranges.

The VSCode UI is a bit different, it just needs Alt+Shift+{Left,Right},
whereas we need ":enter-user-mode lsp<ret>v" followed by j/k.

We also allow to navigate outside the user mode using
lsp-selection-range-select, but we don't yet have a way to
automatically request ranges whenever we need them (like VSCode).
We should try to do that, by simply requesting new ranges whenever the
cursor positions are outside the cone of the previously fetched ranges.

Backstory:
I started looking into this feature because I wanted to implement
Kakuone text objects for functions and type definitions (generalizing
lsp-next-symbol).  Usually, <a-a>{ and }} are good enough but there
are cases where they're not (like Python). It looks like this feature
won't help much because it doesn't allow to filter out uninteresting
ranges, but I guess it can be useful by itself.
For function/type defintion text objects we just need to go back to
our documentSymbol approach.

We might still extend this textDocument/selectionRange to text objects,
but it's probably not worth it, since it feels gimmicky.

Closes #288

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -142,6 +142,8 @@ Either way you get:
 * `lsp-references` command to find references to the symbol under the main cursor, mapped to `gr` by default
 ** for the previous five commands, the `\*goto*` buffer has filetype `lsp-goto`, so you can press `<ret>` on a line or use the `lsp-jump` command
 * `lsp-find-error` command to jump to the next or previous error in the current file
+- `lsp-selection-range` command to quickly select interesting ranges around selections.
+  - `lsp-selection-range-select` to navigate ranges fetched by `lsp-selection-range`.
 * `lsp-next-location` and `lsp-previous-location` to jump to the next or previous location listed in a buffer with the `lsp-goto` filetype. These also work for buffers `\*grep*`, `\*lint*` and `\*make*`
 * `lsp-highlight-references` command to highlight all references to the symbol under the main cursor in the current buffer with the `Reference` face (which is equal to the `MatchingChar` face by default)
 * `lsp-document-symbol` command to list the current buffer's symbols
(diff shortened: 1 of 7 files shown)


sidkshatriya:
This is nice feature. Thanks!

The lsp user mode `v` description is "select parent nodes". 

However when you press `v`, `j k b t` keys are "select inner node", "select outer nodes", "select innermost nodes", "select outermost nodes". To me this description sounds more intuitive.

My description of `v` would therefore be "select inner/outer nodes" instead of "select parent nodes". "select parent node" implies direction of navigation is in one direction only. And once you go to the parent node, which child node would you traverse to, if you could? The inner/outer terminology works better for me at least.

P.S. Also the behavior of "V" is not intuitive for me. It seems it can be used to recover your last selection that used `lsp v +  k|j|b|t`. With the binding of `V` it seems it somehow composes with `v` which is not what it does. Generally `V` confused me but I need to explore that more.

krobelus:
Yeah it shows nicely what LSP can do (it should also work well with multiple selections). Probably it's best if we tweak the UI to be like VSCode's.

Thanks, the "parent" was an oversight I missed that we can also select child nodes (e.g. nodes that are inside the initial selection). Fixed with 05bfe1921f49d99762591b6741e34ffa32fba1db

I added V because sometimes you need to ajdust the view (if the selections go outside the window). Then it's nice to be able to jump back to the user mode with the same node selection.
It should be fairly easy to define your own commands in a way that is less confusing than V.
But as I said, we should maybe request selection ranges automatically behind the scenes, so the user should only see commands for inner/outer.

Issue #580, opened 2021-12-20
Query LSP from shell


TeddyDD:
I'd like to be able to fetch information from LSP from shell.

Use case: using FZF for finding symbols etc.

Possible implementation - flag that enables printing output of `--request` to stdout instead of piping it to Kakoune

```sh
printf '
session   = "%s"
client    = "%s"
buffile   = "%s"
filetype  = "%s"
version   = %d
method    = "textDocument/signatureHelp"
[params.position]
line      = %d
column    = %d
' "${kak_session}" ... |
 eval "${kak_opt_lsp_cmd} --request-stdout | my-command"
```

sidkshatriya:
Interesting idea. But most of kak-lsp's output is in terms of kakoune commands that only kakoune understands.

For this to be useful, kak-lsp would need to simply give you the JSON that the downstream LSP gave kak-lsp and pass it to `my-command` unaltered. 

Would that work?

TeddyDD:
Sure, that would be even better!

krobelus:
Sounds promising. I believe this has been brought up on Discord before.
Today you can start language servers manually in stdio mode, and input JSON-RPC (bypassing kak-lsp).
So I guess the main advantages are
1. it reuses kak-lsp language server config
2. it reuses a language server (if already running)

the design looks reasonable. It's a bit weird to input TOML and output JSON, but we can extend that (also https://github.com/kislyuk/yq is our friend)

TeddyDD:
> It's a bit weird to input TOML and output JSON

JSON both as input and output would be preferable (since it's easy to handle JSON in shell - jq, gron, jo...) but for good start reusing `--request` format is fine as well.

krobelus:
I have implemented a quick draft in https://github.com/kak-lsp/kak-lsp/pull/584
it should be enough to create something usable, but we can definitely polish it.

PR #584, opened 2021-12-30
Add request option to write raw LSP response to fifo

krobelus:
This allows to write a script like the one below that prints the
response from the language server. We could hide the fifo creation
behind a command lien flag, but maybe that's not worth it.

Now kak-lsp will print to the given fifo (instead of passing the
response to the editor) when "write_response_to_fifo" is set to true.
This is probably not the final interface.

One thing to note is that positions passed via `[params]` are Kakoune
coordinates (1-based), but LSP responses are 0-based.

See #580

Here is the example script I used to test.

	#!/bin/sh
	kak_session=session # Only works if a server for this session is up.
	kak_client=client0
	kak_buffile=$PWD/src/main.rs
	kak_opt_filetype=rust

	tmp=$(mktemp -d)
	pipe=${tmp}/fifo
	mkfifo ${pipe}

	printf '
	session                = "%s"
	client                 = "%s"
	buffile                = "%s"
	filetype               = "%s"
	fifo                   = "%s"
	write_response_to_fifo = true
	method    = "textDocument/documentSymbol"
	[params]
	' ${kak_session} ${kak_client} "${kak_buffile}" "$kak_opt_filetype" ${pipe} |
	    kak-lsp -s ${kak_session} --request &

	cat ${pipe}
	rm -r ${tmp}


krobelus:
I've pushed a new version with a minor breaking change to the old script:
`version = ${kak_opt_lsp_timestamp}` is now mandatory again.

- patches 1 & 2 are pure refactorings
- patch 4 adds `lsp-connect` (any other name?) that reuses an existing lsp-* request function and passes the JSON response to a user-defined handler. Let me know if that works for your use case.

krobelus:
> lsp-connect lsp-diagnostics show-response-debug`

Right, I forgot to remove that from the completions. Maybe let's also make it an error if an unsupported call is used.

LSP Diagnostics are not supported yet, because they don't use the request->response model. Instead, the server sends them whenever they are ready (and we store them in `%opt{lsp_diagnostics}`. If someone needs the latest JSON with diagnostics, we could cached it.

<details>

Funnily enough, if we try it

```diff
diff --git a/rc/lsp.kak b/rc/lsp.kak
index 41c91dcf..895b033f 100644
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -762,6 +762,7 @@ buffile  = \"${kak_buffile}\"
 filetype = \"${kak_opt_filetype}\"
 version  = ${kak_timestamp}
 method   = \"textDocument/diagnostics\"
+${kak_opt_lsp_connect_fifo}
 [params]
 " | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }
```

we literally get this string instead of JSON

```
lsp-show-diagnostics '/home/johannes/git/kak-lsp/t' 'test.c:3:12: warning:Non-void function does not return a value
test.c:7:21: warning:Non-void function does not return a value'
```

</details>

I think in general, this JSON forwarding is a bit hacky, but it's very nice for experimenting.
I'm not sure which requests are really useful. `lsp-document-symbol` seems very nice (it allows us to implement virtual Kakoune textobjects for functions/classes) but the rest seems pretty meh at first glance. `lsp-workspace-symbol` needs a parameter, so it currently needs a hack with `def -override`.

kkga:
> I'm not sure which requests are really useful.

Yeah, not all of them are useful for external connections, I imagine. Although your python example for showing the current symbol is something I'd never come up with -- really smart.

My current use-case is quite simple -- I just want to use a single interface for "picking things": be it buffers, files or LSP lists. So my top choices would be:

- document symbols
- workspace symbols
- references
- diagnostics

krobelus:
> I just want to use a single interface for "picking things": be it buffers, files or LSP lists.

Makes a ton of sense. For buffers/files I use Kakoune's built-in prompt completions but for grep/make and LSP lists I currently use scratch buffers, using a [derivative of grep-next-match](https://discuss.kakoune.com/t/single-command-for-grep-next-match-in-similar-buffers-lsp-make-find/1215)
(more recent version somewhere [here](https://git.sr.ht/~krobelus/dotfiles/tree/f0fbf9c77bd413973b82899c73820034430e502b/item/.config/kak/kakrc#L926))
Scratch buffers do have their advantages but for an ephemeral menu they are overkill.

We could make scratch buffers the data source for all these things but since
scratch buffers are global `lsp-document-symbol` will overwrite any existing
`*goto*` buffer.  Same for references, and multiple workspace-symbol queries.
(OTOH this limitation doesn't apply to diagnostics.)

So I think the callback approach of `lsp-connect` is nice because it doesn't
modify global state. Of course there are limitations (it blocks Kakoune
until done).

So I'd try to add (JSON) diagnostics to `lsp-connect`.

Also it shouldn't be hard to get workspace-symbol working, we just need to
switch the argument order, then it will look like

	lsp-connect show-response-debug lsp-workspace-symbol foo

I take it you usually wouldn't need the nesting information from documentSymbol, or its distinction between the range/selectionRange fields?
If not, then we should probably add a simple output format to `lsp-connect`,
that matches the corresponding scratch buffer:

	test.c:3:12: int foo()
	test.c:7:21: void bar()


Once we have that we could, we could use lsp-connect to reimplement callbacks
that are currently being called from Rust like `lsp-show-document-symbol`.
Though that would not be a good idea if we had to give up the async behavior.

krobelus:
I've fixed diagnostics and changed the argument order of lsp-connect to allow passing request args

The new commit uses lsp-connect to add menus for jumping to document symbols, references or diagnostics.
workspace symbols are missing; that looks trickier.
Anyway, I intend to merge soon.

Issue #582, opened 2021-12-29
Feature request: lsp-code-action-sync


whereswaldon:
Firstly, thanks to everyone who works on kak-lsp. You all save me so much time every day. :D

I'd like to request that a companion to the recent `:lsp-code-action` command be added that runs the code action synchronously. That would allow for use cases like using `gopls` to "Organize Imports" in a `BufWritePre` hook. Since the code actions are currently run asynchronously, the import rewrite currently may finish _after_ the buffer is written, which has all kinds of weird side effects.

Is this possible/easy to add?

eliasnaur:
I tried

```
  define-command lsp-code-action-sync -params 1 -docstring "lsp-code-action <pattern>: perform the code action that matches the given regex" %{
      lsp-did-change-and-then "lsp-code-actions-request-sync true '%sh{printf %s ""$1"" | sed ""s/'/''/g""}'"
  }

  define-command -hidden lsp-code-actions-request-sync -params 1..2 -docstring "Request code actions for the main cursor position" %{
      evaluate-commands -no-hooks %sh{
  tmp=$(mktemp -q -d -t 'lsp-formatting.XXXXXX' 2>/dev/null || mktemp -q -d)
  pipe=${tmp}/fifo
  mkfifo ${pipe}

  code_action_pattern=""
  if [ $# -eq 2 ]; then
      code_action_pattern="codeActionPattern = \"$(printf %s "$2" | sed 's/\\/\\\\/g; s/"/\\"/g')\""
  fi

  (printf '
  session   = "%s"
  client    = "%s"
  buffile   = "%s"
  filetype  = "%s"
  version   = %d
  fifo      = "%s"
  method    = "textDocument/codeAction"
  [params]
  position.line     = %d
  position.column   = %d
  performCodeAction = %s
  %s
  ' "${kak_session}" "${kak_client}" "${kak_buffile}" "${kak_opt_filetype}" "${kak_timestamp}" "${fifo}" ${kak_cursor_line} ${kak_cursor_column} "$1" "$code_action_pattern" | eval   "${kak_opt_lsp_cmd} --request")

  cat ${pipe} | tee /tmp/pipe
  rm -rf ${tmp}
  }}
```

based on lsp-formatting-sync, but gave up (`lsp-code-action-sync "Organize Imports"` hangs).

krobelus:
thanks for the report.
The fifo solution does not work because code actions (at least the ones sent by gopls) need two round trips.
[...]

Issue #485, opened 2021-05-07
Support offset encoding negotiation


krobelus:
This is an unofficial LSP extension, see https://clangd.llvm.org/extensions.html#utf-8-offsets

Since 98295215a7a2b6fb256d73bd1d85334d086aa800 we always send our preferred encoding during initialization, and assume it will be used.

PR #587, opened 2022-01-21
Write semantic tokens config with one rule per line using inline tables

krobelus:
In our config, each entry takes up 5 lines.  Since some language
servers (like rust-analyzer) provide a large number of tokens (>30),
this means that a semantic token configuration quickly exceeds the
size of a terminal window, which makes it a bit less convenient to
skim/read/edit.

We can already use [TOML Inline
Tables](https://github.com/toml-lang/toml/blob/master/toml.md#user-content-inline-table)
to use only one line for each token-face-rule.
However, that would mean we'd need to write something like

	semantic_tokens = [ ... ]

which we can only write at the beginning of kak-lsp.toml (because it
must be inside the nameless global TOML table).  I'd like to keep the
semantic tokens configuration at the bottom of this file because unlike
server configuration, this feature is not required for basic operation.

Use "semantic_tokens.rules" instead of "semantic_tokens" so we can
write it as inline table anywhere in the TOML file. The old syntax
continues to be valid.

Potential concerns:
- the "rules" name feels arbitrary, I couldn't come up with a better name.

Closes #558

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -353,11 +353,11 @@ The faces used for semantic tokens and modifiers can be modified in `kak-lsp.tom
 ----
 [semantic_tokens]
 faces = [
-    {token="variable", face="const_variable_declaration", modifiers=["constant", "declaration"]},
+    {face="const_variable_declaration", token="variable", modifiers=["constant", "declaration"]},
 ]
 ----
 
-where `token` is the token's name as reported by the language server (see `lsp-capabilities`), `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config) and `modifiers` is an array of modifier names (also reported by the language server). `modifiers` may be omitted, but `token` and `face` are required.
+where `face` is the face that will be applied in Kakoune (you'll want to define these in your theme/config), `token` is the token's name as reported by the language server (see `lsp-capabilities`) and `modifiers` is an array of modifier names (also reported by the language server). `modifiers` may be omitted, but `token` and `face` are required.
 
 You may create any arbitrary number of definitions with permutations between the token names and modifiers reported by the server. For an entry to match a token, all the entry's modifiers must exist on the token. However, the token may have additional modifiers not assigned in the config entry. +
 `kak-lsp` will find the most specific matching configuration to apply, where specificity is defined as the number of matching modifiers. If multiple matching entries have the same number of modifiers, the one that was defined last in the configuration wins.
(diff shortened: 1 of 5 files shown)


sclu1034 (approved):
> the "rules" name feels arbitrary, I couldn't come up with a better name.

I can't say "Yes, that's the one" here either, but it's not a bad option.
Other candidates in no particular order:
- `config`
- `map`
- `mappings`
- `tokens`
- `faces`
- `list`
- `values`
- `match`
- `matches`

krobelus:
I like `faces` because it sets the output (face) apart from the inputs (token/modifiers).
Might also want to switch the order to face/token/modifiers or token/modifiers/face instead of token/face/modifiers.

sclu1034:
> Might also want to switch the order to face/token/modifiers or token/modifiers/face instead of token/face/modifiers.

I'd prefer the first one, then. Since `modifiers` is optional, it should come last to keep positioning somewhat consistent.

sclu1034 (approved):

krobelus:
Thanks for the review. I'll tag a release soon; this is the changelog:

Breaking changes:
- [`python-lsp-server`](https://github.com/python-lsp/python-lsp-server) replaces [python-language-server](https://github.com/palantir/python-language-server) as default language server for Python.
- `rust-analyzer` replaces `rls` as default language server for Rust (#578).

Additions:
- New command `lsp-code-action-sync` is the synchronous variant of `lsp-code-action`, suitable for use in `BufWritePre` hooks (#582).
- The code action menu now provides fuzzy (subsequence) filtering but no longer provides `<c-n>`/`<c-p>` due to implementation reasons (b4ee2a3).
- When accepting a completion, any `additionalTextEdits` are now applied. For example rust-analyzer adds import statements this way (2637b26).
- `lsp-handle-progress` gained a default implementation that shows an hourglass (⌛) in the modeline when a language server is busy (f8a8cf8).
- The default configuration for `semantic_tokens` now uses a more condensed syntax (#587).
- Some systems' C libraries ship old Unicode databases, but terminals often bundle newer versions. To avoid graphical glitches, emoji are only used if they are assigned display width 2 by the C library's `wcwidth(3)` (73c2d1a, c3aeb0d).
- The [clangd protocol extension for offset encoding negotiation](https://clangd.llvm.org/extensions.html#utf-8-offsets) is now supported, which means kak-lsp now uses UTF-8 offsets if the server supports them (#485).

Bug fixes:
- `lsp-formatting-sync` now works again with `texlab` instead of blocking forever (regressed in 11.0.0) (a12831a).
- Certain text edits (such as some code actions from `rust-analyzer`) are now applied correctly (35ba8de).

sclu1034:
> Certain text edits [...] are now applied correctly

This sounds like only that subset works, while all others are still broken 😅 
Maybe reword to "Fix certain text edits [...] not being applied correctly"?

krobelus:
On Tue, Jan 25, 2022 at 01:15:34AM -0800, Lucas Schwiderski wrote:
> > Certain text edits [...] are now applied correctly
> 
> This sounds like only that subset works, while all others are still broken 😅 
> Maybe reword to "Fix certain text edits [...] not being applied correctly"?

Yup, thanks. (I hope all text edits work now)

PR #602, opened 2022-03-30
Add textDocument/inlayHint and experimental/inlayHints

### LINKED ISSUE #600: Support LSP Inlay Hints
Rust analyzer recently switched away from their custom inlay hints, to the upstream inlay hints: https://github.com/rust-analyzer/rust-analyzer/pull/11445

We should support these as well. I have not looked closer at the implementation to find differences with the current api

Spec: https://github.com/microsoft/language-server-protocol/pull/1249

topisani:
Close #600 

For now this doesnt remove the old rust-analyzer api, and adds both the `textDocument/inlayHint` and `experimental/inlayHints`. The former is proposed for LSP v3.17, and the latter is the temporary name used for the proposed api by rust-analyzer.

I think we should remove both old apis when the new one is standardized, and adapted by rust-analyzer, since i have seen no indication that other servers have used the rust-analyzer or experimental prefix versions.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -209,7 +209,7 @@ declare-option -hidden range-specs cquery_semhl
 declare-option -hidden int lsp_timestamp -1
 declare-option -hidden range-specs lsp_references
 declare-option -hidden range-specs lsp_semantic_tokens
-declare-option -hidden range-specs rust_analyzer_inlay_hints
+declare-option -hidden range-specs lsp_inlay_hints
 declare-option -hidden range-specs lsp_diagnostics
 declare-option -hidden str lsp_project_root
 
@@ -1158,6 +1158,39 @@ incomingOrOutgoing = $1
 " | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }
 
+
+define-command -hidden lsp-inlay-hints -docstring "lsp-inlay-hints: request inlay hints" %{
+  lsp-did-change-and-then lsp-inlay-hints-request
+}
+
+define-command -hidden lsp-inlay-hints-request %{
+    nop %sh{ (printf %s "
+session  = \"${kak_session}\"
+buffile  = \"${kak_buffile}\"
+filetype = \"${kak_opt_filetype}\"
+version  = ${kak_timestamp:-0}
+method   = \"textDocument/inlayHint\"
+[params]
+buf_line_count = ${kak_buf_line_count}
+" | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
+}
+
+define-command -hidden lsp-experimental-inlay-hints -docstring "lsp-experimental-inlay-hints: Request inlay hints with experimental prefix (lsp)" %{
+  lsp-did-change-and-then lsp-experimental-inlay-hints-request
+}
+
+define-command -hidden lsp-experimental-inlay-hints-request %{
+    nop %sh{ (printf %s "
+session  = \"${kak_session}\"
+buffile  = \"${kak_buffile}\"
+filetype = \"${kak_opt_filetype}\"
+version  = ${kak_timestamp:-0}
+method   = \"experimental/inlayHints\"
+[params]
+buf_line_count = ${kak_buf_line_count}
+" | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
+}
+
 # CCLS Extension
 
 define-command ccls-navigate -docstring "Navigate C/C++/ObjectiveC file" -params 1 %{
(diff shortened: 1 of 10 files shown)


rc/lsp.kak
```diff
@@ -1158,6 +1158,39 @@ incomingOrOutgoing = $1
 " | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }
 
+
+define-command lsp-inlay-hints -docstring "lsp-inlay-hints: Request inlay hints (lsp)" %{
```
krobelus, line 1162: I think we should drop the `(lsp)` suffix because it will become obsolete once we remove the rust-analyzer variant.
I also prefer lowercase after colon (like most Kakoune commands) but that's an unrelated change.

```suggestion
define-command lsp-inlay-hints -docstring "lsp-inlay-hints: request inlay hints" %{
```

src/language_features/inlay_hints.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use crate::{
+    context::Context,
+    position::lsp_position_to_kakoune,
+    types::{EditorMeta, EditorParams},
+    util::{editor_quote, escape_tuple_element},
+};
+
+#[derive(Debug, PartialEq, Clone, Default, Serialize, Deserialize)]
+struct InlayHintsOptions {
```
krobelus, line 15: good point defining this here. I just realized that a lot of types in `src/types.rs` should rather be in the module where they're used.
The name collision with InlayHintsParams is ugly but fair enough.

src/language_features/inlay_hints.rs
```diff
@@ -0,0 +1,105 @@
...
+                    ""
+                };
+                // Escape markup inside label
+                let label = escape_tuple_element(&label).replace('{', "\\{");
+                editor_quote(&format!(
+                    "{position}+0|{padding_left}{{InlayHint}}{label}{padding_right}",
+                ))
+            },
+        )
+        .collect::<Vec<String>>()
```
krobelus, line 96: unrelated but there's no need to collect() before join() (if we import `itertools::Itertools`)

src/util.rs
```diff
@@ -52,12 +52,12 @@ impl Drop for TempFifo {
 
 /// Escape Kakoune string wrapped into single quote
 pub fn editor_escape(s: &str) -> String {
-    s.replace("'", "''")
+    s.replace('\'', "''")
```
krobelus, line 55: not sure about this. If we want this change it should be in a separate commit, and handle other single-char replace like in `src/language_features/hover.rs` too.

krobelus:
looks nice, I have never used inlay hints before; they seem much less invasive than inlay diagnostics

src/util.rs
topisani: Oh yeah, I just naively resolved the clippy fixmes out of habit - I'll clear them out of this

src/language_features/inlay_hints.rs
```diff
@@ -0,0 +1,105 @@
...
+                } else {
+                    ""
+                };
+                let padding_right = if padding_right.unwrap_or(false) {
+                    "{Default} "
+                } else {
+                    ""
+                };
+                // Escape markup inside label
+                let label = escape_tuple_element(&label).replace('{', "\\{");
```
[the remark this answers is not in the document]
topisani: The doubly escaped backslashes are why I inlined the markup escape - this works, your suggestion does not, since it doubly escapes the backslashes, undoing the markup escaping

src/util.rs
krobelus: Huh, I didn't see clippy complaining. OTOH I do get code actions from rust-analyzer to transform string -> char and vice versa, but those are just assists.

src/language_features/inlay_hints.rs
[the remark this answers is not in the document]
krobelus: wow I had no idea. This means that if the label contains \{} then there is no way to render it properly; we are either losing the backslash or the braces.
The markup seems more important, so LGTM

src/util.rs
topisani: It might have been those - i may have just assumed they were from clippy

topisani: https://rust-lang.github.io/rust-clippy/master/index.html#single_char_pattern
Ah, yes, there it is

topisani:
This should address the comments - i added `lsp-inlay-hints-{enable,disable}`, and I added docs and an entry in the changelog

As mentioned, i havent removed the old rust-analyzer-inlay-hints api, but i guess we could do that as well?

krobelus:
Nice changes, thanks! They should be squashed before merging so the history is less noisy. I generally squash instead of pushing fixup commits.

I think instead of removing `rust-analyzer-inlay-hints` we should just make it delegate to `lsp-experimental-inlay-hints`.
(and as soon as rust-analyzer switches we can change it to `lsp-inlay-hints`).
This avoids breaking existing configurations, at essentially no cost (save for confusing users who are switching back to old rust-analyzer versions and expect that `rust-analyzer-inlay-hints` would do the old thing). We'll still deprecate the command (removing it from docs and making it hidden).

src/util.rs
krobelus: strange that my clippy doesn't suggest that. Anyway, since it's blessed by clippy the idea seems okay.
If possible we should have a reproducible `cargo clippy` config, not sure how that would work

krobelus (approved):

rc/lsp.kak
```diff
@@ -1873,6 +1906,31 @@ define-command lsp-stop-on-exit-disable -docstring "Don't end kak-lsp session on
     alias global lsp-exit lsp-exit-editor-session
 }
 
+define-command lsp-inlay-hints-enable -params 1 -docstring "lsp-inlay-hints-enable <scope>: Enable inlay hints for <scope>" %{
+    add-highlighter "%arg{1}/lsp_inlay_hints" replace-ranges lsp_inlay_hints
+    hook -group lsp-inlay-hints %arg{1} BufReload .* lsp-inlay-hints
```
krobelus, line 1911: I guess this means we should make `lsp-inlay-hints` hidden because most users will be happy with `lsp-inlay-hints-enable`

topisani:
Alright, i removed the rust-analyzer/inlayHints implementation, and let the kak commands delegate to lsp-experimental-inlay-hints.

I rebased and squashed, so this should be ready for merge

src/language_features/inlay_hints.rs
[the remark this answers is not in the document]
krobelus: I can't reproduce the problem with my suggestion anymore.
OTOH, the current version fails to display inlay hints if I add something like `let label = r#"123\{456"#;` above.

This command illustrates that it works fine to double backslashes at every level of escaping (in addition to escaping { and |)

```
kak -e '
	exec %{%dihello world<esc>}
	declare-option range-specs my_ranges
	add-highlighter global/ replace-ranges my_ranges
	# raw value:                      \{
	#    -> escaped for markup:     \\\{
	#      -> escaped for tuple: \\\\\\{
	set-option global my_ranges %val{timestamp} %<1.5+0|\\\\\\{>
	# renders as hell\{o world
'
```

will commit a fix

PR #607, opened 2022-04-04
Add envs parameter to language config

### LINKED ISSUE #606: Configuration for server-specific environment variables
Example use case: right now, I have to run `kotlin-language-server` with a custom `CLASSPATH` variable to configure a specific workaround. Other language servers may have configuration which is done exclusively through environment variables.

Right now, I run this via `command = "env"` and `args = ["CLASSPATH=...", "kotlin-language-server"]`, but it would be nice to be able to just set an `envs = { CLASSPATH = "..." }` variable alongside a `command = "kotlin-language-server"` config.

So, the main functionality here would be something along the lines of an extra `envs = { ... }` table for each section in `kak-lsp.toml` which would set the given environment variables before running the server.

Going to see if I can tackle this myself right now, but filing an issue regardless just for context.

clarfonthey:
Justification and use case in #606; closes #606.

I chose the name `envs` since it's what Rust uses and seems reasonable, but can rename depending on what's desired.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -47,6 +47,8 @@ pub struct LanguageConfig {
     pub command: String,
     #[serde(default)]
     pub args: Vec<String>,
+    #[serde(default)]
+    pub envs: HashMap<String, String>,
     pub settings_section: Option<String>,
     pub settings: Option<Value>,
     pub offset_encoding: Option<OffsetEncoding>,
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -40,7 +40,7 @@ pub fn start(
         // should be fine to unwrap because request was already routed which means language is configured
         let lang = &config.language[&route.language];
         offset_encoding = lang.offset_encoding;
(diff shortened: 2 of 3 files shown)


src/types.rs
```diff
@@ -47,6 +47,8 @@ pub struct LanguageConfig {
     pub command: String,
     #[serde(default)]
     pub args: Vec<String>,
+    #[serde(default)]
+    pub envs: HashMap<String, String>,
```
krobelus, line 51: In hindsight, it would probably have been better to give users a more natural way to run shell commands. Right now we have

```toml
command = "sh"
args = ["-c", """CLASSPATH=... kotlin-language-server"""]
```

it could be as simple as

```toml
command = """CLASSPATH=... kotlin-language-server"""
```

Anyway, that would be a bigger change.
The envs addition looks good to me.

clarfonthey: FWIW I opened #608 so that this is tracked somewhere, since I noticed it didn't have an issue already.

PR #610, opened 2022-04-07
Update julia lsp config

uncomfyhalomacro:


kak-lsp.toml
```diff
@@ -149,7 +149,12 @@ roots = ["package.json"]
 command = "json-languageserver"
 args = ["--stdio"]
 
-# Requires Julia packages "LanguageServer", "StaticLint" and "SymbolServer"
+# Requires Julia packages "LanguageServer"
+# Run: `julia --project=@kak-lspconfig -e 'import Pkg; Pkg.add("LanguageServer")'` to install it
+# No need to import SymbolServer because it is no longer required
```
krobelus, line 154: I think these two lines should be moved to the commit message.
They don't seem useful to new users.

kak-lsp.toml
```diff
@@ -149,7 +149,12 @@ roots = ["package.json"]
 command = "json-languageserver"
 args = ["--stdio"]
 
-# Requires Julia packages "LanguageServer", "StaticLint" and "SymbolServer"
+# Requires Julia packages "LanguageServer"
+# Run: `julia --project=@kak-lspconfig -e 'import Pkg; Pkg.add("LanguageServer")'` to install it
```
krobelus, line 153: so this means we give precedence to the LanguageServer package installed to ~/.julia/environments/kak-lspconfig?
Why is that better than installing LanguageServer without the --project flag?

A non-obvious manual step is not great.
Also, ~/.julia/environments/kak-lsp would be a more fitting name.

kak-lsp.toml
```diff
@@ -149,7 +149,12 @@ roots = ["package.json"]
 command = "json-languageserver"
 args = ["--stdio"]
 
-# Requires Julia packages "LanguageServer", "StaticLint" and "SymbolServer"
+# Requires Julia packages "LanguageServer"
```
krobelus, line 152:
```suggestion
# Requires the Julia package "LanguageServer"
```

kak-lsp.toml
uncomfyhalomacro: sure. i will remove that.

kak-lsp.toml
uncomfyhalomacro: I think this is the best way to avoid deps from other packages conflicting each other. There are instances where LanguageServer.jl and another package depends on the same dependency and the other one requires a version lower and may break some functionalities with LanguageServer.jl. It is better for it to be in a separate environment.

uncomfyhalomacro: Also, if the user does not have `LanguageServer.jl` in `~/.julia/environments/kak-lsp`, it will look for `LanguageServer.jl` in the default global environment.

kak-lsp.toml
```diff
@@ -159,15 +160,31 @@ args = [
...
-        server.runlinter = true;
-        run(server);
+    ls_install_path = joinpath(get(DEPOT_PATH, 1, joinpath(homedir(), ".julia")), "environments", "kak-lsp");
+    pushfirst!(LOAD_PATH, ls_install_path);
+    using LanguageServer;
+    popfirst!(LOAD_PATH);
+    depot_path = get(ENV, "JULIA_DEPOT_PATH", "");
+    buffer_file = ENV["kak_buffile"];
+    project_path = let 
+      dirname(something(
```
krobelus, line 170: please consider squashing your changes (I don't care about your past mistakes) and documenting the behavior changes (+ rationale) in the commit message.

kak-lsp.toml
```diff
@@ -159,15 +160,31 @@ args = [
...
+        Base.load_path_expand((
+          p = get(ENV, "JULIA_PROJECT", nothing);
+          p === nothing ? nothing : isempty(p) ? nothing : p
+        )),
+        # 2. Check for Project.toml in current working directory
+        Base.current_project(pwd()),
+        # 3. Check for Project.toml from buffer's full file path exluding the file name
+        Base.current_project(dirname(buffer_file)),
+        # 4. Fallback to global environment
+        Base.active_project()
```
krobelus, line 181: Why are there differences to the neovim config?

The "Configuration taken from https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/julials.lua"
comment is useful, we should keep it but use a permalink (commit SHA instead of "master").

kak-lsp.toml
```diff
@@ -1,4 +1,4 @@
-snippet_support = false
+snippet_support = true
```
krobelus, line 1: ?

kak-lsp.toml
uncomfyhalomacro: Okay :+1:

kak-lsp.toml
uncomfyhalomacro: I accidentally copied my config, going to revert that line with the squashed commits later

kak-lsp.toml
uncomfyhalomacro: Opening a new *unsaved* buffer to file will cause the LSP to use a different environment. I experimented the default config from nvim-lspconfig, and created a file inside a directory with no `Project.toml`. What I noticed is that instead of falling back to default global environment, it assumed an **empty** project environment in `$HOME` instead of `$HOME/.julia/environments/<current-version-of-julia>`. So to fix the issue:
1. We have to check if there is a explicitly created `JULIA_PROJECT` environmental variable so when we do `JULIA_PROJECT=<path-to-project> kak newfile.jl`, it will use that environment regardless if the Project.toml is empty or not.
2. Use the current working directory as the basis of the current project environment.
3. If a file does not exist yet, we try using the path where we *will* create the file and use it as the basis for the project environment, hence, the `dirname` function from Julia's Base module.
4. We fallback to the global environment, hence, `Base.active_project()` *since we didn't use the `--project` flag in the arguments anyway*.

There are slight differences actually with the nvim-lspconfig but *I actually made my own lsp config instead of using theirs due to some experiences I have*. However, both follow the same goal through "hacky" workarounds. For example, these lines 

```julia
ls_install_path = joinpath(get(DEPOT_PATH, 1, joinpath(homedir(), ".julia")), "environments", "kak-lsp")
pushfirst!(LOAD_PATH, ls_install_path);
using LanguageServer;
popfirst!(LOAD_PATH);
```

It is a hacky workaround so we won't populate the LSP with symbols from the LanguageServer.jl and just use the plain environment of the global environment.

I hope this answers your questions. :smile:

krobelus:
Great work, thank you!
cc @koehlerson who might be interested.
This PR
1. Drops the need to explicitly install StaticLint and SymbolServer
2. Loads LanguageServer from an isolated environment if possible (to avoid breakage when dependencies are upgraded).
3. Improves detection of the project root
     This probably already works correctly if there is a `Project.toml` because the language server will be started in that directory. But other scenarios are fixed, for example when `JULIA_PROJECT` is given.

koehlerson:
really cool! thanks to both of you :)

PR #616, opened 2022-05-15
Fix snippets

topisani:
Its been a long time since ive used snippets, and they have been broken for me for a while, so i spent some time fixing it while listening to the new kendrick album :)

Demo Gif

 - Snippets are now only fully expanded when the completions menu is closed, making them much more robust
 - The text that is inserted until the menu is closed is the snippets contents merged onto one line and with tabstops (cursor jump points) replaced with their placeholder text.

These two things result in a pretty smooth snippets experience, that is a lot more practical than the old implementation, while staying fairly simple.

Currently, i have only tested with rust-analyzer, but it should be better than the old implementation everywhere.

At some point i would like to rewrite the placeholder selection and location jumping, preferably not using perl, but that wont be this time.

Also, this includes a bit of infrastructure to perform actions when completions are closed, including properly deleting the temporary text. This could be used to implement #40 pretty easily, but text edits would then only be applied when the completions box is closed, and a temporary label would have to be used as insert_text. This is also probably not something ill do this time around

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -519,22 +519,31 @@ kak-lsp has experimental support for snippets. It is enabled by setting `snippet
 
 It uses the two faces `SnippetsNextPlaceholders` and `SnippetsOtherPlaceholders`, you may want to customize those.
 
-To properly use snippets, you'll probably want something like this:
+You'll probably want to add a mapping for `lsp-snippets-select-next-placeholders`, which allows
+to jump between tabstops (like function call arguments).
+
+For example, you can use `<tab>` to select a placeholder if there is one, or else execute
+`<tab>` as normal:
 
 [source,kak]
 ----
-def -hidden insert-c-n %{
- try %{
-   lsp-snippets-select-next-placeholders
-   exec '<a-;>d'
- } catch %{
-   exec -with-hooks '<c-n>'
- }
-}
-map global insert <c-n> "<a-;>: insert-c-n<ret>"
+map global normal <tab> ': try lsp-snippets-select-next-placeholders catch %{ execute-keys -with-hooks <lt>tab> }<ret>' -docstring 'Select next snippet placeholder'
+map global insert <tab> '<a-;>: try lsp-snippets-select-next-placeholders catch %{ execute-keys -with-hooks <lt>tab> }<ret>' -docstring 'Select next snippet placeholder'
 ----
 
-This maps `<c-n>` to select the next placeholder if there is one, and otherwise executes `<c-n>` as normal
+or `<c-n>` (while the completion menu is hidden):
+
+[source,kak]
+----
+map global normal <c-n> ': try lsp-snippets-select-next-placeholders catch %{ execute-keys -with-hooks <lt>c-n> }<ret>' -docstring 'Select next snippet placeholder'
+map global insert <c-n> '<a-;>: lsp-snippets-select-next-placeholders<ret>' -docstring 'Select next snippet placeholder'
+hook global InsertCompletionShow .* %{
+  unmap global insert <c-n> '<a-;>: lsp-snippets-select-next-placeholders<ret>'
+}
+hook global InsertCompletionHide .* %{
+  map global insert <c-n> '<a-;>: lsp-snippets-select-next-placeholders<ret>' -docstring 'Select next snippet placeholder'
+}
+----
 
 
 == Limitations


topisani:
Another small demo with placeholders, which now work fairly smoothly

rc/lsp.kak
```diff
@@ -2123,18 +2153,15 @@ decl -hidden int-list lsp_snippets_placeholder_groups
...
-            lsp-snippets-select-next-placeholders
-            exec '<a-;>d'
-        }
+declare-option -hidden str lsp_snippet_to_insert ""
+def -hidden lsp-snippets-insert-completion -params 1 %{ eval %{
+    set window lsp_snippet_to_insert %arg{1}
+    lsp-completion-on-dismiss %{
+        # Delete the inserted text
+        select %opt{lsp_completion_selection}
+        exec '<a-;>d'
```
krobelus, line 2162: Wow Kendrick is back! I've almost given up hope, same for proper snippets support. I tried to write a similar patch last week but quickly gave up.
Your approach looks good, I'll take a closer look tomorrow.

There seems to be an issue still with gopls: the right parenthesis of a call is deleted.
I'm sure we can fix that one way or another. If worst comes to worst we can even make a temporary RawKey hook.
But since that was already broken the same way it's not a blocker.

>  - Snippets are now only fully expanded when the completions menu is closed, making them much more robust

yep, given how Kakoune's completion engine works that's definitely the way to go

> This could be used to implement #40 pretty easily,

Right. I think we already support most text edits you'd find in the wild, let me know if you find a missing one. Then we have motivation to add full support.

topisani: > There seems to be an issue still with gopls: the right parenthesis of a call is deleted.

Hmm, i thought i fixed this on the last force push by replacing empty placeholders with a single space, but i may have missed some cases.
Placeholder parsing and jumping should probably be fully rewritten though. Maybe using replace-ranges to show the placeholder instead, though this also has its issues.

src/language_features/completion.rs
```diff
@@ -196,21 +192,23 @@ pub fn editor_completion(
...
+                }
                 let snippet = insert_text;
-                let insert_text = snippet_prefix_re
-                    .find(&snippet)
-                    .map(|x| x.as_str())
-                    .unwrap_or(&snippet);
+                let insert_text = SNIPPET_TABSTOP_RE.replace_all(&snippet, "");
+                let insert_text = SNIPPET_PLACEHOLDER_RE.replace_all(&insert_text, "$placeholder");
+                // Theres some issue with multiline insert texts, and they also dont work well in the UI, so display on one line
+                let insert_text = insert_text.replace('\n', "");
```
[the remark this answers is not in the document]
topisani: This might be a problem yes, but im not really sure how to solve it. We could trim it to just the first line, or even use ellipsis and set a max length.

We could also experiment with moving the cursor to the beginning, but i would prefer to avoid messing too much with the cursor and such while selecting completions.

It may also be possible to actually insert multiline text, but this is also not really a good experience in kakounes completion system, as the inserted text is hidden by the menu

rc/lsp.kak
krobelus: Here is a failing test that shows the problem with gopls. Anyway, we can fix this later

```diff
diff --git a/test/gopls-completion-snippet.sh b/test/gopls-completion-snippet.sh
new file mode 100755
index 00000000..0ca60e9c
--- /dev/null
+++ b/test/gopls-completion-snippet.sh
@@ -0,0 +1,34 @@
+#!/bin/sh
+
+# REQUIRES: command -v gopls
+
+. test/lib.sh
+
+cat > .config/kak-lsp/kak-lsp.toml << EOF
+snippet_support = true
+[language.go]
+filetypes = ["go"]
+roots = ["*.go"]
+command = "gopls"
+EOF
+
+
+cat > main.go << EOF
+package main
+
+import "os"
+
+func foo(i int, j int) {
+}
+EOF
+
+test_tmux_kak_start 'edit main.go'
+
+test_tmux send-keys 4jo fo
+test_sleep
+test_tmux send-keys C-n
+test_sleep
+test_tmux send-keys 1
+test_sleep
+test_tmux capture-pane -p | grep '^\s*foo'
+# CHECK: foo(1)
```

src/language_features/completion.rs
```diff
@@ -70,17 +69,14 @@ pub fn editor_completion(
...
+                "lsp-completion-item-resolve-request true".to_string()
             } else {
                 "".to_string()
             };
-            let on_select = format!(
-                "{}info -markup -style menu -- %§{}§",
-                &maybe_set_index,
+            let on_select = formatdoc!(
+                "lsp-completion-item-selected {completion_item_index}
+                {maybe_set_index}
```
krobelus, line 78: now that this variable no longer sets the index, it should be renamed

also, it's our convention to align the line contents, so the "l" should align with the "{".

```suggestion
                 {maybe_resolve} 
```

README.asciidoc
```diff
@@ -523,19 +523,11 @@ To properly use snippets, you'll probably want something like this:
...
- try %{
-   lsp-snippets-select-next-placeholders
-   exec '<a-;>d'
- } catch %{
-   exec -with-hooks '<c-n>'
- }
-}
-map global insert <c-n> "<a-;>: insert-c-n<ret>"
+map global insert <c-a-n> '<a-;>: lsp-snippets-select-next-placeholders<ret>'
+map global normal <c-a-n> ': lsp-snippets-select-next-placeholders<ret>'
```
krobelus, line 527: I have never used it but the old c-n seems nice because of the simple UI,
Maybe we should map c-n (or tab?) by default eventually. Usually, plugins aren't supposed to add key bindings but kak-lsp ought to be an exception.

I think the old behavior still works if we just tweak the try-block a bit:

	define-command -override -hidden insert-c-n %{
		try %{
			lsp-snippets-select-next-placeholders
		} catch %{
			exec -with-hooks '<c-n>'
		}
	}

rc/lsp.kak
```diff
@@ -342,9 +342,39 @@ have_kakoune_feature_filtertext = ${kak_opt_lsp_have_kakoune_feature_filtertext}
 " | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }}
 
+declare-option -hidden str lsp_completion_selection ""
+
 define-command -hidden lsp-completion-dismissed -docstring "Called when the completion pager is closed" %{
+    set-option window lsp_completion_selection %val{hook_param}
+    try %{
+        eval -draft %{
```
krobelus, line 350: please use long names here ("evaluate-commands").
(The only exception is the snippets code, which was vendored from occivink's snippets plugin.)

rc/lsp.kak
```diff
@@ -342,9 +342,39 @@ have_kakoune_feature_filtertext = ${kak_opt_lsp_have_kakoune_feature_filtertext}
...
+            select %opt{lsp_completion_selection}
+        }
+    } catch %{
+        # Completion was cancelled, nothing was inserted
+        # Clear these so the following lines do nothing.
+        # This is to avoid runing them in the above try block, and silencing their errors
+        remove-hooks window lsp-completion-dismissed
+        set-option window lsp_completions_selected_item -1
+    }
     lsp-completion-item-resolve-request false
```
krobelus, line 360: I think this can be moved inside the `try` block since in the failure cause the selected item is always -1 (I hope).
Also the next line that triggers the user hook because we don't need to run insert snippets if the above `select` fails right?

rc/lsp.kak
```diff
@@ -342,9 +342,39 @@ have_kakoune_feature_filtertext = ${kak_opt_lsp_have_kakoune_feature_filtertext}
...
+    remove-hooks window lsp-completion-dismissed
     set-option window lsp_completions_selected_item -1
+    set-option window lsp_completion_selection ""
+}
+
+# User: lsp-completion-dismissed <command>
+# Run <command> when the completion menu is closed.
+# The inserted range is available as %opt{lsp_completion_selection} in the format expected by `select`.
+define-command -hidden lsp-completion-on-dismiss -params 1 %{
+    hook -once -group lsp-completion-dismissed window User "^LspCompletionDismissed$" %arg{1}
```
krobelus, line 371: The regex anchors ^ and $ are not necessary because `:doc hooks` says:

> filtering_regex must match the entire parameter string in order for the commands to be executed.

rc/lsp.kak
```diff
@@ -342,9 +342,39 @@ have_kakoune_feature_filtertext = ${kak_opt_lsp_have_kakoune_feature_filtertext}
...
+    try %{
+        eval -draft %{
+            select %opt{lsp_completion_selection}
+        }
+    } catch %{
+        # Completion was cancelled, nothing was inserted
+        # Clear these so the following lines do nothing.
+        # This is to avoid runing them in the above try block, and silencing their errors
+        remove-hooks window lsp-completion-dismissed
+        set-option window lsp_completions_selected_item -1
```
krobelus, line 358: I'm pretty sure the selected item is always -1 in this case, so this looks redundant

rc/lsp.kak
```diff
@@ -342,9 +342,39 @@ have_kakoune_feature_filtertext = ${kak_opt_lsp_have_kakoune_feature_filtertext}
...
+    }
     lsp-completion-item-resolve-request false
+    trigger-user-hook "LspCompletionDismissed"
```
krobelus, line 361: I'm not sure if a user hook is the simplest solution (I'd need to look up why user hooks exist)
Can we instead use a window-scoped option? I feel like that'll end up being less complicated

rc/lsp.kak
```diff
@@ -2234,7 +2261,7 @@ foreach my $i (0 .. $#sel_content) {
         $def =~ s/'\''/'\'''\''/g;
         print(" '\''$def'\''");
     } else {
-        print(" '\'''\''");
+        print(" '\'' '\''");
```
krobelus, line 2264: Ok this means we now always have a non-empty placeholder text? Sounds great, this can potentially simplify `lsp-snippets-select-next-placeholders`

src/language_features/completion.rs
```diff
@@ -196,21 +192,23 @@ pub fn editor_completion(
             // we'll need to perform some transformations on the completion commands.
             if ctx.config.snippet_support && x.insert_text_format == Some(InsertTextFormat::SNIPPET)
             {
+                lazy_static::lazy_static! {
+                    static ref SNIPPET_TABSTOP_RE: Regex = Regex::new(r"\$(?P<i>\d+)").unwrap();
+                    static ref SNIPPET_PLACEHOLDER_RE: Regex = Regex::new(r"\$\{(?P<i>\d+):?(?P<placeholder>[^}]+)\}").unwrap();
```
krobelus, line 197: If we're not using the named captured group "i", then we should remove it to reduce confusion.

I think one usecase of snippets that don't just use lexical order for placeholders is something like this:

	print(r"${1:foo}")
	print(${1:foo})

This means that the first placeholder should get two cursors. In this case this allows to print a variable and its value. VSCode supports this.
Not sure if any language server ever sends something like this. Anyay, not important now.

src/language_features/completion.rs
krobelus, line 203: sounds good. Do you happen to know an example of multiline insert texts? (not important, feel free to ignore)
I just observed that `if let` has no `insert_text` field, so the label is used, which is just one line.

rc/lsp.kak
krobelus: So this seems to work fine for me:

	define-command -hidden lsp-completion-dismissed -docstring "Called when the completion pager is closed" %{
	    set-option window lsp_completion_selection %val{hook_param}
	    try %{
	        eval -draft %{
	            select %opt{lsp_completion_selection}
	        }
	        lsp-completion-item-resolve-request false
	        trigger-user-hook "LspCompletionDismissed"
	    }
	    remove-hooks window lsp-completion-dismissed
	    set-option window lsp_completions_selected_item -1
	    set-option window lsp_completion_selection ""
	}

rc/lsp.kak
krobelus, line 2162: Given

	fn somefunction() {}
	fn main() {
	}

if I type `somefun<c-n>;<ret>` the semicolon will be erased.
I think that's pretty annoying. Sadly that's the range given to us by the InsertCompletionHide hook.
Maybe we can change the hook parameter. It should not include the semicolon.

krobelus:
sorry for the delay, looks great overall.
Would be nice to keep the old c-n binding but that's not a huge deal.
So I'm fine with merging the current behavior

README.asciidoc
krobelus: Ok I realized that this c-n binding will break cyclying through completions, probably there's an easy fix

src/language_features/completion.rs
```diff
@@ -70,17 +69,14 @@ pub fn editor_completion(
...
-                )
+                "lsp-completion-item-resolve-request true".to_string()
             } else {
                 "".to_string()
             };
-            let on_select = format!(
-                "{}info -markup -style menu -- %§{}§",
-                &maybe_set_index,
+            let on_select = formatdoc!(
+                "lsp-completion-item-selected {completion_item_index}
```
krobelus, line 77: You now set `lsp_completions_selected_item` even if the server does not
support the `completionItem/resolve` request. This means we'll send the
`completionItem/resolve` request anyway and gopls will respond with an error.

Since we currently only use `lsp_completions_selected_item` for
resolve-requests, we should only set it we actually want to amek resolve-requests.
Maybe we can improve the naming.

rc/lsp.kak
krobelus: OK we were just missing a case for gopls:

```diff
diff --git a/rc/lsp.kak b/rc/lsp.kak
index 76c69a49..565d11ac 100644
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -2261,6 +2275,8 @@ foreach my $i (0 .. $#sel_content) {
         $def =~ s/\}\}/}/g;
         # double up single-quotes
         $def =~ s/'\''/'\'''\''/g;
+        # avoid empty placeholders because they complicate movement
+        if (length $def == 0) { $def = " " };
         print(" '\''$def'\''");
     } else {
         print(" '\'' '\''");
```

krobelus: It's fairly easy to patch Kakoune to not include the last inserted key in the range specified by the hook parameter..
but it's probably not all that useful. How about instead we just replay the last key (that closed the completion pager) *if it's likely that it inserted something? This should definitely be a separate commit, doesn't need to be here. Roughly like below except we should probably reset to the empty string immediately after removing the `lsp-completion-dismissed` hooks

```
index 76c69a49..8beea37f 100644
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -367,8 +367,13 @@ define-command -hidden lsp-completion-dismissed -docstring "Called when the comp
 # User: lsp-completion-dismissed <command>
 # Run <command> when the completion menu is closed.
 # The inserted range is available as %opt{lsp_completion_selection} in the format expected by `select`.
+declare-option str lsp_key_that_dismissed_completion
 define-command -hidden lsp-completion-on-dismiss -params 1 %{
     hook -once -group lsp-completion-dismissed window User "^LspCompletionDismissed$" %arg{1}
+    set-option window lsp_key_that_dismissed_completion ""
+    hook -group lsp-completion-dismissed window RawKey [^<].*|<backspace>|<del>|<gt>|<lt>|<minus>|<percent>|<plus>|<right>|<semicolon>|<space>|<tab> %{
+        set-option window lsp_key_that_dismissed_completion %val{hook_param}
+    }
 }
 
 # Is called when a completion item is selected
@@ -381,6 +386,7 @@ define-command -hidden lsp-completion-item-resolve-request -params 1 \
     -docstring "Request additional attributes for a completion item" %{
     nop %sh{
         index=$kak_opt_lsp_completions_selected_item
+        exit
         [ "${index}" -eq -1 ] && exit
 
         (printf %s "
@@ -2164,6 +2170,7 @@ def -hidden lsp-snippets-insert-completion -params 1 %{ eval %{
         exec '<a-;>d'
         eval -draft -verbatim lsp-snippets-insert %opt[lsp_snippet_to_insert]
         try lsp-snippets-select-next-placeholders
+        exec -with-hooks %opt{lsp_key_that_dismissed_completion}
     }
 }}
```

topisani: Ah, that makes a lot of sense!

README.asciidoc
topisani: That is why i removed it. This issue, and simply trying to simplify the example.
This is what i use, in my own `lsp-setup` function:

```kak

  hook -group lsp window InsertCompletionShow .* %{
    unmap window insert <c-n>
  }
  hook -group lsp window InsertCompletionHide .* %{
    map window insert <c-n> '<a-;>: lsp-snippets-select-next-placeholders<ret>'
  }
  map window insert <c-n> '<a-;>: lsp-snippets-select-next-placeholders<ret>'
  map window normal <c-n> ': lsp-snippets-select-next-placeholders<ret>'
```

rc/lsp.kak
topisani: I wanted to avoid this, because it means errors in the user hook are silenced

rc/lsp.kak
topisani: User hooks are made for this purpose - i.e. to implement custom hooks, without having to use a dummy option and a SetOption hook. However, since we already have options that are set here, it might be possible to simply detect WinSetOption for `lsp_completions_selected_item` for example.

rc/lsp.kak
topisani: this is fixed by the above, right?

rc/lsp.kak
topisani: yes - and i hope to do that as well at some point

src/language_features/completion.rs
topisani: ah yes, i was experimenting with using a LspCompletionSelected={index} hook, but decided against it. I do however think that it is useful (or at least more intuitive) to have this option always set, so i would rather do the lsp-completion-item-resolve-request another way.

src/language_features/completion.rs
topisani: I really want to support this yes, it would be a part of rewriting select-next-placeholders. And yes, thats why the capture group is named

src/language_features/completion.rs
topisani: i don't. I did consider using the label as the actual completion text until the menu is closed, but the snippet text collapsed to a single line seemed like a better experience.

topisani (commented):
There are some changes still to be made, ill try to find the time during the week

README.asciidoc
krobelus: this sounds great, we should probably make it the default behavior

rc/lsp.kak
krobelus: Of course, sorry I missed that comment

rc/lsp.kak
krobelus: The original request for user hooks is in [this post](https://discuss.kakoune.com/t/ann-coqoune/1127) where they want to run user hooks when a plugin creates a new window.
This allows users of the plugin to customize window placement by defining the corresponding user hooks.

OTOH this patch doesn't intend for users to define a custom `LspCompletionDismissed` hook, so we might as well set a `declare-option -hidden str lsp_completion_on_dismiss` and evaluate it in our `InsertCompletionHide` hook.

src/language_features/completion.rs
krobelus: Sure. First we should find a language server that sends snippets with either multiple cursors or with indices in non-lexical order.

rc/lsp.kak
krobelus: yeah it means I can type `somefun<c-n>;<ret>` to get `somefunction();`, which is correct behavior I think. I'll test some more and submit it eventually unless you do

src/language_features/completion.rs
krobelus: Yeah I see that the current implementation is surprising. 
I think a truly elegant solution would avoid even the shell call in `lsp-completion-item-resolve-request` if the server doesn't support `completionItem/resolve`.
This could be implemented by removing the `InsertCompletionHide` hook (that would call `lsp-completion-item-resolve-request`) if the server lacks support.

Until we have something like that, we can keep the current implementation or make it more intuititve by adding a boolean variable for the capability.

src/language_features/completion.rs
topisani: I would assume rust-analyzer does, but to be fair, I haven't checked. I'll look at that if/when I continue down this road

rc/lsp.kak
topisani: It isnt - this case happens when the completion menu is opened, something is selected (which set selected_item and added on-dismissed hooks), and then deselected (by <c-p> for example), and the menu is closed with nothing inserted. In that case, the selected_item will still hold a value, but needs to be cleared, as no dismiss hooks should run.

rc/lsp.kak
topisani: Oh right, i understand the issue now. It might be worth looking at patching kakoune here, i cant exactly tell what the implications of that would be. Otherwise your solution above could work, but its the kind of tomfoolery i would really like to avoid

rc/lsp.kak
krobelus: You're right, that's a bug in the implementation of `completionItem/resolve`.

krobelus:
I made some minor changes I hope that's okay.

rc/lsp.kak
```diff
@@ -1973,6 +2011,21 @@ map global lsp-selection-range b '<esc>: lsp-selection-range-select bottom<ret>'
...
+    map %arg{1} normal <tab> ': lsp-snippets-select-next-placeholders-or <lt>tab><ret>' -docstring 'Select next snippet placeholder'
+
+    map %arg{1} insert <c-i> '<a-;>: lsp-snippets-select-next-placeholders-or <lt>c-i><ret>' -docstring 'Select next snippet placeholder'
+    map %arg{1} normal <c-i> ': lsp-snippets-select-next-placeholders-or <lt>c-i><ret>' -docstring 'Select next snippet placeholder'
+
+    hook -group lsp window InsertCompletionShow .* "
+      unmap %arg{1} insert <c-n> '<a-;>: lsp-snippets-select-next-placeholders<ret>'
+    "
+    hook -group lsp window InsertCompletionHide .* "
+      map %arg{1} insert <c-n> '<a-;>: lsp-snippets-select-next-placeholders<ret>' -docstring 'Select next snippet placeholder'
```
krobelus, line 2024: I don't like this specific option because it doesn't fix the underlying problem that keys like `goto r` are still mapped every time we run `lsp-enable`.

rc/lsp.kak
```diff
@@ -1983,6 +2036,15 @@ define-command -hidden lsp-mappings-enable -params 1 -docstring "Add LSP mapping
 }
```
krobelus, line 2036: So I dropped this for now, looking to find a more general solution eventually.

I also found a problem with overloading `<c-n>` this way, so I've mapped `<tab>` as well, which seems nicer to use.
If you like `<c-n>` I'll keep it but otherwise I'd stick to `<tab>` only.

rc/lsp.kak
topisani: sounds good!

src/language_features/completion.rs
```diff
@@ -195,7 +195,10 @@ pub fn editor_completion(
                     static ref SNIPPET_TABSTOP_RE: Regex = Regex::new(r"\$(?P<i>\d+)").unwrap();
                     static ref SNIPPET_PLACEHOLDER_RE: Regex = Regex::new(r"\$\{(?P<i>\d+):?(?P<placeholder>[^}]+)\}").unwrap();
                 }
-                let snippet = insert_text;
+                let mut snippet = insert_text;
+                if !snippet.contains("$0") {
+                    snippet += "$0";
+                }
```
topisani, line 201: This is great for some rust-analyzer snippets as well

rc/lsp.kak
topisani: Mapping tab breaks my tab indent, and i assume it can break a lot of things for other people as well, so unless we find a way to make it completely transparent, i don't think it is a good idea as a default. `<c-n>` is easier, as i assume it is less likely to be in use for other things already.

krobelus: > Mapping tab breaks my tab indent

it indents just fine for me if there is no placeholder.
I assume you're seeing breakage because you created a mapping for tab?

`<c-n>` is a problematic default because we can't use it while the completion pager is active.

I think, whatever hotkey we end up using, we should tell users
to do something like this if they'd normally remap the key:

```kak
map global insert <tab> '<a-;>: tab-in-insert-mode<ret>'
define-command -hidden tab-in-insert-mode %{
    try %{
        lsp-snippets-select-next-placeholders
    } catch %{
        # do whatever tab normally does, the default is:
        execute-keys -with-hooks %arg{1}
    }
}
# similar for "normal" instead of "insert"
```

krobelus: I've merged all commits except for the one that adds the mappings. I'll try both mappings for a bit

rc/lsp.kak
krobelus: I think I finally (after like 2 years) realized what needs to happen

When the completion pager is active, a key like 'a' "accepts" the completion.
After the completion is accepted, InsertCompletionHide hooks run.
What we actually want is that InsertCompletionHide runs before 'a' is processed.
That'd allow us to insert snippets, select the first placeholder and have 'a' be inserted there.
Of course this would mean that `<ret>` is usually not what the user wants to accept a snippet, but that's how it should be. Kakoune completions are accepted implicitly once the user continues typing.

So we just need to hack Kakoune to run InsertCompletionHide earlier. We're probably the only users who'd notice this change.

topisani:
Yes, i use https://github.com/alexherbo2/indent.kak, which maps `<tab>` to `<a-;><gt>`. I could rewrite the mapping as you've shown above, but im not sure i think its a good idea for kak-lsp to break integration with other plugins like that.

> Unfortunately there is a problem with overloading <c-n>.  If the
completion pager is active, users expect it to cycle though the
completion candidates [*], so it can only be active if the completion
pager is not.  This means that to go to the next placeholder we need
"<esc><c-n>i".

Just for the record, I use `<c-o>` to exit the completion menu, meaning going to the next placeholder is just `<c-o><c-n>`, which is slightly better than exiting insert mode

rc/lsp.kak
krobelus: submitted as https://github.com/mawww/kakoune/pull/4632
people who used `<ret>` to accept snippet completions would need to change habits, or should we special case `<ret>`? :thinking:

krobelus:
You have a point - I didn't think of `<c-o>`.
I merely documented the two approaches for now. We should revisit once we have a good solution for adding mappings.

topisani:
Apparently github thinks i shouldnt be allowed to leave an approving review on my own pull request, so i'll do it like this instead:

I think this is a good approach for now, until we find a better way to map these. Maybe we can get some input when people actually use them.

Also, i think your kakoune patch is the exact right solution for that problem, that should help as well

alexherbo2:
We can also use a window mapping, for a given context.

It is used in [`auto-pairs.kak`]. See the `<ret>` and `<c-ret>` mappings.

[`auto-pairs.kak`]: https://github.com/alexherbo2/auto-pairs.kak/blob/master/rc/auto-pairs.kak

PR #617, opened 2022-05-29
Simplify completion hooks

krobelus:


krobelus:
@topisani I found a small simplification, not sure if it's even worth the churn (depends on future direction I guess)

topisani:
lgtm!

rc/lsp.kak
```diff
@@ -342,45 +342,32 @@ have_kakoune_feature_filtertext = ${kak_opt_lsp_have_kakoune_feature_filtertext}
...
-        # This is to avoid running them in the above try block, and silencing their errors.
-        remove-hooks window lsp-completion-dismissed
-        set-option window lsp_completions_selected_item -1
-    }
-    trigger-user-hook LSPCompletionDismissed
-    remove-hooks window lsp-completion-dismissed
-    set-option window lsp_completions_selected_item -1
-    set-option window lsp_completion_selection ""
+define-command -hidden lsp-completion-accepted -docstring "Called when a completion is accepted" %{
+    set-option window lsp_completion_inserted_range %val{hook_param}
```
alexherbo2, line 348: I didn’t know we could access `%val{hook_param}` in a command definition.

rc/lsp.kak
```diff
@@ -2057,7 +2044,7 @@ define-command lsp-enable -docstring "Default integration with kak-lsp" %{
     hook -group lsp global BufSetOption lsp_config=.* lsp-did-change-config
     hook -group lsp global BufSetOption lsp_server_configuration=.* lsp-did-change-config
     hook -group lsp global InsertIdle .* lsp-completion
-    hook -group lsp global InsertCompletionHide .* lsp-completion-dismissed
+    hook -group lsp global InsertCompletionHide .+ lsp-completion-accepted
```
alexherbo2, line 2047: Is the completion always accepted on hide?

I guess it can be due to the change from `.*` to `.+`, where there is no range when empty, but it is not obvious to translate as cancel when reading the code.

rc/lsp.kak
```diff
@@ -342,45 +342,32 @@ have_kakoune_feature_filtertext = ${kak_opt_lsp_have_kakoune_feature_filtertext}
...
+declare-option -hidden str lsp_completion_inserted_range
 
-define-command -hidden lsp-completion-dismissed -docstring "Called when the completion pager is closed" %{
-    set-option window lsp_completion_selection %val{hook_param}
-    try %{
-        evaluate-commands -draft %{
-            select %opt{lsp_completion_selection}
-        }
-    } catch %{
-        # Completion was cancelled, nothing was inserted
```
alexherbo2, line 354: If I understand correctly, you moved the logic in the regex to trigger the hook with a `.+` to ensure the completion is accepted.

It could be worth to readd comments to explain it.

rc/lsp.kak
krobelus: yeah that's a bit strange

rc/lsp.kak
krobelus: Yep. An empty hook parameter means no competion candidate was inserted.
The documentation of InsertCompletionHide does not state this explicitly, we should probably improve that.
Added a comment.

rc/lsp.kak
```diff
@@ -342,18 +342,18 @@ have_kakoune_feature_filtertext = ${kak_opt_lsp_have_kakoune_feature_filtertext}
 " | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }}
 
-declare-option -hidden str lsp_completion_inserted_range
+declare-option -hidden str lsp_completion_inserted_ranges
```
alexherbo2, line 345: Why not just change the type to `str-list` and use `eval` when setting the option from `%val{hook_param}`.

krobelus: great catch, thanks!

Issue #490, opened 2021-06-01
Support code lenses


expipiplus1:
see: https://vscode.rocks/codelens/

It would be great to be able to use code-lenses in kakoune!

In nvim they're rendered as "virtual text" hanging off the end of lines.

![image](https://user-images.githubusercontent.com/857308/120261309-3ac36b80-c2ca-11eb-9bd5-e4b059c58cff.png)

expipiplus1:
See here for a discussion on virtual text: https://github.com/mawww/kakoune/issues/1813

krobelus:
added as of ab07f913 (Support LSP code lenses, 2022-04-28)

expipiplus1:
Wow, thanks!

Rotaerk:
This doesn't seem to work.  Using v12.2.0, when I run the `lsp-code-lens` command (or type `,ll`), with the cursor on the line marked by `>`, I get an empty `lsp-menu-select` menu.

This used to work.  On April 28th, the head of your code-lens branch was commit d7fb3b56a85c21299838c414d2379fa6ad28ed1c.  (This doesn't appear to be in your branch history, but it can be directly checked out.)  While the unicode gutter character you had at that time caused issues, the `lsp-code-lens` command itself worked.

krobelus:
My bad I should have added a test. Fixed as of ed1fb46a7ce0821e11ad96da31b12b917dbdb245

Rotaerk:
Yep, that works for me.  Thanks for the quick fix!

PR #623, opened 2022-07-02
Feature: Code lenses as inlay hints

sidkshatriya:
### Motivation

kak-lsp added support for code-lenses relatively recently (See #490). Briefly speaking, the code-lense feature allows you to trigger an action relevant to a particular line of code or otherwise see meta information specific to a source code line sent by the LSP.

#### Current Status in Kakoune
If a line has a code lens, it receives an annotation of `>` besides the line number in kakoune. Then you can use `<lsp-mode> l` to see available code-lenses.

While this works, you explicitly need to _ask_ kakoune for the code lens. Also, many times code lenses are not actions really but (meta) information that language servers want to add to the code. Example of code lens would be the type signature of a function e.g. in ocaml-lsp-server or how many times a function is used (in erlang_ls). In Haskell Language Server, code lenses show information around module imports and so on.

In vscode for instance, ocaml-lsp-server shows the function signatures via code-lens. Example screenshot below (**see the code lens above source lines 420, 425, 433 and so on in the slightly smaller font**):

### What does this PR do?
Currently only the _existence_ of a code lens is known via the availability of a `>`. It would be great if the code lens was actually shown (similar to vscode) _inline with_ the rest of the source code. 

In Kakoune, this would be an inlay hint. Just like types are shown as inlay-hints. We can show code-lenses as inlay hints. **This PR accomplishes that**.

**This is the final result. See the code-lens inlay hints inserted on the lines annotated with `>` -- these are the inferred type signatures `[🔎 ... ]`.**

I personally find this feature quite useful. In languages with Hindley-Milner type inferencing like OCaml you (mostly) don't need to annotate functions with types. This is a double edged sword. It would be useful for us to be able to see the inferred types in Kakoune.

This is useful in a lot of other LSPs also. I can easily see what information and actions are available -- without having to explicitly list them.

### Meta

This PR is currently quite basic and gets the job done. If this is deemed a useful feature, I'll be happy to work on the minimum fixes/adjustments required by the maintainer to merge to master. Others can build upon this to add more functionality in subsequent commits. Please let me know what is minimally required.

### Issues

~~If the inlay-hint is big, you can run into mawww/kakoune#4659. However this is a limitation of kakoune and not related to code-lenses or this PR. In fact the same problem arises with long length inferred types inserted as inlay hints.~~

Edit: mawww/kakoune#4659 has received a fix.

In closing, I have found this feature useful in many LSPs -- it's good to see the code-lenses inline with code.

### DIFF
--- a/src/language_features/code_lens.rs
+++ b/src/language_features/code_lens.rs
@@ -1,5 +1,4 @@
 use super::code_action::execute_command_editor_command;
-use crate::capabilities::server_has_capability;
 use crate::capabilities::CAPABILITY_CODE_LENS;
 use crate::capabilities::CAPABILITY_EXECUTE_COMMANDS;
 use crate::context::*;
@@ -7,6 +6,8 @@ use crate::diagnostics::gather_line_flags;
 use crate::position::*;
 use crate::types::*;
 use crate::util::editor_quote;
+use crate::util::escape_tuple_element;
+use crate::{capabilities::server_has_capability, markup::escape_kakoune_markup};
 use itertools::Itertools;
 use lsp_types::request::*;
 use lsp_types::*;
@@ -27,8 +28,52 @@ pub fn text_document_code_lens(meta: EditorMeta, ctx: &mut Context) {
         partial_result_params: PartialResultParams::default(),
     };
     ctx.call::<CodeLensRequest, _>(meta, req_params, |ctx: &mut Context, meta, result| {
-        editor_code_lens(meta, result, ctx)
+        editor_code_lens(meta.clone(), result.clone(), ctx);
+        inlay_hints_from_editor_code_lens(meta, result.unwrap_or_default(), ctx)
     });
+}
+
+fn inlay_hints_from_editor_code_lens(
+    meta: EditorMeta,
+    mut lenses: Vec<CodeLens>,
+    ctx: &mut Context,
+) {
+    let document = match ctx.documents.get(&meta.buffile) {
+        Some(document) => document,
+        None => return,
+    };
+    lenses.sort_by_key(|lens| lens.range.start);
+
+    let inlay_hints_from_lenses = lenses
+        .iter()
+        .map(|lens| {
+            let label = lens.command.as_ref().map_or("", |v| &v.title);
+            let position =
+                lsp_position_to_kakoune(&lens.range.start, &document.text, ctx.offset_encoding);
+            let line = position.line;
+            let col = position.column;
+            let icon = if meta.has_bad_emoji_wcwidth {
+                "[C]"
+            } else {
+                "🔎"
+            };
+
+            editor_quote(&format!(
+                "{line}.{col}+0|{{InlayHint}}[{icon} {}] ",
+                escape_tuple_element(&escape_kakoune_markup(label))
+            ))
+        })
+        .join(" ");
+    let command = format!(
+        "set-option buffer lsp_inlay_hints {} {}",
+        meta.version, inlay_hints_from_lenses
+    );
+    let command = format!(
+        "evaluate-commands -buffer {} -verbatim -- {}",
+        editor_quote(&meta.buffile),
+        command
+    );
+    ctx.exec(meta, command);
 }
 
 fn editor_code_lens(meta: EditorMeta, result: Option<Vec<CodeLens>>, ctx: &mut Context) {
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -164,6 +164,7 @@ define-command -hidden lsp-show-code-actions -params 1.. -docstring "Called when
 # Work around terminals/libc using different versions of wcwidth(3).
 declare-option -hidden int lsp_emoji_lightbulb_ok
 declare-option -hidden int lsp_emoji_hourglass_ok
+declare-option -hidden bool lsp_has_bad_emoji_wcwidth true
 evaluate-commands -save-regs t %{
     set-register t %{
         edit -scratch *lsp-scratch*
@@ -179,6 +180,14 @@ evaluate-commands -save-regs t %{
     } catch %{
         evaluate-commands %reg{t}
     }
+}
+
+set-option global lsp_has_bad_emoji_wcwidth %sh{
+    if [ $kak_opt_lsp_emoji_lightbulb_ok = 3 ]; then
+        echo "false"
+    else
+        echo "true"
+    fi
 }
 
 define-command -hidden lsp-hide-code-actions -docstring "Called when no code action is available for the main cursor position" %{
@@ -286,6 +295,7 @@ buffile  = \"${kak_buffile}\"
 filetype = \"${kak_opt_filetype}\"
 version  = ${kak_timestamp:-0}
 method   = \"textDocument/didChange\"
+has_bad_emoji_wcwidth = ${kak_opt_lsp_has_bad_emoji_wcwidth}
 [params]
 draft    = \"\"\"
 ${lsp_draft}\"\"\"
@@ -332,6 +342,7 @@ buffile  = \"${kak_buffile}\"
 filetype = \"${kak_opt_filetype}\"
 version  = ${kak_timestamp:-0}
 method   = \"textDocument/completion\"
+has_bad_emoji_wcwidth = ${kak_opt_lsp_has_bad_emoji_wcwidth}
 [params.position]
 line     = ${kak_cursor_line}
 column   = ${kak_cursor_column}
@@ -385,6 +396,7 @@ buffile  = \"${kak_buffile}\"
 filetype = \"${kak_opt_filetype}\"
 version  = ${kak_timestamp:-0}
 method   = \"completionItem/resolve\"
+has_bad_emoji_wcwidth = ${kak_opt_lsp_has_bad_emoji_wcwidth}
 [params]
 completion_item_index = ${index}
 pager_active = ${1}
@@ -429,6 +441,7 @@ buffile  = \"${kak_buffile}\"
 filetype = \"${kak_opt_filetype}\"
 version  = ${kak_timestamp:-0}
 method   = \"textDocument/hover\"
+has_bad_emoji_wcwidth = ${kak_opt_lsp_has_bad_emoji_wcwidth}
 ${kak_opt_lsp_connect_fifo}\
 [params]
 $hover_buffer_args
@@ -488,6 +501,7 @@ buffile  = \"${kak_buffile}\"
 filetype = \"${kak_opt_filetype}\"
 version  = ${kak_timestamp:-0}
 method   = \"kak-lsp/next-or-previous-symbol\"
+has_bad_emoji_wcwidth = ${kak_opt_lsp_has_bad_emoji_wcwidth}
 [params]
 position.line   = ${kak_cursor_line}
 position.column = ${kak_cursor_column}
@@ -538,6 +552,7 @@ buffile  = \"${kak_buffile}\"
 filetype = \"${kak_opt_filetype}\"
 version  = ${kak_timestamp:-0}
 method   = \"kak-lsp/object\"
+has_bad_emoji_wcwidth = ${kak_opt_lsp_has_bad_emoji_wcwidth}
 [params]
 count           = $kak_count
 mode            = \"$kak_opt_lsp_object_mode\"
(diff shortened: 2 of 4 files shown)


sidkshatriya:
Changed the code lens inlay hint to use the emoji 🔎  to make it more clear what the code lens is. Updated the kakoune screenshot above.

sidkshatriya:
You might be wondering, why are the code-lens on the same line? Wouldn't it be better to put on previous line?

Actually I tried putting it on the previous line -- The code lenses can quickly become confusing when you have functions that are not separated by a blank line. You have a situation where the code lens for a function is on the last line of the previous function. 

There is another serious issue when the code-lens will simply vanish if last line was blank _and_ the next line has some indentation. Kakoune does not allow applying an inlay hint on a position that does not exist: since the last line is blank you cannot put a code-lens at some indentation level.

*TL;DR code lens should be on the same line*

krobelus:
> *TL;DR code lens should be on the same line*

for now, yes. Once replace-ranges can insert new lines (or at least insert text past EOL) then we can move them to the last line

src/language_features/code_lens.rs
```diff
@@ -32,10 +32,54 @@ pub fn text_document_code_lens(meta: EditorMeta, ctx: &mut Context) {
         partial_result_params: PartialResultParams::default(),
     };
     ctx.call::<CodeLensRequest, _>(meta, req_params, |ctx: &mut Context, meta, result| {
-        editor_code_lens(meta, result, ctx)
+        editor_code_lens(meta.clone(), result.clone(), ctx);
+        inlay_hints_from_editor_code_lens(meta, result.unwrap_or_default(), ctx)
     });
 }
 
+fn inlay_hints_from_editor_code_lens(
```
krobelus, line 40: Great idea. I forgot to document why I didn't do this yet: it's just because I hadn't used a language server where always-visible code lenses are useful.

(optional) I wonder if this should be enabled by default. I'd definitely vote for that if replace-ranges was less buggy.

src/language_features/code_lens.rs
```diff
@@ -32,10 +32,54 @@ pub fn text_document_code_lens(meta: EditorMeta, ctx: &mut Context) {
...
+                    "{line}.{col}+0|{{InlayHint}}[\u{1F50E} {}] {{InlayHint}}",
+                    escape_tuple_element(label)
+                ))
+            } else {
+                String::new()
+            }
+        })
+        .join(" ");
+    let command = format!(
+        "set-option buffer lsp_inlay_hints {} {}",
```
krobelus, line 72: I believe LSP servers may send both `textDocument/inlayHint` and `textDocument/codeLens`, and the two are not necessarily related.
To allow this, I think we should give code lenses a dedicated replace-ranges highlighter with documented commands like `lsp-inlay-code-lenses-enable` (and use the name "inlay hints" only for code related to `textDocument/inlayHint`).

src/language_features/code_lens.rs
```diff
@@ -32,10 +32,54 @@ pub fn text_document_code_lens(meta: EditorMeta, ctx: &mut Context) {
...
+        meta.version, inlay_hints_from_lenses
```
krobelus, line 73: (optional) what do you think about disabling all virtual text (from inlay hints or code lenses) while in insert mode, like we already do for inlay diagnostics?

src/language_features/code_lens.rs
```diff
@@ -32,10 +32,54 @@ pub fn text_document_code_lens(meta: EditorMeta, ctx: &mut Context) {
...
 
+fn inlay_hints_from_editor_code_lens(
+    meta: EditorMeta,
+    mut lenses: Vec<CodeLens>,
+    ctx: &mut Context,
+) {
+    let document = match ctx.documents.get(&meta.buffile) {
+        Some(document) => document,
+        None => {
+            return;
```
krobelus, line 48: can you drop the braces to be consistent with existing code?

src/language_features/code_lens.rs
```diff
@@ -32,10 +32,54 @@ pub fn text_document_code_lens(meta: EditorMeta, ctx: &mut Context) {
...
+        .iter()
+        .map(|lens| {
+            let label = lens.command.as_ref().map_or("", |v| &v.title);
+            if label.trim().len() > 0 {
+                let position =
+                    lsp_position_to_kakoune(&lens.range.start, &document.text, ctx.offset_encoding);
+                let line = position.line;
+                let col = position.column;
+                editor_quote(&format!(
+                    "{line}.{col}+0|{{InlayHint}}[\u{1F50E} {}] {{InlayHint}}",
```
krobelus, line 63: I like the emoji :)
if we added a test, we would probably see it fail in macOS CI because their glibc uses (?) an old version of Unicode.
This means macOS users might see glitches when we use emoji.
Our workaround is 73c2d1ab (Work around inconsistent wcwidth by using ASCII instead of emoji, 2022-01-21).
It should be fairly easy to extend this to this emoji (look for `lsp_emoji_lightbulb_ok`). Although we should probably document the workaround better.

krobelus, line 63: I think it's better to use the literal emoji, for better greppability

src/language_features/code_lens.rs
```diff
@@ -32,10 +32,54 @@ pub fn text_document_code_lens(meta: EditorMeta, ctx: &mut Context) {
...
+            return;
+        }
+    };
+    lenses.sort_by_key(|lens| lens.range.start);
+
+    let inlay_hints_from_lenses = lenses
+        .iter()
+        .map(|lens| {
+            let label = lens.command.as_ref().map_or("", |v| &v.title);
+            if label.trim().len() > 0 {
```
krobelus, line 57: clippy demands we use `is_empty()` but actually I think it's better to use a `filter()` before the `map()`

TBH I think an empty title is probably a bug in the language server, and I don't expect that special-casing it is useful. Happy to proven wrong of course, is there any server that sends empty titles?

Of course we do want to filter away cases where the command is not set (because it's not resolved yet; we only resolve upon user action). Below we do this with

        .filter(|lens| lens.command.is_none())

so the most obvious code here would be

        .filter(|lens| lens.command.is_some())

src/language_features/code_lens.rs
```diff
@@ -32,10 +32,54 @@ pub fn text_document_code_lens(meta: EditorMeta, ctx: &mut Context) {
         partial_result_params: PartialResultParams::default(),
     };
     ctx.call::<CodeLensRequest, _>(meta, req_params, |ctx: &mut Context, meta, result| {
-        editor_code_lens(meta, result, ctx)
+        editor_code_lens(meta.clone(), result.clone(), ctx);
+        inlay_hints_from_editor_code_lens(meta, result.unwrap_or_default(), ctx)
```
krobelus, line 36: If we merge the additions into `editor_code_lens`, we get rid of some boilerplate and the clones here

src/language_features/code_lens.rs
krobelus, line 63: I don't think the second `{{InlayHint}}` does anything, better remove it

krobelus, line 63: You escape special tuple-characters (`|` and backslash) in `label` but we also need to escape `{`,
because it's interpreted as markup string.

There are two ways to do that (we still use both but should converge eventually).

`:doc faces` says

> The {\} string disables markup processing for the rest of the line, and can be used to avoid having to escape text that might be mistaken for markup instructions.

So we could end the string with `{{\\}} {}]`,

or add a `.replace('{', "\\{")`. This is what the inlay-hints implementation does.
Unfortunately we can't use `escape_kakoune_markup` because that would escape backslashes too much. That's a limitation in Kakoune, I'll raise an issue later.

src/language_features/code_lens.rs
sidkshatriya: Editors like vscode I think choose to show all the code-lens always. 

However due to https://github.com/mawww/kakoune/issues/4659 the experience in kakoune is not always ideal. However, in-spite of that, I find it useful to keep this on for my own editing always. If it were not for that issue, I would definitely show these always. 

Anyways I leave it to you to take a call!

src/language_features/code_lens.rs
sidkshatriya: Virtual text should be _on_ during insert mode I think. 

While I'm doing the actual code editing I actually want to know the types even more...

src/language_features/code_lens.rs
sidkshatriya: Ok, I will try to incorporate the workaround.

src/language_features/code_lens.rs
sidkshatriya: I'm usually in favor of keeping _some_ boilerplate in favor of increasing code readability and separation of concerns. It's not really going to increase the speed of the LSP as io would probably dominate everything.

Let's leave it the same for now while this PR works itself towards completion. You can take a call again later.

src/language_features/code_lens.rs
sidkshatriya: Will do

src/language_features/code_lens.rs
sidkshatriya: Will do

sidkshatriya: Will do

src/language_features/code_lens.rs
sidkshatriya: Interesting point. I will explore this issue a bit more.

krobelus: I guess the alternative to my suggestion is to implement some kind of merging logic (like we do for our flag-lines highlighter, which is a combination of diagnostics and code lenses).

The question is if we want to expose a single option for enabling both, or separate options.
I'm fine with either direction.
At first, I was a bit apprehensive of using the term "inlay hints" for an umbrella option but it sounds like a reasonable choice.

sidkshatriya:
I plan to address the feedback items by Monday (July 11) !

src/language_features/code_lens.rs
krobelus: Long-term we might have to split them up.
But I think today there is no server that supports both inlay hints and code lenses (though I haven't checked at all), so making `lsp-inlay-hints-enable` activate both might just be fine..

krobelus: (I think the current version of this PR breaks rust-analyzer's inlay hints, both approaches would fix that).

sidkshatriya:
(Sorry I will definitely deal with this over the next couple of days as soon as I get some spare cycles!)

src/language_features/code_lens.rs
sidkshatriya: Done!

src/language_features/code_lens.rs
sidkshatriya: Done!

src/language_features/code_lens.rs
sidkshatriya: Removed the whole `if label.trim().len() > 0` business. Assumption: if there is a code lens sent by the server, there would always be a title. Simplifies the logic and I can't think of scenarios where we would have a title but not a code lens.

src/language_features/code_lens.rs
sidkshatriya: This TODO is still pending. Have a look at some other feedback items that I've addressed that I've made in the meanwhile. I will address this point in my next round of changes.

krobelus: Looks fine. The emoji thing is ugly but I'm sure we can remove that workaround eventually (not sure if it's necessary for current macOS).

sidkshatriya:
I plan to prepare the (hopefully) final set of changes by coming Sunday. I'll also resolve the conflicts.

src/language_features/code_lens.rs
krobelus: > As emoji's become more important generally, it would be useful to always know if we can sent emoji's or not.

I agree but instead of adding an request parameter (which won't ever change) we should just move the awkward computation of wcwidth (see `*lsp-scratch*`) into the controller code. From Rust we can easily call `wcwidth` and then just set the options during initialization. Note that it's easy to simulate the bad wcwidth with `LC_ALL=C` (though such a non-UTF-8 locale might cause other issue, so it may not be worth supporting).

We can always use the emoji for now and leave the wcwidth-probing to a follow up (lsp-inlay-code-lenses are an optional feature after all)

PR #630, opened 2022-07-21
Fix blocking request hanging if unconfigured

Spferical:
To reproduce, run `lsp-formatting-sync` on a file without a configured lsp server.

I think this works for all the blocking commands I see in `lsp.kak`. I'm not sure what's appropriate for `write_response_to_fifo = true` calls, but I figure this is better than deadlock.

I'd like to add more visibility to other error conditions, but if a `lsp-show-error` triggers on _every_ request that fails due to a missing lsp configuration, it could bother auto hover users. Maybe that could be solved with a smarter auto hover implementation?

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -80,6 +80,14 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
                         "Language server is not configured for filetype `{}`",
                         &request.meta.filetype
                     );
+                    // If the editor is expecting a fifo response, give it one, so it won't hang.
+                    if let Some(ref fifo) = request.meta.fifo {
+                        let command = format!(
+                            "lsp-show-error 'Language server is not configured for filetype `{}`'",
+                            request.meta.filetype
+                        );
+                        std::fs::write(fifo, &*command).expect("Failed to write command to fifo");
+                    }
                     continue 'event_loop;
                 }
                 let language_id = language_id.unwrap();


krobelus:
Nice fix, thanks!

To reproduce, run `lsp-formatting-sync` on a file without a configured lsp server.

I think this works for all the blocking commands I see in `lsp.kak`.

> I'm not sure what's appropriate for `write_response_to_fifo = true` calls

Your fix works for the `<c-o>` lsp mode binding, so it's good.
I'm thinking of removing the `write_response_to_fifo` feature altogether. It's not essential and blocking requests add lots of room for failure.

> I'd like to add more visibility to other error conditions, but if a `lsp-show-error` triggers on _every_ request that fails due to a missing lsp configuration, it could bother auto hover users. Maybe that could be solved with a smarter auto hover implementation?

Yeah I think we try to be forgiving of promiscuous hooks.
We could add a request parameter that tells us whether the request was sent from a hook.

You're right though, requests that are not sent from a hook should error with `lsp-show-error`
Same for unsupported requests, which currently never run `lsp-show-error`, see:

- b29989a3 (Send documentHighlight/hover/signatureHelp requests only if supported by the server, 2022-07-06)
- 8660d5ab (Send code action requests only if supported by the server, 2022-04-29)
- 156e1679 (Teach controller to not send unsupported requests to the server, 2022-07-10)

src/session.rs
```diff
@@ -80,6 +80,14 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
                         "Language server is not configured for filetype `{}`",
                         &request.meta.filetype
                     );
+                    // If the editor is expecting a fifo response, give it one, so it won't hang.
+                    if let Some(ref fifo) = request.meta.fifo {
+                        let command = format!(
+                            "lsp-show-error 'Language server is not configured for filetype `{}`'",
+                            request.meta.filetype
+                        );
+                        std::fs::write(fifo, &*command).expect("Failed to write command to fifo");
```
krobelus, line 89: Looks like we don't need the &* (at least with Rust 1.59)
In a couple other places where we pass "command.as_bytes()" to fs::write, which also doesn't seem necessary.

I guess this is because `String` implements `AsRef<[u8]>`, so maybe we should exploit that?

Issue #628, opened 2022-07-20
Instructions for building on Cygwin (if possible)


whereswaldon:
Kakoune itself supports running on Cygwin, but it isn't clear whether kak-lsp is expected to work there also or not. When I tried to build it, I got:

<details>
<summary>error log</summary>
<pre>
Cloning into 'kak-lsp'...
  Installing kak-lsp v12.2.1-snapshot (C:\cygwin64\home\chris\Code\builds\configuration\kak\plugins\kak-lsp)
    Updating crates.io index
 Downloading crates ...
  Downloaded dirs-sys-next v0.1.2
  Downloaded ryu v1.0.9
  Downloaded version_check v0.9.4
  Downloaded trackable_derive v1.0.0
  Downloaded rand_chacha v0.3.1
  Downloaded str_indices v0.3.2
  Downloaded slog-stdlog v4.1.1
  Downloaded slog-term v2.9.0
  Downloaded winapi v0.3.9
  Downloaded time-macros v0.2.4
  Downloaded semver-parser v0.10.2
  Downloaded rustversion v1.0.6
  Downloaded rle-decode-fast v1.0.3
  Downloaded ropey v1.4.1
  Downloaded tinyvec_macros v0.1.0
  Downloaded serde v1.0.136
  Downloaded take_mut v0.2.2
  Downloaded strsim v0.8.0
  Downloaded slog-async v2.7.0
  Downloaded slab v0.4.5
  Downloaded textwrap v0.11.0
  Downloaded thread_local v1.1.4
  Downloaded time v0.1.43
  Downloaded unindent v0.1.8
  Downloaded whoami v1.2.1
  Downloaded tinyvec v1.5.1
  Downloaded unicode-xid v0.2.2
  Downloaded widestring v0.4.3
  Downloaded unicode-bidi v0.3.7
  Downloaded serde_json v1.0.79
  Downloaded url v2.2.2
  Downloaded aho-corasick v0.7.18
  Downloaded trackable v1.2.0
  Downloaded regex-syntax v0.6.25
  Downloaded daemonize v0.4.1
  Downloaded enum_primitive v0.1.1
  Downloaded rand v0.8.5
  Downloaded lsp-types v0.92.1
  Downloaded indoc v1.0.4
  Downloaded dirs v2.0.2
  Downloaded diffs v0.4.1
  Downloaded dirs-sys v0.3.7
  Downloaded boxfnonce v0.1.1
[...]

PR #633, opened 2022-07-29
Show more errors in the editor

Spferical:
Hey! I rolled with [the idea of reporting errors for only calls that happen outside of hook context](https://github.com/kak-lsp/kak-lsp/pull/630#issuecomment-1191106482), and I think it works pretty well so far. I haven't gotten to unsupported requests yet, but I added the error cases I ran into from bad config and flaky lsp servers. I also removed some panics after I brought down `kak-lsp` a few times.

The hook detection relies on a bit of a hack. From what I can tell, [`%val{hook_param}` is always set when running in a hook, though it's empty for hooks that don't have a hook_param](https://github.com/mawww/kakoune/blob/e83dbdcd2cbddb30ac63ec503c76b46b80ffe44d/src/hook_manager.cc#L114). I was thinking of asking for a more official way to detect if we're running in a hook or not in kakoune after this.

Some ideas from while I was working on this:
- It's awkward if someone installs a new kak-lsp that gets started in a kakoune session that's still running outdated kak-side code; after this PR that would make it print errors for all commands, even in a hook. I'm guessing this isn't supported, but it could be useful in general to warn the user if they update kak-lsp swithout restarting kakoune?
- It's easy to write sh code that can mess up request toml formatting in  `lsp.kak`. e.g. try `:e foo."'"` and see the requests fail to parse. Has anyone considered moving some or all of the env parsing logic into `kak-lsp --request` itself? (This might require a more complicated cli, with different cli commands for different operations... But shell escapes are also tricky.)

### DIFF
--- a/src/editor_transport.rs
+++ b/src/editor_transport.rs
@@ -127,8 +127,13 @@ pub fn start_unix(path: &path::Path, sender: Sender<EditorRequest>) {
                             continue;
                         }
                         debug!("From editor: {}", request);
-                        let request: EditorRequest =
-                            toml::from_str(&request).expect("Failed to parse editor request");
+                        let request: EditorRequest = match toml::from_str(&request) {
+                            Ok(req) => req,
+                            Err(err) => {
+                                error!("Failed to parse editor request: {}", err);
+                                continue;
+                            }
+                        };
                         if sender.send(request).is_err() {
                             return;
                         };
--- a/src/editor_transport.rs
+++ b/src/editor_transport.rs
@@ -12,7 +12,7 @@ use std::process::{Command, Stdio};
 pub struct EditorTransport {
     // Not using Worker here as listener blocks forever and joining its thread
     // would block kak-lsp from exiting.
-    pub from_editor: Receiver<EditorRequest>,
+    pub from_editor: Receiver<String>,
     pub to_editor: Worker<EditorResponse, Void>,
 }
 
@@ -39,8 +39,6 @@ pub fn start(session: &str, initial_request: Option<String>) -> Result<EditorTra
     }
     std::thread::spawn(move || {
         if let Some(initial_request) = initial_request {
-            let initial_request: EditorRequest =
-                toml::from_str(&initial_request).expect("Failed to parse initial request");
             if sender.send(initial_request).is_err() {
                 return;
             };
@@ -108,7 +106,7 @@ pub fn start(session: &str, initial_request: Option<String>) -> Result<EditorTra
     })
 }
 
-pub fn start_unix(path: &path::Path, sender: Sender<EditorRequest>) {
+pub fn start_unix(path: &path::Path, sender: Sender<String>) {
     let listener = match UnixListener::bind(&path) {
         Ok(listener) => listener,
         Err(e) => {
@@ -127,13 +125,6 @@ pub fn start_unix(path: &path::Path, sender: Sender<EditorRequest>) {
                             continue;
                         }
                         debug!("From editor: {}", request);
-                        let request: EditorRequest = match toml::from_str(&request) {
-                            Ok(req) => req,
-                            Err(err) => {
-                                error!("Failed to parse editor request: {}", err);
-                                continue;
-                            }
-                        };
                         if sender.send(request).is_err() {
                             return;
                         };
--- a/src/session.rs
+++ b/src/session.rs
@@ -1,3 +1,4 @@
+use crate::context::meta_for_session;
 use crate::controller;
 use crate::editor_transport;
 use crate::project_root::find_project_root;
@@ -7,6 +8,7 @@ use crate::util::*;
 use crossbeam_channel::{after, never, select, Sender};
 use lsp_types::notification::Notification;
 use lsp_types::*;
+use regex::Regex;
 use std::collections::HashMap;
 use std::time::Duration;
 
@@ -63,6 +65,18 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
                 // should be safe to unwrap as we just checked request for being None
                 // done this way instead of `match` to reduce nesting
                 let request = request.unwrap();
+                let request: EditorRequest = match toml::from_str(&request) {
+                    Ok(req) => req,
+                    Err(err) => {
+                        error!("Failed to parse editor request: {}", err);
+                        handle_broken_editor_request(
+                            editor.to_editor.sender(),
+                            request,
+                            &config.server.session
+                        );
+                        continue 'event_loop;
+                    }
+                };
                 // editor explicitely asked us to stop kak-lsp session
                 // (and we stop, even if other editor sessions are using this kak-lsp session)
                 if request.method == "stop" {
@@ -131,6 +145,40 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
     }
     stop_session(&mut controllers);
     0
+}
+
+/// Tries to send an error to the client about a request that failed to parse.
+fn handle_broken_editor_request(
+    to_editor: &Sender<EditorResponse>,
+    request: String,
+    session: &str,
+) {
+    // Try to parse enough of the broken toml to send the error to the editor.
+    lazy_static::lazy_static! {
+        static ref CLIENT_RE: Regex = Regex::new(r#"(?m)^client *= *"([a-zA-Z0-9_-]*)""#)
+            .expect("Failed to parse client name regex");
+        static ref HOOK_RE: Regex = Regex::new(r"(?m)^hook *= *true")
+            .expect("Failed to parse hook regex");
+    }
+    if let Some(client_name) = CLIENT_RE
+        .captures(&request)
+        .and_then(|cap| cap.get(1))
+        .map(|cap| cap.as_str())
+    {
+        // We still don't want to spam the user if a hook triggered the error.
+        if !HOOK_RE.is_match(&request) {
+            let msg = "Failed to parse editor request";
+            let meta = meta_for_session(session.to_string(), Some(client_name.to_string()));
+            let command = format!("lsp-show-error {}", editor_quote(msg));
+            let response = EditorResponse {
+                meta,
+                command: command.into(),
+            };
+            if let Err(err) = to_editor.send(response) {
+                error!("Failed to send error message to editor: {err}");
+            };
+        }
+    }
 }
 
 /// Sends an error back to the editor.
--- a/src/session.rs
+++ b/src/session.rs
@@ -76,30 +76,12 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
 
                 let language_id = filetypes.get(&request.meta.filetype);
                 if language_id.is_none() {
-                    debug!(
+                    let msg = format!(
                         "Language server is not configured for filetype `{}`",
                         &request.meta.filetype
                     );
-
-                    let command = format!(
-                        "lsp-show-error 'Language server is not configured for filetype `{}`'",
-                        request.meta.filetype
-                    );
-
-                    // If the editor is expecting a fifo response, give it one, so it won't hang.
-                    if let Some(ref fifo) = request.meta.fifo {
-                        std::fs::write(fifo, &command).expect("Failed to write command to fifo");
-                    }
-
-                    if !request.meta.hook {
-                        let response = EditorResponse {
-                            meta: request.meta.clone(),
-                            command: command.into(),
-                        };
-                        if let Err(err) = editor.to_editor.sender().send(response) {
-                            error!("Failed to send stop message to editor: {err}");
-                        };
-                    }
+                    debug!("{}", msg);
+                    return_request_error(editor.to_editor.sender(), &request, &msg);
 
                     continue 'event_loop;
                 }
@@ -118,11 +100,13 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
                 match controllers.entry(route.clone()) {
                     Entry::Occupied(controller_entry) => {
                         if let Err(err) = controller_entry.get().worker.sender().send(request.clone())  {
-                            if let Some(fifo) = request.meta.fifo {
-                                cancel_blocking_request(fifo);
-                            }
-                            controller_entry.remove();
                             error!("Failed to send message to controller: {}", err);
+                            return_request_error(
+                                editor.to_editor.sender(),
+                                &request,
+                                "Language server is no longer running"
+                            );
+                            controller_entry.remove();
                             continue 'event_loop;
                         }
                     }
@@ -149,14 +133,27 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
     0
 }
 
-/// When server is not running it's better to cancel blocking request.
-/// Because server can take a long time to initialize or can fail to start.
-/// We assume that it's less annoying for user to just repeat command later
-/// than to wait, cancel, and repeat.
-fn cancel_blocking_request(fifo: String) {
-    debug!("Blocking request but LSP server is not running");
-    let command = "lsp-show-error 'language server is not running, cancelling blocking request'";
-    std::fs::write(fifo, command).expect("Failed to write command to fifo");
+/// Sends an error back to the editor.
+///
+/// This will cancel any blocking requests and also print an error if the
+/// request was not triggered by an editor hook.
+fn return_request_error(to_editor: &Sender<EditorResponse>, request: &EditorRequest, msg: &str) {
+    let command = format!("lsp-show-error {}", editor_quote(msg));
+
+    // If editor is expecting a fifo response, give it one, so it won't hang.
+    if let Some(ref fifo) = request.meta.fifo {
+        std::fs::write(fifo, &command).expect("Failed to write command to fifo");
+    }
+
+    if !request.meta.hook {
+        let response = EditorResponse {
+            meta: request.meta.clone(),
+            command: command.into(),
+        };
+        if let Err(err) = to_editor.send(response) {
+            error!("Failed to send error message to editor: {err}");
+        };
+    }
 }
 
 /// Reap controllers associated with editor session.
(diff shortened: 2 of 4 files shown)


src/controller.rs
```diff
@@ -44,27 +44,25 @@ pub fn start(
...
-                    .send(EditorResponse {
-                        meta: initial_request.meta,
-                        command: Cow::from(command),
-                    })
-                    .is_err()
-                {
-                    error!("Failed to send command to editor");
+                error!("failed to start language server: {}", err);
+                // If the server command isn't from a hook (e.g. auto-hover),
+                // then send a prominent error to the editor.
```
krobelus, line 49: Very cool. We couldn't show an error before because we want to support a global `lsp-enable`.
Now that we can track hooks we can better capture the user's intent.

src/controller.rs
```diff
@@ -44,27 +44,25 @@ pub fn start(
         lang_srv = match language_server_transport::start(&lang.command, &lang.args, &lang.envs) {
             Ok(ls) => ls,
             Err(err) => {
-                // If we think that the server command is not from the default config, then we
-                // send a prominent error to the editor, since it's likely configuration error.
-                let might_be_from_default_config =
-                    !lang.command.contains('/') && !lang.command.contains(' ');
```
krobelus, line 50: Makes sense. Running a single lsp command in a buffer without an installed language server shouldn't affect other language servers.
Especially since we already tell the user via the info box.
I wonder if we have more bugs like that.

src/editor_transport.rs
```diff
@@ -127,8 +127,13 @@ pub fn start_unix(path: &path::Path, sender: Sender<EditorRequest>) {
                             continue;
                         }
                         debug!("From editor: {}", request);
-                        let request: EditorRequest =
-                            toml::from_str(&request).expect("Failed to parse editor request");
+                        let request: EditorRequest = match toml::from_str(&request) {
+                            Ok(req) => req,
+                            Err(err) => {
+                                error!("Failed to parse editor request: {}", err);
```
krobelus, line 133: I'm not convinced this is an improvement.
Unlike in the previous commit, the user has to actively create a bad configuration.
I feel like it's better to fail noisily and force them to fix it instead of ignoring the configuration (users might not read the log until later).

I think it would be okay to avoid panics if we send an `lsp-show-error` command here, to tell the user.

(Optionally, we could run `lsp-show-error` only if the wrong configuration applies to the language in context.
However, deducing that from invalid TOML input sounds like very high effort for very low gain.)

BTW `lsp-server-configuration` is deprecated in favor of `lsp-config`.

rc/lsp.kak
```diff
@@ -286,6 +286,7 @@ buffile  = \"${kak_buffile}\"
 filetype = \"${kak_opt_filetype}\"
 version  = ${kak_timestamp:-0}
 method   = \"textDocument/didChange\"
+hook     = true
```
krobelus, line 289: OK. Feels like a bug in Kakoune.

src/session.rs
```diff
@@ -120,6 +117,11 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
                         // before. In that case didClose can be safely ignored as well.
                         if request.method != notification::DidCloseTextDocument::METHOD {
                             debug!("Spawning a new controller for {:?}", route);
+                            return_request_error(
+                                editor.to_editor.sender(),
+                                &request,
+                                "Starting language server"
```
krobelus, line 123: I usually run `lsp-enable` via a key binding. Whenever I do that, this shows "Starting language server" in the info box.
This info box is also shown when running a request (like `lsp-hover`) after the language server has exited (could be after a timeout).

Not sure how I feel about these, they seem a bit redundant.

It's not really an error; most requests should proceed as normal.
Only blocking requests can't be fulfilled, due to technical limitations - hence the `cancelling blocking request` message.

src/session.rs
```diff
@@ -105,9 +100,11 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
                 match controllers.entry(route.clone()) {
                     Entry::Occupied(controller_entry) => {
                         if let Err(err) = controller_entry.get().worker.sender().send(request.clone())  {
-                            if let Some(fifo) = request.meta.fifo {
-                                cancel_blocking_request(fifo);
-                            }
+                            return_request_error(
+                                editor.to_editor.sender(),
+                                &request,
+                                "Language server is no longer running"
```
krobelus, line 106: Right, this is definitely an error scenario where it seems good to show the error (I believe this won't happen on a mere timeout).

krobelus, line 106: > Display an error when lsp server is down
> 
> If an LSP server crashes, two requests fail in a row.

Yeah that's weird alright. We could try to restart automatically but maybe it's safer not to.
I'm testing with gopls (simulating crashes with `pkill gopls`) and I usually have only one failing request but sometimes two.

- sometimes the first lsp-hover after crash shows "Language server no longer running" and the second one works
- sometimes the first lsp-hover after crash shows nothing, the second "Language server no longer running" and the third one works

krobelus:
> - It's awkward if someone installs a new kak-lsp that gets started in a kakoune session that's still running outdated kak-side code; after this PR that would make it print errors for all commands, even in a hook. I'm guessing this isn't supported, but it could be useful in general to warn the user if they update kak-lsp swithout restarting kakoune?

Yeah this issue surfaces from time to time. Usually it's because someone updated kak-lsp/rc/lsp.kak via plug.kak but they didn't recompile the binary (or have the wrong one in $PATH).
My skeleton idea is to run something like `cmp "${kak_val_source}" <(kak-lsp --kakoune)` to check if the binary and lsp.kak differ.
We'd probably add option to turn off that check because it can be convenient while developing kak-lsp (no need to recompile).

> - It's easy to write sh code that can mess up request toml formatting in  `lsp.kak`. e.g. try `:e foo."'"` and see the requests fail to parse. Has anyone considered moving some or all of the env parsing logic into `kak-lsp --request` itself? (This might require a more complicated cli, with different cli commands for different operations... But shell escapes are also tricky.)

kak-lsp is [really slow on Cygwin](https://github.com/kak-lsp/kak-lsp/issues/628#issuecomment-1196798205).
Still need to get to the bottom of that but I it just might be enough motivation to avoid shell calls entirely.
We can use `echo %val{..} -to-file kak-lsp-fifo` to reach the controller, which in turn can directly write to Kakoune's socket.
(Or easier: a long-lived `kak -p process` but that'd need changes in Kakoune)

... but until we actually do that we should move in smaller increments

> `:e foo."'"`

Right, I think I have a drive-by fix for that 6efa0e77de88386e553b58b546579ee91967a701 in the cygwin branch.

> Has anyone considered moving some or all of the env parsing logic into `kak-lsp --request` itself? (This might require a more complicated cli, with different cli commands for different operations...)

That sounds like a reasonable idea.
Kakoune exports `$kak_buffile` anyway, so `kak-lsp --request` should probably append that information automatically before passing the TOML request to the server process.
This part of the CLI (`--request`) is not really used by the public so it's fine to change it.

>  But shell escapes are also tricky.

I think it's pretty simple, it's just easy to forget.
In this case we're actually talking about TOML string literal escaping, which has a simple definition as well.

krobelus:
I've merged most of these already as 40197196..8a14601c~ to avoid a trivial merge conflict

src/editor_transport.rs
Spferical: I wasn't sure how to know which client to notify if we can't parse the toml. Is there an easy way for a user to realize that `kak-lsp` panicked if they aren't manually running it in a separate terminal? As far as I can tell, it just silently restarts in the background with no indication of what's happening. Maybe there's a quick and dirty solution in grepping the input for  `client = "..."`.

Or, of course, this crash could probably be mitigated enough by fixing any remaining shell escaping errors. I just wonder if there are any spooky edge cases that might remain, e.g. if it's possible for a command to be half-written if the user ctrl+c's at the wrong time, or their terminal emulator crashes, etc.

src/session.rs
Spferical: Yeah, it's probably too noisy. I think it should be possible to collapse this into the "no longer running" case anyways so that a language server dying only results on one failed request and a regular successful start is always silent.

src/editor_transport.rs
krobelus: > Is there an easy way for a user to realize that `kak-lsp` panicked if they aren't manually running it in a separate terminal?

Yeah, sadly there is no easy way other than looking at the output (or log file).
I want to make something like neovim's :LspInfo that shows debug info about all running servers.
  (Unfortunately the "all" bit doesn't seem straightforward: I might be wrong but I think kak-lsp will spawn a separate controller process for each filetype, so the languages don't share any state; there is no central list of all controllers.)
  Still, we can write an `:lsp` command (removing the old one) that shows the status of the current buffer's server in a scratch buffer.
  It should display the last few lines of the log.

---

Normally I'd rather fix the escaping straight away. Probably not worth the effort, here's why:
We serialize it like

    value="$(printf %s "$value"|sed -e 's/\\=/=/g')"
    printf '%s = %s\n' "$quotedkey" "$value"

so we don't add quotes to value, because it could be any of bool/string/number, etc.
We escape it (only if it's a string), but that might change the behavior for users who already worked around this bug and escaped the value themselves.

We should encourage users to migrate to lsp_config, which uses real TOML syntax (and uses lsp-show-error if it's invalid). How did you find lsp_server_configuration? It should have been removed from docs, the option is hidden. I guess you already used it before?

> Maybe there's a quick and dirty solution in grepping the input for  `client = "..."`.

Fine by me. Alternatively, sending a message to all clients would also work, but that's hardly easier.

> I just wonder if there are any spooky edge cases that might remain, e.g. if it's possible for a command to be half-written if the user ctrl+c's at the wrong time, or their terminal emulator crashes, etc.

Possible but unlikely. When kak reads ctrl+c, it sends SIGINT to the entire process group, which includes the `kak-lsp --request` process (but not the controller process which is daemonized). But `kak-lsp --request`is so short-lived, and the usual TOML payload easily fits in one write(2) syscall, which won't be interrupted even by signals.

src/session.rs
krobelus: Yeah let's try to do that. Our default timeout of 30 minutes means this may happen a few times in a working day, so it's worth improving

src/editor_transport.rs
Spferical: > > Is there an easy way for a user to realize that `kak-lsp` panicked if they aren't manually running it in a separate terminal?
> 
> Yeah, sadly there is no easy way other than looking at the output (or log file). I want to make something like neovim's :LspInfo that shows debug info about all running servers. (Unfortunately the "all" bit doesn't seem straightforward: I might be wrong but I think kak-lsp will spawn a separate controller process for each filetype, so the languages don't share any state; there is no central list of all controllers.) Still, we can write an `:lsp` command (removing the old one) that shows the status of the current buffer's server in a scratch buffer. It should display the last few lines of the log.

I don't completely understand why panicking is better than logging an error in the failed-to-parse case, if it's as hard for regular users to notice either way. But sounds like the debug info command is the right way forward.

Fun fact: I managed to get failed-to-parse errors from editing files with some bad unicode characters (mostly scratch text files I paste things into), whereas e.g. vscode seems to forward broken files like this to its language servers to complain about instead. I guess the buffer contents would need to be base64'd to allow passing raw bytes through toml.

```
Aug 07 07:27:25.367 ERRO Failed to parse editor request: invalid character in string: `\u{1}` at line 5013 column 18, module: kak_lsp::editor_transport:133
```

> We should encourage users to migrate to lsp_config, which uses real TOML syntax (and uses lsp-show-error if it's invalid). How did you find lsp_server_configuration? It should have been removed from docs, the option is hidden. I guess you already used it before?

Looks like it's from when I first configured kak-lsp... About 3 years ago, haha.

> > Maybe there's a quick and dirty solution in grepping the input for  `client = "..."`.
> 
> Fine by me. Alternatively, sending a message to all clients would also work, but that's hardly easier.

I implemented this, with an extra regex for `hook = true`. Feels a bit hacky, but IMO better than crashing.

rc/lsp.kak
Spferical: Created https://github.com/mawww/kakoune/issues/4723 for it (and https://github.com/mawww/kakoune/issues/4724 for a more official detection method)

src/session.rs
```diff
@@ -149,14 +145,59 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
...
 
-/// When server is not running it's better to cancel blocking request.
-/// Because server can take a long time to initialize or can fail to start.
-/// We assume that it's less annoying for user to just repeat command later
-/// than to wait, cancel, and repeat.
-fn cancel_blocking_request(fifo: String) {
-    debug!("Blocking request but LSP server is not running");
-    let command = "lsp-show-error 'language server is not running, cancelling blocking request'";
-    std::fs::write(fifo, command).expect("Failed to write command to fifo");
+/// Tries to send an error to the client about a request that failed to parse.
```
krobelus, line 148: I would probably instead have stuffed the hole in our TOML escaping.
But sure, why not, looks fine.

src/util.rs
```diff
@@ -126,3 +126,17 @@ pub fn short_file_path<'a>(target: &'a str, current_dir: &str) -> &'a str {
         .and_then(|p| p.to_str())
         .unwrap_or(target)
 }
+
+pub fn meta_for_session(session: String, client: Option<String>) -> EditorMeta {
```
krobelus, line 130: making this a free function is good but I think it should stay in `context.rs` which is a better name than the low-entropy `util.rs`

src/session.rs
```diff
@@ -149,14 +145,59 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
...
+fn handle_broken_editor_request(
+    to_editor: &Sender<EditorResponse>,
+    request: String,
+    session: &str,
+) {
+    // Try to parse enough of the broken toml to send the error to the editor.
+    let client_re = regex::Regex::new(r#"(?m)^client\W*=\W*"([^"\n]+)""#)
```
krobelus, line 155: even though it won't matter in practice, I think it's good style to compile this effectively-const regex only once using `lazy_static`

src/editor_transport.rs
```diff
@@ -127,8 +125,6 @@ pub fn start_unix(path: &path::Path, sender: Sender<EditorRequest>) {
                             continue;
                         }
                         debug!("From editor: {}", request);
-                        let request: EditorRequest =
-                            toml::from_str(&request).expect("Failed to parse editor request");
                         if sender.send(request).is_err() {
                             return;
                         };
```
krobelus, line 130: > This generally indicates an internal error with kak-lsp with escaping in
> the toml-serializing shell code. A fatal error is not ideal here because
> it brings down all language servers and causes requests to fail during
> the kak-lsp and language server startup process.

Huh, I thought there's only one kak-lsp "controller" process for each language
server and that's it. Apparently there's also a central kak-lsp process that
listens on the session socket. Not sure. Anyway, this seems fine.

src/session.rs
```diff
@@ -149,14 +145,59 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
...
+        std::fs::write(fifo, &command).expect("Failed to write command to fifo");
+    }
+
+    if !request.meta.hook {
+        let response = EditorResponse {
+            meta: request.meta.clone(),
+            command: command.into(),
+        };
+        if let Err(err) = to_editor.send(response) {
+            error!("Failed to send error message to editor: {err}");
```
krobelus, line 198: this fixup should go in the previous commit

src/session.rs
```diff
@@ -76,30 +76,12 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
 
                 let language_id = filetypes.get(&request.meta.filetype);
                 if language_id.is_none() {
-                    debug!(
+                    let msg = format!(
```
krobelus, line 91: can you add to the commit message a scenario that this commit improves?
It's late here and I think I forgot how to reproduce it.

src/session.rs
Spferical: I think I understand what's happening here now -- the language server transport layer takes a second to shut down. Upon a regular `pkill rust-analyzer` I get:

```
Oct 27 03:25:07.795 DEBG Language server closed pipe, stopping reading, module: kak_lsp::language_server_transport:134
Oct 27 03:25:07.796 INFO Waiting for Messages to language server to finish..., module: kak_lsp::thread_worker:19
Oct 27 03:25:07.796 DEBG Received signal to stop language server, closing pipe, module: kak_lsp::language_server_transport:193
Oct 27 03:25:07.796 DEBG Waiting for language server process end, module: kak_lsp::language_server_transport:92
```

If requests are sent during the one-second window it waits for process shutdown (?), they are ignored. Then, a second later it fully shuts down:
```
Oct 27 03:25:08.796 INFO ... Messages to language server terminated with ok, module: kak_lsp::thread_worker:21
Oct 27 03:25:08.796 INFO Waiting for Messages from language server to finish..., module: kak_lsp::thread_worker:19
Oct 27 03:25:08.796 INFO ... Messages from language server terminated with ok, module: kak_lsp::thread_worker:21
Oct 27 03:25:08.797 INFO Waiting for Language server errors to finish..., module: kak_lsp::thread_worker:19
Oct 27 03:25:08.797 INFO ... Language server errors terminated with ok, module: kak_lsp::thread_worker:21
```

src/session.rs
Spferical: Added this and more context. For me it's mostly either rust-analyzer crashing or (maybe very rarely) the out-of-memory killer. I reproduce it by `pkill`ing the language server, though note that there's a small race condition where requests sent to the language server within a second of `pkill`ing it are ignored.

Spferical:
Sorry for the delay -- I think I addressed all of your comments.

Followup work:
- Filenames like `'"'` should be escaped in request toml.
- The current sh toml escaping doesn't escape control characters (U+0000 to U+001F). Maybe for this one it's easiest to base64 the file contents and forward the raw file bytes through to the language server.
- The language server transport code takes a second to shut down for me, during which requests are silently ignored. Would be slightly nicer to block the channel while waiting for the server process to stop to be a little more responsive.

PR #636, opened 2022-08-05
Disable completion snippet placeholders when leaving insert mode

krobelus:
When snippets are enabled, we can type

	openat<c-n>somefd<tab>somefile<tab>O_CREAT,<space>0644

to get

	openat(somefd, somefile, O_CREAT, 0644)

Very fancy.
Sadly, I often can't fill in all arguments without looking
around or fixing some other code.  If I stop somewhere in
the middle, highlighting remains, and can only be cleared
with `lsp-snippets-select-next-placeholders`.  For example
`openat<c-n><esc>` gives

	openat(%(), const char *file, int oflag, ...)

where the tail of arguments is still highlighted.  This is supremely
annoying because I usually don't want to simply resume where I left
of.

Also, the last snippet-placeholder is usually after the closing
parenthesis. Jumping there makes sense when staying in insert mode
throughout but not when when exiting insertion mode halfway through
the arguments.

Let's just hide the placeholder highlighting as soon we exit insert
mode. This should make snippets interfere less with normal editing.

It seems fine to leave the raw placeholder text because it serves
as documentation and other editors like VSCode do the same.

If anyone really cares about this, we can use a separate hook group,
that can be disabled specifically.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -526,7 +526,6 @@ For example, you can use `<tab>` to select a placeholder if there is one, or els
 
 [source,kak]
 ----
-map global normal <tab> ': try lsp-snippets-select-next-placeholders catch %{ execute-keys -with-hooks <lt>tab> }<ret>' -docstring 'Select next snippet placeholder'
 map global insert <tab> '<a-;>: try lsp-snippets-select-next-placeholders catch %{ execute-keys -with-hooks <lt>tab> }<ret>' -docstring 'Select next snippet placeholder'
 ----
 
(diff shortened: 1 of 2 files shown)


krobelus:
cc @topisani - this has been bugging me for a while, IDK why I put up with it so long :(

topisani:
I've had issues with snippets for the same reason - this might be a good way to fix it - I approve!

topisani:
I've had issues with snippets for the same reason - this might be a good way to fix it - I approve!

krobelus:
This means we can enable snippets by default now :)
Of course they are only really useful with a mapping to `lsp-snippets-select-next-placeholders`.
Since we can't decide the mapping for the user, I'd add just a prominent section with suggested mappings in https://github.com/kak-lsp/kak-lsp/pull/637

PR #637, opened 2022-08-07
Add section with suggested mappings instead of mapping keys ourselves

krobelus:
By convention, Kakoune plugins do not add their own key mappings.
This makes them more predictable and avoids clashes with the user's
mappings.  This is especially important as we add more mappings,
for example the suggested binding for `<tab>`.

Let's add a "suggested mappings" section, so new users are less likely
to skip mapping `<tab>` (which is quite important for snippets).
Keep keys like "gd" to avoid breakage (also no one has complained
about them). Not sure if we should keep the object mode keys.

### DIFF
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -1,4 +1,4 @@
-snippet_support = false
+snippet_support = true
 verbosity = 2
 
 [server]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Here are some notable changes. See the git log for a full list of changes.
 
 Breaking changes:
 - Completion snippet support is now enabled again in the default config.
+- The default object mode mappings have been removed. Users are expected to add their preferred mappings. The README now has a section with recommended mappings.
 
 ## 12.2.1 - 2022-05-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Unreleased
 
+Here are some notable changes. See the git log for a full list of changes.
+
+Breaking changes:
+- Completion snippet support is now enabled again in the default config.
+
 ## 12.2.1 - 2022-05-08
 
 Fixes:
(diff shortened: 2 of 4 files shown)


kak-lsp.toml
```diff
@@ -1,4 +1,4 @@
-snippet_support = false
+snippet_support = true
```
sclu1034, line 1: This should probably be mentioned in `CHANGELOG.md`.

krobelus: Yep, did that in the [changelog branch](https://github.com/kak-lsp/kak-lsp/commit/changelog).
I use that so I only have one commit per release cycle that touches the changelog. Perhaps we should drop the changelog file from the master branch entirely and use annotated git tags.

sclu1034: I guess I never noticed the disconnect. What's the benefit from having only that one commit?

Downsides are, as far as I can tell, that it makes the "Unreleased" section practically useless, as it will just appear empty at all times (unless someone stumbles upon that `changelog` branch).
And it also makes it very hard to correlate a line in the changelog with the actual change. If it was part of the code change, I could just `git blame CHANGELOG.md` and immediately get the commit corresponding to each line in the changelog. With your way, the PR/Issue IDs can help sometimes, but for the rest I'd have to grep commit messages or code for keywords.

Release descriptions would have the same disconnect, and have the additional issue that they don't exist until the release has been tagged, and so they have to be written from memory/walking back through the history.

krobelus: > What's the benefit from having only that one commit?

When I write prose I usually need many of attempts to get a good result.
I don't want to have commits in master that update the changelog with style changes etc.
Committing the changelog to master only at the end of a release cycle enables that.
The same workflow could be used for drafting the annotated tag.

> And it also makes it very hard to correlate a line in the changelog with the actual change.

I'm happy to add issue/commit references to the every line if that's useful.

sclu1034: Maybe a combination, then? A PR adds a quick-and-dirty summary with the change and you do a consolidated style change once before release.
That way, the "Unreleased" section would be easily accessible in `master` and the `git blame` after a release would still be usable, just be a single commit deeper.

> I'm happy to add issue/commit references to the every line if that's useful.

Doing that manually sounds like quite a bit of extra work.

krobelus: > Maybe a combination, then?

makes sense, thanks!

> Doing that manually sounds like quite a bit of extra work.

yeah, it also doesn't feel super useful since the commits are easy to find anyway

sclu1034 (approved):

PR #639, opened 2022-08-24
lsp-menu: customize prompt string for each caller

krobelus:
This will break users who customize lsp-menu (cc @topisani) but it hopefully improves
the experience for new users (not sure).


topisani:
Ill adjust my `lsp-menu`, no problem, but i agree this is a good idea

krobelus:
with this change, `lsp-menu` is no longer a drop-in replacement for `menu`.
If we go through with that, we might even generate the `prompt` command directly on the Rust side.

I'm surprised that people want to override this - Kakoune's fuzzy ranking algorithm should be on par with `fzy` with https://github.com/mawww/kakoune/pull/4653 - but of course overriding should be possible.
I'm not sure what you are using - is it a fzf split window?

Actually we should probably change Kakoune's `menu` to accept a `-prompt` switch `menu -prompt 'Jump to symbol' <name1> <cmd1> <name2> <cmd2>`.
I'm not very interested in today's `menu` (it has no fuzzy filtering, so I prefer `prompt`) but its interface seems alright so we might as well improve it.

In exchange for the nice interface of `menu`, the implementation of `lsp-menu` is fairly tricky.
Long-term, I think we should give `menu` the same fuzzy filtering, then `lsp-menu` can just delegate to `menu`.

topisani:
This is what it gives me - i dont know why i like it better, and i may try the default lsp-menu again. I usually dont even use the fuzzy finding, and just select the one i want with my arrowkeys. 

I do prefer them stacked vertically though, it makes it easier for me to read them.

image

PR #673, opened 2023-02-23
Replace Perl with shell script in LSP snippet insertion

poweredbypie:
Snippet support is currently broken on NixOS because of perl not being on path by default, and I don't really see why Perl is even needed for this, so I figured I would try to rewrite this part in shell.


poweredbypie:
I'm not 100% confident that this works; from my tests snippets work fine but since posix sh doesn't have the concept of a map, if a snippet has multiple of the same identifier, it won't substitute properly. I'm hoping there's some way around this, but I'm not particularly experienced in writing shell, so any suggestions are appreciated.

Example:
`'${0:thing} $0'` comes out to `set-register dquote 'thing' 'thing'` with the existing Perl script, but `set-register dquote 'thing' ' '` with my shell script.

sclu1034:
Is it really that much of a problem to install something as ubiquitous as Perl on NixOS?

In addition to not being feature-complete, the shell implementation is much slower:

```
+ export kak_quoted_selections='${0:expression}' '${1:const void *p}' '$0'
+ hyperfine --warmup 3 -N perl test.pl dash test.sh bash test.sh
Benchmark 1: perl test.pl
  Time (mean ± σ):      15.6 ms ±   2.5 ms    [User: 10.0 ms, System: 5.0 ms]
  Range (min … max):    10.1 ms …  22.0 ms    246 runs

Benchmark 2: dash test.sh
  Time (mean ± σ):     146.1 ms ±   5.1 ms    [User: 109.8 ms, System: 44.1 ms]
  Range (min … max):   134.4 ms … 153.9 ms    19 runs

Benchmark 3: bash test.sh
  Time (mean ± σ):     171.5 ms ±   5.7 ms    [User: 121.4 ms, System: 66.4 ms]
  Range (min … max):   159.7 ms … 178.1 ms    17 runs

Summary
  'perl test.pl' ran
    9.38 ± 1.53 times faster than 'dash test.sh'
   11.02 ± 1.80 times faster than 'bash test.sh'
```

150ms will likely produce a noticeable latency between running the command and seeing the result.

sidkshatriya:
We could require `perl` in the kak-lsp nix derivation. 

At a glance, I think we might need to add this to the `propagatedBuildInputs` of the kak-lsp nix derivation, though I haven't tried this out.

krobelus:
Kakoune's `git.kak` and `diff.kak` already depend on Perl. The first is warranted because Git already depends on Perl. The second is a historical accident. If you're not using `<ret>` in git/diff buffers, you're missing out ;)
It's probably a good idea to remove the dependency. A close relative of Perl would be POSIX `awk` which would enable a more elegant implementation than shell.

poweredbypie:
> Is it really that much of a problem to install something as ubiquitous as Perl on NixOS?

It's not an issue with installation, but to maintain a truly clean derivation (I, with my current knowledge of Nix, believe) I have to maintain a [patch upstream to fix the perl path](https://github.com/NixOS/nixpkgs/pull/217658/commits/814d56148baa80e6ca92638e2d6874edefd4651d), which I don't really want to do.

> In addition to not being feature-complete, the shell implementation is much slower:

Hmm... I guess I didn't consider this. I guess I'm a bit confused by the architecture of the `lsp.kak` file: why aren't we doing most of the string processing on the Rust side? I'd much rather implement the formal BNF from the LSP definition than use a regex to parse it.

> I think we might need to add this to the `propagatedBuildInputs` of the kak-lsp nix derivation, though I haven't tried this out.

I'm fairly sure that `propagatedBuildInputs` is for development purposes only, but I guess I couldn't really tell from the sparse documentation... I have [a PR upstream](https://github.com/NixOS/nixpkgs/pull/217658) that I'd love for someone to code review, if you'd be willing to.

poweredbypie:
> Kakoune's `git.kak` and `diff.kak` already depend on Perl.

Ah, I didn't even realize this was a thing! Guess I have a lot to discover in kakoune still.

> A close relative of Perl would be POSIX `awk` which would enable a more elegant implementation than shell.

While I'm fine with reimplementing this in awk, I'm still a bit curious to why string parsing isn't mostly handled in Rust? `lsp.kak` looks like it does a lot of work here where I feel like it shouldn't.

PR #675, opened 2023-03-01
Fix for hover in docsclient not working

sidkshatriya:
I created a fresh window via `:new`, the client name was `client1`. Then I did

`:set global docslient 'client1'`

I expected to see any hover triggered from `client0` appear in `client1`. However that did not happen. 

This commit should fix the issue.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -446,9 +446,9 @@ If a client name argument is given use that client. Create it with :new if it do
 define-command -hidden lsp-hover-request -params 0..1 -docstring "Request hover info for the main cursor position" %{
     evaluate-commands %sh{
         hover_buffer_args=""
-        if [ $# -eq 1 ]; then
-            client=${1:-"${kak_opt_docsclient:-"$kak_client"}"}
-            hover_buffer_args="hoverClient = \"${client}\""
+        if [ $# -eq 1 ] || [ -n "${kak_opt_docsclient}" ]; then
+            hover_client=${1:-"${kak_opt_docsclient:-"$kak_client"}"}
+            hover_buffer_args="hoverClient = \"${hover_client}\""
         fi
 
         (printf %s "


rc/lsp.kak
```diff
@@ -446,9 +446,9 @@ If a client name argument is given use that client. Create it with :new if it do
 define-command -hidden lsp-hover-request -params 0..1 -docstring "Request hover info for the main cursor position" %{
     evaluate-commands %sh{
         hover_buffer_args=""
-        if [ $# -eq 1 ]; then
-            client=${1:-"${kak_opt_docsclient:-"$kak_client"}"}
-            hover_buffer_args="hoverClient = \"${client}\""
+        if [ $# -eq 1 ] || [ -n "${kak_opt_docsclient}" ]; then
+            hover_client=${1:-"${kak_opt_docsclient}"}
+            hover_buffer_args="hoverClient = \"${hover_client}\""
```
krobelus, line 451: we should still fall back to `$kak_client`

sidkshatriya: If there is something wrong with `$1` or `$kak_opt_docsclient`,  falling back to `$kak_client` that will have the strange effect of your current window being taken over by a hover. 

It will be even more confusing because the window will be totally overtaken and it will not be an information popup.

sidkshatriya: (hoverClient, essentially replaces the whole content area, it is not an information popup as you know)

sidkshatriya: (Discussing the fallback to `$kak_client` contd...)

Actually to be more accurate, it will not replace the new area but create a new buffer which will be made active. This could be confusing for the user. 

Also when auto hover is enabled, going back to the code via `<c-o>` can be impractical because auto-hover will send the user back to hover in a loop and that can be frustrating for the user...

sidkshatriya: I have pushed a new commit.

I now agree that there _does need_ to be a fallback to `$kak_client` to deal with `lsp-hover-buffer` called without any arguments, specifically when we use `lsp H`

```kak
map global lsp H '<esc>: lsp-hover-buffer<ret>'            -docstring 'show info for current position in a scratch buffer'
```

In that case `$1` is `''`

sidkshatriya:
The second force push was just a stylistic change. No impact on functionality.

rc/lsp.kak
krobelus: yeah I only ever used lsp-hover-buffer via the H key binding.
The single-window use case is important despite the limitations

PR #679, opened 2023-04-10
Implement `window/showMessageRequest`

### LINKED ISSUE #183: Implement window/showMessageRequest
This would be rally helpful for scala support as scala metals uses this to trigger automatic build imports. I think it could work as follows:

kak-lsp buffers requests. Users add a hook for idling events. On idle poll for server requests. If there is are pending requests dequeue one display via infobox with keymaps to select option to send back.

If we had an implementation path I'd be willing to work on this.

natsukagami:
Implement [`window/showMessageRequest`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessageRequest). Closes #183.

- Adds a new modeline option, `lsp_modeline_message_requests`, that displays the
number of pending message requests on the modeline, as a bell + number of requests.
- A new user-facing command, `lsp-show-message-request-next`, pops and shows the
next pending request.

- The requests are stored in a queue (a `VecDeque`) in the context.
- We display the message and the prompt by an `info` + `lsp-menu` call.
- Responses to the request are made by the menu calling back with `lsp-show-message-request-respond`.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -128,6 +128,11 @@ info=$lsp_info \
             lines++
         }
 
+        if (ENVIRON["kak_opt_lsp_modeline_message_requests"]) {
+            r = r "There are unread messages (use lsp-show-message-request-next to read)\n"
+            lines++
+        }
+
         info_lines = split(info, info_line, /\n/)
         for (i = 1; i <= info_lines && (max_lines <= 0 || i+lines+2 <= max_lines); i++)
             print info_line[i]
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -207,6 +207,7 @@ define-command -hidden lsp-menu -params 1.. -docstring "Like menu but with promp
                     esac
                 ¶
             §" "$cases"
+            printf ' %s' "$on_abort"
             printf ' -menu -shell-script-candidates %%§
                 printf %%s %s
                 §\n' "$(shellquote "$completion")"
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -177,6 +177,11 @@ define-command -hidden lsp-menu -params 1.. -docstring "Like menu but with promp
             shellquote() {
                 printf "'%s'" "$(printf %s "$1" | sed "s/'/'\\\\''/g; s/§/§§/g; $2")"
             }
+            on_abort=
+            if [ $1 = "-on-abort" ]; then
+                ob_abort="-on-abort $(shellquote "$2" s/¶/¶¶/g)"
+                shift 2
+            fi
             cases=
             completion=
             nl=$(printf '\n.'); nl=${nl%.}
@@ -261,7 +266,8 @@ declare-option -hidden str lsp_project_root
 
 declare-option -hidden str lsp_modeline_code_actions
 declare-option -hidden str lsp_modeline_progress ""
-declare-option -hidden str lsp_modeline '%opt{lsp_modeline_code_actions}%opt{lsp_modeline_progress}'
+declare-option -hidden str lsp_modeline_message_requests ""
+declare-option -hidden str lsp_modeline '%opt{lsp_modeline_code_actions}%opt{lsp_modeline_progress}%opt{lsp_modeline_message_requests}'
 set-option global modelinefmt "%opt{lsp_modeline} %opt{modelinefmt}"
 
 ### Requests ###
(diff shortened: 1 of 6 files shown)


rc/lsp.kak
```diff
@@ -1734,6 +1740,56 @@ define-command -hidden lsp-show-message-log -params 1 -docstring %{
...
+define-command -hidden lsp-show-message-request -params 4.. -docstring %{
+    lsp-show-message-request <prompt> <on-abort> <opt> <command> [<opt> <command>]...
+    Render a prompt message with options.
+} %{
+    evaluate-commands -try-client %opt{toolsclient} %{
+        info -title "From LSP" %arg{1}
+        evaluate-commands %sh{
+            on_abort="$2"
+            shift 2
+            echo lsp-menu -on-abort $on_abort "$@"
```
natsukagami, line 1752: I'm passing the args naked here, so I had to quote twice on kak-lsp side, I wonder if there's any better way to do it?

natsukagami:
Perhaps instead of just updating the modeline we can have something similar to `lsp-handle-progress`, an override-able function to handle message requests arriving.

rc/lsp.kak
```diff
@@ -177,6 +177,11 @@ define-command -hidden lsp-menu -params 1.. -docstring "Like menu but with promp
             shellquote() {
                 printf "'%s'" "$(printf %s "$1" | sed "s/'/'\\\\''/g; s/§/§§/g; $2")"
             }
+            on_abort=
+            if [ $1 = "-on-abort" ]; then
+                ob_abort="-on-abort $(shellquote "$2" s/¶/¶¶/g)"
```
krobelus, line 182: typo (ob_abort -> on_abort)

To make this do something we'd need `prompt -on-abort` support in Kakoune.
Is it actually necessary here though? What's the worst that can happen if we never send our reply?

rc/lsp.kak
```diff
@@ -1734,6 +1740,56 @@
...
+        info -title "From LSP" %arg{1}
+        evaluate-commands %sh{
+            on_abort="$2"
+            shift 2
+            echo lsp-menu -on-abort $on_abort "$@"
+        }
+    }
+}
+
+define-command lsp-show-message-request-next -params 0 -docstring %{
```
krobelus, line 1757: lose the `-params 0` for consistency please

src/show_message.rs
```diff
@@ -0,0 +1,137 @@
...
+use lsp_types::{MessageActionItem, MessageType, ShowMessageParams, ShowMessageRequestParams};
+use serde::Deserialize;
+
+use crate::{context::Context, types::EditorMeta, util::editor_quote};
+
+/// Queues the message request from the LSP server.
+pub fn show_message_request(meta: EditorMeta, request: MethodCall, ctx: &mut Context) {
+    let params: ShowMessageRequestParams = request
+        .params
+        .clone()
```
krobelus, line 14: don't clone the request here, rather use `let request_id = request.id` or similar

rc/lsp.kak
```diff
@@ -1734,6 +1740,56 @@
     echo -debug "kak-lsp: log:" %arg{1}
 }
 
+define-command -hidden lsp-show-message-request -params 4.. -docstring %{
+    lsp-show-message-request <prompt> <on-abort> <opt> <command> [<opt> <command>]...
+    Render a prompt message with options.
+} %{
+    evaluate-commands -try-client %opt{toolsclient} %{
+        info -title "From LSP" %arg{1}
```
krobelus, line 1748: "From language server"

src/show_message.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub item: Option<toml::Value>,
+}
+
+/// Handles an user's response to a message request (or the user's request to display the next message request).
+pub fn show_message_request_respond(meta: EditorMeta, params: toml::Value, ctx: &mut Context) {
+    let resp =
+        MessageRequestResponse::deserialize(params).expect("Cannot parse message request response");
+    let req_id = match resp.message_request_id {
+        None => return show_message_request_next(meta, ctx),
+        Some(id) => id,
```
krobelus, line 33: can write this as

	let Some(req_id) = resp.message_request_id else { return show_message_request_next(meta, ctx); }

but actually I don't think it's great to overload `window/showMessageRequest` with two disjoint meanings

Make it two separate methods instead?

- window/showMessageRequest/showNext
- window/showMessageRequest/respond

src/show_message.rs
```diff
@@ -0,0 +1,137 @@
...
+        }
+    };
+
+    let options = match &params.actions {
+        Some(opts) if !opts.is_empty() => &opts[..],
+        _ => {
+            // a ShowMessageRequest with no actions is just a ShowMessage notification.
+            show_message(
+                meta,
+                ShowMessageParams {
```
krobelus, line 57: It does not feel right to construct a ShowMessageParams object here.
Maybe pass two parameters instead (for the type and message)

src/show_message.rs
```diff
@@ -0,0 +1,137 @@
...
+    let request_id = editor_quote(
+        toml::to_string(&id)
+            .expect("cannot convert request id to toml")
+            .as_ref(),
+    );
+    let option_menu_opts = options
+        .iter()
+        .flat_map(|item| {
+            let cmd = format!(
+                "%{{lsp-show-message-request-respond {} {}}}",
```
krobelus, line 76: don't use nested quoting for user input because we don't know if `item` has balanced quotes.
It should be like

	editor_quote(&format!("lsp-show-message-request-respond {} {}", ...))

rc/lsp.kak
```diff
@@ -1734,6 +1740,56 @@
...
+        evaluate-commands %sh{
+            on_abort="$2"
+            shift 2
+            echo lsp-menu -on-abort $on_abort "$@"
```
krobelus, line 1752: needs to be quoted as `"$on_abort"` here for perfect forwarding

rc/lsp.kak
krobelus: the double Kakoune-quoting makes sense, because we evaluate the `lsp-menu` command here and later the `lsp-menu` command will evaluate its args once again.
It's not super pretty but it saves processes so it's okay.
We do still want to quote `$on_abort` to protect against shell evaluation, which is unrelated to Kakoune quoting/evaluation

krobelus:
> Perhaps instead of just updating the modeline we can have something similar to lsp-handle-progress, an override-able function to handle message requests arriving.

I'd need to try it (not sure which servers send it beyond scala metals).
We can merge this and improve it later.
I think it's not great that there's different commands for code action, hover, code lenses, show-message.

One way to improve the discoverability of the current approach would be to add to `lsp-hover` a hint at messages.
Something like
`5 unread messages (use lsp-show-messages-next to read)`
like we already do for diagnostics, code lenses and code actions.

We could also have an umbrella `lsp-interact` command that offers a menu to do any of the above.

rc/lsp.kak
natsukagami: Aside from `metals` waiting forever for a reply, I don't think anything else would happen. 
Btw `prompt -on-abort` has been in Kakoune since https://github.com/mawww/kakoune/commit/2edb7d62ef50c65a04de499768435c2fc41a832c (6 years ago)

krobelus: good, then let's add it to the `prompt` command below, something like

```diff
diff --git a/rc/lsp.kak b/rc/lsp.kak
index 10db851..917a7e7 100644
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -208,6 +208,7 @@ define-command -hidden lsp-menu -params 1.. -docstring "Like menu but with promp
                     esac
                 ¶
             §" "$cases"
+            printf ' %s' "$on_abort"
             printf ' -menu %s -shell-script-candidates %%§
                 printf %%s %s
                 §\n' "$($kak_opt_lsp_have_kakoune_feature_completion_priority && printf %s -priority)" "$(shellquote "$completion")"
```

which means we don't need shellquote, only kakoune quoting so something like `s/¶/¶¶/g`

natsukagami:
I added a simple message to `lsp-hover`.

rc/lsp.kak
natsukagami: I added it, but 
> which means we don't need shellquote, only kakoune quoting so something like s/¶/¶¶/g

I'm not sure what you mean here. Are you referring to this part?
```bash
                on_abort="-on-abort $(shellquote "$2" s/¶/¶¶/g)"
```

krobelus: yeah we don't need shellquote for on_abort because on_abort will never be evaluated by a shell (unlike the case statements)

rc/lsp.kak
```diff
@@ -128,6 +128,11 @@ info=$lsp_info \
             lines++
         }
 
+        if (ENVIRON["kak_opt_lsp_modeline_message_requests"]) {
```
krobelus, line 131: `kak_opt_lsp_modeline_message_requests` is not exported at this point, need to add it to this line:

```
    content=$(eval "${kak_opt_lsp_show_hover_format}") # kak_opt_lsp_hover_max_lines kak_opt_lsp_modeline_code_actions
```

krobelus:
merged as 72a8816..39dcd5a, thanks

PR #686, opened 2023-05-29
Handle multiple servers per language

### LINKED ISSUE #17: Support multiple language servers per filetype
Use case: for javascript files I want diagnostics from both eslint and flow language servers.

gbrlsnchs:
This PR is my first formal attempt of pushing necessary changes for handling multiple language servers per filetype.

The code is still broken and doesn't compile. I'll keep pushing commits here when I have time to work on this, preferably organized in meaningful commits, in order to track progress and promote discussion over how it is looking.

Also, I'll keep updating this description as new questions arise. Hopefully this PR will serve as a kick-start for solving this long-standing issue (only if desired).

Fixes #17.

To do:
- [x] Make it work properly
- [x] Tidy commits
- [x] Remove labels for single server
- [ ] Sort responses by priority
- [x] Update config file
- [x] Update docs
- [x] Make sure goto buffers are properly set when 2+ servers have different root paths

---

### Configuring multiple language servers

```toml
[language_ids]
javascript = "javascriptreact"
typescript = "typescriptreact"

[language_server.tsserver]
filetypes = ["javascript", "typescript"]
roots = ["package.json", "tsconfig.json", "jsconfig.json", ".git", ".hg"]
command = "typescript-language-server"
args = ["--stdio"]

[language_server.tailwindcss]
filetypes = ["javascript", "typescript"]
roots = ["tailwind.config.js", "tailwind.config.ts"]
command = "tailwindcss-language-server"
args = ["--stdio"]

[language_server.tailwindcss.settings.tailwindcss]
editor = {}
```

> **Note**
> You can now either use `language` or `language_server` as the field name. However, it should be consistent across your config.

### Capabilities

image

### Formatting (and range formatting)

When there are more than one capable formatting server, a prompt appears so users can choose which formatting server to use (same prompt used in code actions/lenses).

image

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -304,22 +304,52 @@ Please let us know if you have any ideas about how to make the default config mo
 ==== Server-specific configuration
 
 Many servers accept configuration options that are not part of the LSP spec.  The TOML table
-`[language.<filetype>.settings]` holds those configuration options.  It has the same structure
+`[language_server.<server_name>.settings]` holds those configuration options.  It has the same structure
 as the corresponding fragments from VSCode's `settings.json`. For example:
 
 [source,toml]
 ----
-[language.go]
+[language_server.gopls]
 ...
-settings_section = "gopls"
-[language.go.settings.gopls]
+settings_section = "gopls" # Optional, defaults to server name.
+[language_server.gopls.settings.gopls]
 "formatting.gofumpt" = true
 ----
 
-During server initialization, kak-lsp sends the section specified by `settings_section`; in this
-case `{"formatting.gofumpt":true}`.  Whenever you change the Kakoune option `lsp_config`, the
-same section is sent via `workspace/didChangeConfiguration`.  Additionally, kak-lsp will send
-arbitrary sections that are requested by the server in `workspace/configuration`.
+During server initialization, kak-lsp sends either the section specified by `settings_section`.
+or the section with same name as the server; in this case `{"formatting.gofumpt":true}`.
+Whenever you change the Kakoune option `lsp_config`, the same section is sent via
+`workspace/didChangeConfiguration`.  Additionally, kak-lsp will send arbitrary sections that
+are requested by the server in `workspace/configuration`.
+
+==== Multiple language servers
+
+It is possible to map more than one language server to a filetype. For example, if you want to
+set up TSServer and TailwindCSS to use in React projects:
+
+[source,toml]
+----
+[language_ids]
+javascript = "javascriptreact"
+typescript = "typescriptreact"
+
+[language_server.tsserver]
+filetypes = ["javascript", "typescript"]
+roots = ["package.json", "tsconfig.json", "jsconfig.json", ".git", ".hg"]
+command = "typescript-language-server"
+args = ["--stdio"]
+
+[language_server.tailwindcss]
+filetypes = ["javascript", "typescript"]
+roots = ["tailwind.config.ts", "tailwind.config.js"]
+command = "tailwindcss-language-server"
+args = ["--stdio"]
+[language_server.tailwindcss.settings.tailwindcss]
+editor = {}
+----
+
+Note that, just like in the config above, it is possible to override the language ID sent to
+the language server by mapping filetypes using `language_ids`.
 
 === Configuring Kakoune
 
@@ -334,14 +364,14 @@ kak-lsp's Kakoune integration declares the following options:
 * `lsp_auto_highlight_references` (bool): If this option is `true` then `lsp-highlight-references` is executed every time the user pauses in normal mode.
 * `lsp_auto_show_code_actions` (bool): If this option is `true` then `lsp-code-actions` is executed every time the user pauses in normal mode.
 * `lsp_config` (str): This is a TOML string of the same format as `kak-lsp.toml`, except it currently only supports one kind of configuration value:
-** `[language.<filetype>.settings]`: this works just like the static configuration of the same name in `kak-lsp.toml`, see the section about server-specific configuration. This will override the static configuration of the given language.
+** `[language_server.<server_name>.settings]`: this works just like the static configuration of the same name in `kak-lsp.toml`, see the section about server-specific configuration. This will override the static configuration of the given language server.
 
 For example, you can toggle an option dynamically with a command like this:
 
 [source,kak]
 ----
 set-option global lsp_config %{
-    [language.go.settings.gopls]
+    [language_server.gopls.settings.gopls]
     "formatting.gofumpt" = true
 }
 ----
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -6,129 +6,90 @@ verbosity = 2
 # set to 0 to disable
 timeout = 1800 # seconds = 30 minutes
 
-[language.bash]
+# This section overrides language IDs.
+# By default, kak-lsp use filetypes for the IDs.
+# See https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentItem
+[language_ids]
+c = "c_cpp"
+cpp = "c_cpp"
+javascript = "javascriptreact"
+typescript = "typescriptreact"
+protobuf = "proto"
+sh = "shellscript"
+
+[language_server.bash-language-server]
 filetypes = ["sh"]
 roots = [".git", ".hg"]
 command = "bash-language-server"
 args = ["start"]
 
-[language.c_cpp]
+[language_server.clangd]
 filetypes = ["c", "cpp"]
 roots = ["compile_commands.json", ".clangd", ".git", ".hg"]
 command = "clangd"
 
-[language.clojure]
+[language_server.clojure-lsp]
 filetypes = ["clojure"]
 roots = ["project.clj", ".git", ".hg"]
 command = "clojure-lsp"
 settings_section = "_"
-[language.clojure.settings._]
+[language_server.clojure-lsp.settings._]
 # See https://clojure-lsp.io/settings/#all-settings
 # source-paths-ignore-regex = ["resources.*", "target.*"]
 
-[language.cmake]
+[language_server.cmake-language-server]
 filetypes = ["cmake"]
 roots = ["CMakeLists.txt", ".git", ".hg"]
 command = "cmake-language-server"
 
-[language.crystal]
+[language_server.crystalline]
 filetypes = ["crystal"]
 roots = ["shard.yml"]
 command = "crystalline"
 
-[language.css]
+[language_server.css-language-server]
 filetypes = ["css"]
 roots = ["package.json", ".git", ".hg"]
 command = "vscode-css-languageserver"
 args = ["--stdio"]
 
-[language.less]
+# [language_server.deno-lsp]
+# filetypes = ["javascript", "typescript"]
+# roots = ["package.json", "tsconfig.json", ".git", ".hg"]
+# command = "deno"
+# args = ["lsp"]
+# settings_section = "deno"
+# [language_server.deno-lsp.settings.deno]
+# enable = true
+# lint = true
+
+[language_server.less-language-server]
 filetypes = ["less"]
 roots = ["package.json", ".git", ".hg"]
 command = "vscode-css-languageserver"
 args = ["--stdio"]
 
-[language.scss]
+[language_server.scss-language-server]
 filetypes = ["scss"]
 roots = ["package.json", ".git", ".hg"]
 command = "vscode-css-languageserver"
 args = ["--stdio"]
 
-[language.d]
+[language_server.dls]
 filetypes = ["d", "di"]
 roots = [".git", "dub.sdl", "dub.json"]
 command = "dls"
 
-[language.dart]
+[language_server.dart-lsp]
 # start shell to find path to dart analysis server source
 filetypes = ["dart"]
 roots = ["pubspec.yaml", ".git", ".hg"]
 command = "sh"
 args = ["-c", "dart $(dirname $(command -v dart))/snapshots/analysis_server.dart.snapshot --lsp"]
 
-[language.elixir]
-filetypes = ["elixir"]
-roots = ["mix.exs"]
-command = "elixir-ls"
-settings_section = "elixirLS"
-[language.elixir.settings.elixirLS]
-# See https://github.com/elixir-lsp/elixir-ls/blob/master/apps/language_server/lib/language_server/server.ex
-# dialyzerEnable = true
-
-[language.elm]
-filetypes = ["elm"]
-roots = ["elm.json"]
-command = "elm-language-server"
-args = ["--stdio"]
-settings_section = "elmLS"
-[language.elm.settings.elmLS]
-# See https://github.com/elm-tooling/elm-language-server#server-settings
-runtime = "node"
-elmPath = "elm"
-elmFormatPath = "elm-format"
-elmTestPath = "elm-test"
-
-[language.elvish]
-filetypes = ["elvish"]
-roots = [".git", ".hg"]
-command = "elvish"
-args = ["-lsp"]
-
-[language.erlang]
-filetypes = ["erlang"]
-# See https://github.com/erlang-ls/erlang_ls.git for more information and
-# how to configure. This default config should work in most cases though.
-roots = ["rebar.config", "erlang.mk", ".git", ".hg"]
-command = "erlang_ls"
-
-[language.go]
-filetypes = ["go"]
-roots = ["Gopkg.toml", "go.mod", ".git", ".hg"]
-command = "gopls"
-settings_section = "gopls"
-[language.go.settings.gopls]
-# See https://github.com/golang/tools/blob/master/gopls/doc/settings.md
-# "build.buildFlags" = []
-
-[language.haskell]
-filetypes = ["haskell"]
-roots = ["hie.yaml", "cabal.project", "Setup.hs", "stack.yaml", "*.cabal"]
-command = "haskell-language-server-wrapper"
-args = ["--lsp"]
-settings_section = "_"
-[language.haskell.settings._]
-# See https://haskell-language-server.readthedocs.io/en/latest/configuration.html
-# haskell.formattingProvider = "ormolu"
-
-[language.html]
-filetypes = ["html"]
-roots = ["package.json"]
-command = "vscode-html-languageserver"
-args = ["--stdio"]
-
 # # Commented out by default because you still need to set the paths in the JDT
 # # Language Server arguments below before this can become a valid configuration.
-# [language.java]
+# [language_server.ejdtls]
 # filetypes = ["java"]
 # roots = [".git", "mvnw", "gradlew"]
 # command = "java"
@@ -152,21 +113,83 @@ args = ["--stdio"]
 #     "-data",
 #     "/path/to/eclipse-workspace",
 # ]
-# [language.java.settings]
+# [language_server.ejdtls.settings]
 # # See https://github.dev/eclipse/eclipse.jdt.ls
 # # "java.format.insertSpaces" = true
 
-[language.javascriptreact]
-filetypes = ["javascript"]
-roots = ["package.json", "tsconfig.json", ".git", ".hg"]
-command = "typescript-language-server"
+[language_server.elixir-ls]
+filetypes = ["elixir"]
+roots = ["mix.exs"]
+command = "elixir-ls"
+settings_section = "elixirLS"
+[language_server.elixir-ls.settings.elixirLS]
+# See https://github.com/elixir-lsp/elixir-ls/blob/master/apps/language_server/lib/language_server/server.ex
+# dialyzerEnable = true
+
+[language_server.elm-language-server]
+filetypes = ["elm"]
+roots = ["elm.json"]
+command = "elm-language-server"
+args = ["--stdio"]
+settings_section = "elmLS"
+[language_server.elm-language-server.settings.elmLS]
+# See https://github.com/elm-tooling/elm-language-server#server-settings
+runtime = "node"
+elmPath = "elm"
+elmFormatPath = "elm-format"
+elmTestPath = "elm-test"
+
+[language_server.elvish]
+filetypes = ["elvish"]
+roots = [".git", ".hg"]
+command = "elvish"
+args = ["-lsp"]
+
+[language_server.erlang-ls]
+filetypes = ["erlang"]
+# See https://github.com/erlang-ls/erlang_ls.git for more information and
+# how to configure. This default config should work in most cases though.
+roots = ["rebar.config", "erlang.mk", ".git", ".hg"]
+command = "erlang_ls"
+
+[language_server.gopls]
+filetypes = ["go"]
+roots = ["Gopkg.toml", "go.mod", ".git", ".hg"]
+command = "gopls"
+[language_server.gopls.settings.gopls]
+# See https://github.com/golang/tools/blob/master/gopls/doc/settings.md
+# "build.buildFlags" = []
+
+[language_server.haskell-language-server]
+filetypes = ["haskell"]
+roots = ["hie.yaml", "cabal.project", "Setup.hs", "stack.yaml", "*.cabal"]
+command = "haskell-language-server-wrapper"
+args = ["--lsp"]
+settings_section = "_"
+[language_server.haskell-language-server.settings._]
+# See https://haskell-language-server.readthedocs.io/en/latest/configuration.html
+# haskell.formattingProvider = "ormolu"
+
+[language_server.html-language-server]
+filetypes = ["html"]
+roots = ["package.json"]
+command = "vscode-html-languageserver"
 args = ["--stdio"]
 settings_section = "_"
-[language.javascriptreact.settings._]
+[language_server.html-language-server.settings._]
 # quotePreference = "single"
 # javascript.format.semicolons = "insert"
 
-[language.json]
+[language_server.intelephense]
+filetypes = ["php"]
+roots = [".htaccess", "composer.json"]
+command = "intelephense"
+args = ["--stdio"]
+settings_section = "intelephense"
+[language_server.intelephense.settings.intelephense]
+storagePath = "/tmp/intelephense"
+
+[language_server.json-language-server]
 filetypes = ["json"]
 roots = ["package.json"]
 command = "vscode-json-languageserver"
@@ -175,7 +198,7 @@ args = ["--stdio"]
 # Requires Julia package "LanguageServer"
 # Run: `julia --project=@kak-lsp -e 'import Pkg; Pkg.add("LanguageServer")'` to install it
 # Configuration adapted from https://github.com/neovim/nvim-lspconfig/blob/bcebfac7429cd8234960197dca8de1767f3ef5d3/lua/lspconfig/julials.lua
-[language.julia]
+[language_server.julia-language-server]
 filetypes = ["julia"]
 roots = ["Project.toml", ".git", ".hg"]
 command = "julia"
@@ -210,7 +233,7 @@ args = [
     run(server);
     """,
 ]
-[language.julia.settings]
+[language_server.julia-language-server.settings]
 # See https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/requests/workspace.jl
 # Format options. See https://github.com/julia-vscode/DocumentFormat.jl/blob/master/src/DocumentFormat.jl
 # "julia.format.indent" = 4
@@ -219,12 +242,125 @@ args = [
 # Other options, see https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/requests/workspace.jl
 # "julia.lint.run" = "true"
 
-[language.latex]
+[language_server.lua-language-server]
+filetypes = ["lua"]
+roots = [".git", ".hg"]
+command = "lua-language-server"
+settings_section = "Lua"
+[language_server.lua-language-server.settings.Lua]
+# See https://github.com/sumneko/vscode-lua/blob/master/setting/schema.json
+# diagnostics.enable = true
+
+[language_server.nimlsp]
+filetypes = ["nim"]
+roots = ["*.nimble", ".git", ".hg"]
+command = "nimlsp"
+
+[language_server.ocamllsp]
+filetypes = ["ocaml"]
+# Often useful to simply do a `touch dune-workspace` in your project root folder if you have problems with root detection
+roots = ["dune-workspace", "dune-project", "Makefile", "opam", "*.opam", "esy.json", ".git", ".hg", "dune"]
+command = "ocamllsp"
+
+[language_server.pls]
+filetypes = ["protobuf"]
+roots = [".git", ".hg"]
+command = "pls" # https://github.com/lasorda/protobuf-language-server
+
+[language_server.pylsp]
+filetypes = ["python"]
+roots = ["requirements.txt", "setup.py", ".git", ".hg"]
+command = "pylsp"
+settings_section = "_"
+[language_server.pylsp.settings._]
+# See https://github.com/python-lsp/python-lsp-server#configuration
+# pylsp.configurationSources = ["flake8"]
+pylsp.plugins.jedi_completion.include_params = true
+
+[language_server.r-language-server]
+filetypes = ["r"]
+roots = ["DESCRIPTION", ".git", ".hg"]
+command = "R"
+args = ["--slave", "-e", "languageserver::run()"]
+
+[language_server.racket-language-server]
+filetypes = ["racket"]
+roots = ["info.rkt"]
+command = "racket"
+args = ["-l", "racket-langserver"]
+
+[language_server.reason-ocamllsp]
+filetypes = ["reason"]
+roots = ["package.json", "Makefile", ".git", ".hg"]
+command = "ocamllsp"
+
+# [language_server.rls]
+# filetypes = ["rust"]
+# roots = ["Cargo.toml"]
+# command = "sh"
+# args = [
+#     "-c",
+#     """
+#         if path=$(rustup which rls 2>/dev/null); then
+#             "$path"
+#         else
+#             rls
+#         fi
+#     """,
+# ]
+# settings_section = "rust"
+# [language_server.rls.settings.rust]
+# # See https://github.com/rust-lang/rls#configuration
+# # features = []
+
+[language_server.rnix-lsp]
+filetypes = ["nix"]
+roots = ["flake.nix", "shell.nix", ".git", ".hg"]
+command = "rnix-lsp"
+
+[language_server.rust-analyzer]
+filetypes = ["rust"]
+roots = ["Cargo.toml"]
+command = "sh"
+args = [
+    "-c",
+    """
+        if path=$(rustup which rust-analyzer 2>/dev/null); then
+            "$path"
+        else
+            rust-analyzer
+        fi
+    """,
+]
+[language_server.rust-analyzer.settings.rust-analyzer]
+# See https://rust-analyzer.github.io/manual.html#configuration
+hoverActions.enable = false # kak-lsp doesn't support this at the moment
+# cargo.features = []
+
+[language_server.solargraph]
+filetypes = ["ruby"]
+roots = ["Gemfile"]
+command = "solargraph"
+args = ["stdio"]
+settings_section = "_"
+[language_server.solargraph.settings._]
+# See https://github.com/castwide/solargraph/blob/master/lib/solargraph/language_server/host.rb
+# diagnostics = false
+
+[language_server.terraform-ls]
+filetypes = ["terraform"]
+roots = ["*.tf"]
+command = "terraform-ls"
+args = ["serve"]
+[language_server.terraform-ls.settings.terraform-ls]
+# See https://github.com/hashicorp/terraform-ls/blob/main/docs/SETTINGS.md
+# rootModulePaths = []
+
+[language_server.texlab]
 filetypes = ["latex"]
 roots = [".git", ".hg"]
 command = "texlab"
-settings_section = "texlab"
-[language.latex.settings.texlab]
+[language_server.texlab.settings.texlab]
 # See https://github.com/latex-lsp/texlab/wiki/Configuration
 #
 # Preview configuration for zathura with SyncTeX search.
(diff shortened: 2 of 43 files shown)


sclu1034:
A heads-up, as you might want to keep this in consideration early, depending on how apt you're with amending and rebasing:

In the interest of keeping `git blame` useful and sane, every individual commit should successfully compile by itself.
It's not as important while the PR is still in draft, but the above should be ensured before it is merged.

src/context.rs
```diff
@@ -33,76 +34,91 @@ pub struct OutstandingRequests {
...
+        BatchNumber,
+        (
+            Vec<(ServerName, serde_json::value::Value)>,
+            ResponsesCallback,
+        ),
+    >,
+    pub code_lenses: HashMap<String, Vec<(ServerName, CodeLens)>>,
+    pub completion_items: Vec<(ServerName, CompletionItem)>,
     pub completion_items_timestamp: i32,
     // We currently only track one client's completion items, to simplify cleanup (else we
```
krobelus, line 66: Impressive work! A couple of comments (I didn't review the whole thing yet):

To optimzie the changes for the reader, can you structure the history so that every commit passes all tests (while being as small as practically possible).
See https://www.bitsnbites.eu/git-history-work-log-vs-recipe/
In particular, I don't care about any "fix" commits that fix bugs in your earlier commits. They should be squashed.

I'd probaly make one commit that extracts the `ServerSettings` type (and makes sure everything still compiles) without changing any behavior.
Then the actual logic changes will be much smaller.

> [language.typescriptreact]

While this may work in practice, I don't think this is kosher.
The "typescriptreact" part is sent to the language server as language ID.
The default config use "jsx" and "tsx" because those happen to work for "javascript" resp. "typescript" with the language servers I tried.

> Breaking changes

Don't. An old config file should keep working. Perhaps we want to make the parser accept both styles.

> [language.typescriptreact.tsserver]
> filetypes = ["javascript", "typescript"]
> roots = ["package.json", "tsconfig.json", "jsconfig.json", ".git", ".hg"]
> command = "typescript-language-server"
> args = ["--stdio"]
> 
> [language.typescriptreact.tailwindcss]
> filetypes = ["javascript", "typescript"]
> roots = ["tailwind.config.js", "tailwind.config.ts"]
> command = "tailwindcss-language-server"
> args = ["--stdio"]
> settings_section = "tailwindCSS"

what's wrong with

	[language.tsserver]
	filetypes = ["javascript", "typescript"]
	...
	[language.tailwindcss]
	filetypes = ["javascript", "typescript"]
	...

src/language_features/formatting.rs
```diff
@@ -1,31 +1,77 @@
...
+                let cmd = if meta.fifo.is_some() {
+                    "lsp-formatting-sync"
+                } else {
+                    "lsp-formatting"
+                };
+                let cmd = format!("{} {}", cmd, server_name);
+                format!("{} {}", editor_quote(server_name), editor_quote(&cmd))
+            })
+            .join(" ");
+        ctx.exec(meta, format!("lsp-menu {}", choices));
```
krobelus, line 44: I'm not sure about this. Probably the user should decide this earlier.
Also we'll want a non-interactive version for hooks.

The same problem exists for many other requests besides formatting, right?

Need to think about which ones have such conflicts.
An alternative solution is to have a primary server for such ambiguous requests.
I wonder what fellow LSP clients do.

src/language_features/hover.rs
```diff
@@ -49,28 +77,43 @@ pub fn editor_hover(
...
                     let message = (x.message.trim().to_string()
-                        + &format_related_information(x, ctx)
+                        + &format_related_information(x, (server_name, server), server, ctx)
                             .map(|s| "\n  ".to_string() + &s)
                             .unwrap_or_default())
                         .replace('\n', "\n  ");
                     if for_hover_buffer {
                         // We are typically creating Markdown, so use a standard Markdown enumerator.
-                        return format!("* {}", message);
+                        return format!("* [{server_name}] {message}");
```
krobelus, line 116: this still shows "c_cpp" here which is not the server name I think.
Also if there is only one server (common case) we don't want to name it here

src/context.rs
```diff
@@ -33,76 +34,91 @@
...
-    pub root_path: String,
     pub session: SessionId,
-    pub documents: HashMap<String, Document>,
-    pub offset_encoding: OffsetEncoding,
-    pub preferred_offset_encoding: Option<OffsetEncoding>,
     pub work_done_progress: HashMap<NumberOrString, Option<WorkDoneProgressBegin>>,
     pub work_done_progress_report_timestamp: time::Instant,
-    pub pending_file_watchers: HashMap<Option<PathBuf>, Vec<CompiledFileSystemWatcher>>,
+    pub pending_file_watchers:
+        HashMap<(ServerName, String, Option<PathBuf>), Vec<CompiledFileSystemWatcher>>,
```
krobelus, line 87: The "To server" and "From server" logs need to be adapted. Right now I can't tell which server the duplicate requests are directed at.

I noticed this when I tried

	[language.foo.c_cpp]
	filetypes = ["c", "cpp"]
	roots = ["compile_commands.json", ".clangd", ".git", ".hg"]
	command = "clangd"
	[language.foo.protobuf]
	filetypes = ["cpp"]
	roots = ["compile_commands.json", ".clangd", ".git", ".hg"]
	command = "pls"

trying to get go-to-definition for protobuf messages inside C++.
Sadly it doesn't work, clangd always responds before [pls](https://github.com/lasorda/protobuf-language-server/)

I think this one we could solve by ranking `pls` before `clangd`.
Defining `pls` as "primary" server feels odd but would work here.

src/context.rs
gbrlsnchs: > Impressive work! A couple of comments (I didn't review the whole thing yet):
> 
> To optimzie the changes for the reader, can you structure the history so that every commit passes all tests (while being as small as practically possible). See https://www.bitsnbites.eu/git-history-work-log-vs-recipe/ In particular, I don't care about any "fix" commits that fix bugs in your earlier commits. They should be squashed.

Sure thing! I just wanted to wait for feedback on some decisions, so I can work around tests based on that feedback. Once I get tests passing I'll squash everything together.

> I'd probaly make one commit that extracts the `ServerSettings` type (and makes sure everything still compiles) without changing any behavior. Then the actual logic changes will be much smaller.

This sounds good, will try it!

> > [language.typescriptreact]
> 
> While this may work in practice, I don't think this is kosher. The "typescriptreact" part is sent to the language server as language ID. The default config use "jsx" and "tsx" because those happen to work for "javascript" resp. "typescript" with the language servers I tried.

What led me to use that was the TailwindCSS server not working without a proper language ID. Then I followed the [suggested values in the specification]( https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentItem) and it started working.

> > Breaking changes
> 
> Don't. An old config file should keep working. Perhaps we want to make the parser accept both styles. (EDIT: FWIW an example of a deserializer that accepts two flavors of TOML structures is [0c10b3c](https://github.com/kak-lsp/kak-lsp/commit/0c10b3c9d072c8e3af0e217f5ee54e31a231dc62). Thankfully we never needed to merge that. Maybe we also don't need ambiguity here)
> 
> > [language.typescriptreact.tsserver]
> > filetypes = ["javascript", "typescript"]
> > roots = ["package.json", "tsconfig.json", "jsconfig.json", ".git", ".hg"]
> > command = "typescript-language-server"
> > args = ["--stdio"]
> > [language.typescriptreact.tailwindcss]
> > filetypes = ["javascript", "typescript"]
> > roots = ["tailwind.config.js", "tailwind.config.ts"]
> > command = "tailwindcss-language-server"
> > args = ["--stdio"]
> > settings_section = "tailwindCSS"
> 
> what's wrong with
> 
> ```
> [language.tsserver]
> filetypes = ["javascript", "typescript"]
> ...
> [language.tailwindcss]
> filetypes = ["javascript", "typescript"]
> ...
> ```

The problem is that we need to use a "valid" language ID, because some servers won't work without them (as mentioned, TailwindCSS only works with specific language IDs from [the specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentItem)).

I agree breaking a user's config is bad, and I would prefer an approach that doesn't as well. I thought we could go with something like:
```toml
[language.tsserver]
language_id = "typescriptreact"
filetypes = ["javascript", "typescript"]

[language.tailwindcss]
language_id = "typescriptreact"
filetypes = ["javascript", "typescript"]
```

In the example above, `language_id` would be optional, and if not present, the `language.<name>` would be used instead (like it currently is).

My biggest gripe with that approach is terminology, as `language` was previously attached to a language ID, and now it is attached to a language server name.

What I propose is that we have two hashmaps --- `language` and `language_server` --- and that they get merged together when config loads. This way we can avoid a breaking change by making `language` and `language_server` aliases, while also allowing the correct terminology to be used in newer versions of the config.

Here's an example of a valid config:

```toml
# This is the old config the user already has,
# and this shall not break by any means.
[language.zig]
filetypes = ["zig"]
command = "zls"

# This is the new config the user can now use.
[language_server.tsserver]
language_id = "typescriptreact"
filetypes = ["javascript", "typescript"]
command = "..."
```

src/language_features/formatting.rs
gbrlsnchs: Indeed, although rarer for other methods and much more common for formatting, this race condition problem exists.

Note that, if only a single server is used, then there is no prompt anyway, so this would be a breaking change for users.

IMO, having a prompt for picking the formatting server is a good default UX for a multi-server setup (just reminding that the prompt won't be shown to most users, who will still use a single server).

Having priorities configured is also good UX IMO, and I think both can co-exist if the user doesn't want a prompt or if the user doesn't want to set up `lsp-formatting <server>` for languages that use multiple servers.

For users that desire to use `lsp-formatting` and the likes in a non-interactive way with multiple servers, one can inform such commands the desired server (for example `lsp-formatting tsserver`). By passing the server name to the command, there will never be a prompt.

src/language_features/hover.rs
gbrlsnchs: Yeah, I can add logic to omit such labels when there's a single server.

As for capabilities, are you okay with the output being ordered? I thought it'd be more consistent and easier to handle.

src/context.rs
gbrlsnchs: Can kak-lsp currently handle gotos for multiple results? I know tsserver does return multiple results at some times. If so, we can go with merging the results instead of picking the first non-empty one. I think ranking them is not the ideal solution, because we would omit results anyway. What do you think?

gbrlsnchs: Btw, I just remembered, the way I set up this was to use a `BTreeMap` for servers, therefore currently they are lexicographically ordered, but it should be simple to add a custom ordering logic that respects a ranking value.

As for gotos, it's possible to merge results, but on Kakoune side, it uses a single root. I think this root should come from the "priority" server.

krobelus: > Can kak-lsp currently handle gotos for multiple results?

yeah in this case it will call `lsp-show-goto-choices` which shows a location list buffer.
We could maybe make that call lsp-menu instead

> If so, we can go with merging the results instead of picking the first non-empty one. I think ranking them is not the ideal solution, because we would omit results anyway. What do you think?

AFAICT the current behavior is the "first non-empty one" meaning it's a race between all language servers.
It would be good to make this deterministic. The simplest approach would be to wait for all responses (perhaps with a timeout) and then combine them. If that's too much work it can be a follow-up but let's keep that in mind.

> Btw, I just remembered, the way I set up this was to use a BTreeMap for servers, therefore currently they are lexicographically ordered, but it should be simple to add a custom ordering logic that respects a ranking value.

Good. We have two possible behaviors, either give the highest-ranked non-empty response, or merge all responses.
For my use case I'd like the first. Maybe there are uses for the second.

src/language_features/hover.rs
krobelus: > As for capabilities, are you okay with the output being ordered? I thought it'd be more consistent and easier to handle.

you mean the `(serverA, serverB)` labels in lsp-capabilities output? Probably ideal to use the input order (or input ranking once we have that) but alphabetic order is okay too

src/context.rs
krobelus: your proposal with `language_server` instead of `language` makes a ton of sense - there can be multiple servers for the same language.

krobelus: you're right, `typescriptreact` is the correct language ID. Here's an idea, maybe we want to write

```toml
[[language.typescriptreact]]
command = "typescript-language-server"
...
[[language.typescriptreact]]
command = "tailwindcss-language-server"
...
...
```

haven't tried it so not sure

gbrlsnchs: > you're right, typescriptreact is the correct language ID. Here's an idea, maybe we want to write
>
> ```toml
> [[language.typescriptreact]]
> command = "typescript-language-server"
> ...
> [[language.typescriptreact]]
> command = "tailwindcss-language-server"
> ...
> ...
> ```
> haven't tried it so not sure

This one is elegant. The only change we'd need to make is adding an optional `server` field, so we can identify each server. It needs to be optional in order to still be backwards compatible. If no server name is informed, we can use the language ID, and if it's a list, we can append an index to it (resulting in `(typescriptreact-1, typescriptreact-2)` for example). It's ugly but not the end of the world, and we can always provide examples in docs explaining that passing a `server` makes info output better. Such behavior can be achieved by using a custom deserializer, allowing the following:

```toml
# Existing config.
[language.rust]
filetypes = ["rust"]
command = "rust-analyzer"

[[language.typescriptreact]]
server = "tsserver"
filetypes = ["javascript", "typescript"]
command = "typescript-language-server"
args = ["--stdio"]

[[language.typescriptreact]]
server = "tailwindcss"
filetypes = ["javascript", "typescript"]
command = "tailwindcss-language-server"
args = ["--stdio"]
```

src/language_features/hover.rs
gbrlsnchs: I mean the capabilities per se. Previously, they were arbitrarily ordered (or maybe there was an implicit order), but because I had to use their names as a map key, I made my life easier by making the output consistent by using a BTreeMap.

If you wish the original order, I think we'd need to use something like [`indexmap`](https://github.com/bluss/indexmap).

src/context.rs
gbrlsnchs: I pushed a commit that merges the results. For goto specifically, I think it's better UX to present all possible results. The logic for choosing the highest-ranked would be hard to mix with the merging strategy.

src/context.rs
gbrlsnchs: @krobelus I tried implementing the approach above and it doesn't work well with TOML, specially when defining `settings` for a language server.

I think maybe the aliasing approach is simpler and easier to maintain.

gbrlsnchs: @krobelus I tried the approach above and 1) it didn't work properly and 2) is difficult to implement and hard to maintain.

Maybe we could go with the aliasing approach? It's still backwards compatible and is a lot cleaner.

krobelus: I don't see why it wouldn't work but we can handle this later. It doesn't block other work and changes to this will be small and isolated

src/language_features/formatting.rs
krobelus: ok a prompt sounds like a reasonable default UI.
As for how to force a specific set of servers, I'm not sure if arguments are the best way to do this.
Maybe we should have a `str-list` option that users can set before their request..

src/session.rs
```diff
@@ -130,10 +137,11 @@ pub fn start(config: &Config, initial_request: Option<String>) -> i32 {
                         // get didClose message without running controller, unless it crashed
                         // before. In that case didClose can be safely ignored as well.
                         if request.method != notification::DidCloseTextDocument::METHOD {
-                            debug!("Spawning a new controller for {:?}", route);
+                            debug!("Spawning a new controller for {:#?}", routes);
                             controller_entry.insert(spawn_controller(
                                 config.clone(),
-                                route,
+                                language_id.clone(),
```
krobelus, line 143: Haven't had time for a full review, some comments in passing:
I'm surprised we only spawn a controller for the first server.
Won't that lead to weird behavior if the configured servers have different root paths or language IDs?

src/session.rs
```diff
@@ -211,9 +219,14 @@
         "Editor session `{}` closed, shutting down associated language servers",
         request.meta.session
     );
-    controllers.retain(|route, controller| {
-        if route.session == request.meta.session {
-            info!("Exit {} in project {}", route.language, route.root);
+    controllers.retain(|routes, controller| {
+        if routes
+            .iter()
```
krobelus, line 224: We should add an assertion somewhere that either all routes are from this session, or none is.
Because if only some are, then the behavior is weird.

src/types.rs
```diff
@@ -14,7 +14,8 @@ pub enum Void {}
 
 #[derive(Clone, Deserialize, Debug)]
 pub struct Config {
-    pub language: HashMap<String, LanguageConfig>,
+    #[serde(alias = "language_server")]
+    pub language: HashMap<ServerName, LanguageServerConfig>,
```
krobelus, line 18: it would probably be better to rename this to `language_server` (also in the default kak-lsp.toml) and add the alias for the legacy name.
That can all go in the parent commit.

rc/lsp.kak
```diff
@@ -1758,11 +1760,11 @@ define-command -hidden lsp-show-message-info -params 1 -docstring %{
     }
 }
 
-define-command -hidden lsp-show-message-log -params 1 -docstring %{
+define-command -hidden lsp-show-message-log -params 2 -docstring %{
```
krobelus, line 1763: I'm not yet sure if this update is a good idea (because it might break users and complicate the interface for a niche use case).
If it is worth it, the other lsp-show-message-* should be updated in tandem.

gbrlsnchs: When I was testing a setup with 3 different servers, this certainly helped with identifying which server was which in the `*debug*` buffer.

gbrlsnchs:
@krobelus I did my best to split commits in a meaningful way. After I split them, I rebased them so I could edit them in order to make ad-hoc changes that would make each commit buildable, which then I tested by running `git rebase -x 'cargo check' upstream/master`.

Here's how I feel after that:
image

There are some items in my to-do list in the PR description, which I'll now resume working on.

One major change I did was using filetypes as language IDs (instead of server name), and then I added a `language_ids` section at the config root that allows overriding language IDs if needed. That makes transition from old config even smoother, since there's no need to add a new field to make things work.

src/session.rs
gbrlsnchs: It's not exactly for the first server, as the `controllers` key is now a vector of routes, and not a single route anymore. So in reality it spawns a controller for the first time a route sequence is seen by kak-lsp.

src/types.rs
gbrlsnchs: This has been addressed in the major commit overhaul I did.

rc/lsp.kak
krobelus: ok sorry I just searched on github; no one overrides these commands, so let's just update the other `lsp-show-message-*` commands as well

src/language_features/hover.rs
krobelus: it's fine to sort capabilities, the previous order seems pretty random

src/types.rs
```diff
@@ -23,12 +24,14 @@ pub struct Config {
     pub snippet_support: bool,
     #[serde(default)]
     pub semantic_tokens: SemanticTokenConfig,
+    #[serde(default)]
```
krobelus, line 27: great idea! This is a nice simplification of our config.

src/types.rs
```diff
@@ -165,12 +169,13 @@
...
 pub type SessionId = String;
 pub type LanguageId = String;
+pub type ServerName = String;
 pub type RootPath = String;
 
 #[derive(PartialEq, Eq, Hash, Debug, Clone)]
 pub struct Route {
     pub session: SessionId,
-    pub language: LanguageId,
+    pub server_name: ServerName,
```
krobelus, line 178: (nit: the removal of `language` should happen in this commit (it's replaced by `server_name`))

src/capabilities.rs
```diff
@@ -527,80 +580,118 @@ pub fn server_has_capability(ctx: &Context, feature: &'static str) -> bool {
...
-                .join(", ")
-        ));
+        if let Some(ref provider) = server_capabilities.semantic_tokens_provider {
+            let legend = match provider {
+                SemanticTokensServerCapabilities::SemanticTokensOptions(options) => &options.legend,
+                SemanticTokensServerCapabilities::SemanticTokensRegistrationOptions(regopts) => {
+                    &regopts.semantic_tokens_options.legend
+                }
+            };
+
```
krobelus, line 652: Using the first server's offset encoding is wrong.
I noticed you correct this in a later commit.
But this has too much potential for confusion that I will continue reading the sum of all changes rather than the individual commits.
If you want to fix this structure (make sure that every commit passes all tests and doesn't have known flaws) that would help but else I can do this, no problem.
I appreciate the amount of work that has gone into this, it's a great exercise.

README.asciidoc
```diff
@@ -304,22 +304,52 @@ Please let us know if you have any ideas about how to make the default config mo
...
+[language_server.gopls.settings.gopls]
 "formatting.gofumpt" = true
 ----
 
-During server initialization, kak-lsp sends the section specified by `settings_section`; in this
-case `{"formatting.gofumpt":true}`.  Whenever you change the Kakoune option `lsp_config`, the
-same section is sent via `workspace/didChangeConfiguration`.  Additionally, kak-lsp will send
-arbitrary sections that are requested by the server in `workspace/configuration`.
+During server initialization, kak-lsp sends either the section specified by `settings_section`
+or the section with same name as the server; in this case `{"formatting.gofumpt":true}`.
```
krobelus, line 319: > or the section with same name as the server

Hmmm, really? IIRC `settings_section` has no default value.
If I'm correct, then we need to add a `settings_section` to the `tailwindcss` example.
(Unfortunately the whole dynamic settings passing is super weird due to the lack of standardization in LSP.)

README.asciidoc
```diff
@@ -304,22 +304,52 @@
...
+==== Multiple language servers
+
+It is possible to map more than one language server to filetypes. For example, if you want to
+set up TSServer and TailwindCSS to use in React projects:
+
+[source,toml]
+----
+[language_ids]
+javascript = "javascriptreact"
+typescript = "typescriptreact"
```
krobelus, line 333: I made another comment suggesting default language ID mappings.
Let's make sure this remains in sync with the default language ID mapping in kak-lsp.toml.

README.asciidoc
```diff
@@ -304,22 +304,52 @@
...
+
+[language_server.tsserver]
+filetypes = ["javascript", "typescript"]
+roots = ["package.json", "tsconfig.json", "jsconfig.json", ".git", ".hg"]
+command = "typescript-language-server"
+args = ["--stdio"]
+
+[language_server.tailwindcss]
+filetypes = ["javascript", "typescript"]
+roots = ["tailwind.config.ts", "tailwind.config.js"]
```
krobelus, line 343: are these config files required?

README.asciidoc
```diff
@@ -304,22 +304,52 @@
...
+command = "tailwindcss-language-server"
+args = ["--stdio"]
```
krobelus, line 345: nit: If other editors enable tailwindcss by default (no idea if they do) maybe we should too

kak-lsp.toml
```diff
@@ -6,163 +6,184 @@ verbosity = 2
 # set to 0 to disable
 timeout = 1800 # seconds = 30 minutes
 
-[language.bash]
+# This section overrides language IDs.
+# By default, kak-lsp use filetypes for the IDs.
+# See https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentItem
+[language_ids]
```
krobelus, line 12: Hypothetically, if a user wanted to use clangd for the `c` and `c++` filetype but with different parameters each, that'd be possible now.
They can define `language_server.clangd-for-c` and `language_server.clangd-for-cpp`.
Nice.

kak-lsp.toml
```diff
@@ -6,163 +6,184 @@
...
+# typescript = "typescriptreact"
+
+[language_server.bash-language-server]
 filetypes = ["sh"]
 roots = [".git", ".hg"]
 command = "bash-language-server"
 args = ["start"]
 
-[language.c_cpp]
+[language_server.clangd]
```
krobelus, line 24: we used to send `c_cpp` always. Maybe we should keep doing that because I'm not sure if all *.h files are detected correctly.

kak-lsp.toml
```diff
@@ -6,163 +6,184 @@
...
-#     "-configuration",
-#     "/path/to/eclipse.jdt.ls/repository/config_linux",
-#     "-data",
-#     "/path/to/eclipse-workspace",
-# ]
-# [language.java.settings]
-# # See https://github.dev/eclipse/eclipse.jdt.ls
-# # "java.format.insertSpaces" = true
-
-[language.jsx] # works for javascript as well
```
krobelus, line 159: The reason why this uses `jsx` instead of `javascript` is a bit hidden,
because it was documented only for [`tsx` instead of `typescript`](https://github.com/kak-lsp/kak-lsp/commit/e8f109d7da52aa9a6c4586dcbc6d4bae2ae47c77).
Essentially `jsx` (perhaps that's the same as `javascriptreact`?) is a
superset of javascript, so it works for plain javascript.
On the other hand, `language_id=javascript` doesn't work for *.jsx files.
So I think what we should do here is add these mappings

	[language_ids]
	javascript = "jsx"
	typescript = "tsx"

unless there is a reason for changing the status quo.. I don't know much about frontend dev.
(I'm not sure if `javascriptreact` is better/worse)

kak-lsp.toml
```diff
@@ -351,47 +326,76 @@
...
 # rootModulePaths = []
 
-[language.tsx] # works for typescript as well
-filetypes = ["typescript"]
-roots = ["package.json", "tsconfig.json", ".git", ".hg"]
+[language_server.texlab]
+filetypes = ["latex"]
+roots = [".git", ".hg"]
+command = "texlab"
+[language_server.texlab.settings.texlab]
```
krobelus, line 357: I wonder if we should keep sorting by "primary filetype".
However I'm not sure, so no need to change this now.

rc/lsp.kak
```diff
@@ -1184,17 +1184,17 @@ $([ -z ${kak_hook_param+x} ] || echo hook = true)
 " | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }
 
-define-command lsp-formatting -docstring "Format document" %{
-    lsp-formatting-request false
+define-command lsp-formatting -params ..1 -docstring "Format document" %{
+    lsp-formatting-request false %arg{1}
```
krobelus, line 1188: Since this is public we should document the parameter in the docstring. Same below.
(Also in future we can equip this function with a `shell-script-candidates` completer that shows provides server names).

rc/lsp.kak
```diff
@@ -1226,17 +1227,17 @@
     fi
 }}
 
-define-command lsp-range-formatting -docstring "Format selections" %{
-    lsp-range-formatting-request false
+define-command lsp-range-formatting -params ..1 -docstring "Format selections" %{
```
krobelus, line 1230: let's use explicit `0..1` instead of `..1` for consistency with the rest of the code

src/context.rs
```diff
@@ -33,76 +42,83 @@ pub struct OutstandingRequests {
...
+    pub batch_sizes: HashMap<BatchNumber, HashMap<ServerName, usize>>,
+    pub batches: HashMap<
+        BatchNumber,
+        (
+            Vec<(ServerName, serde_json::value::Value)>,
+            ResponsesCallback,
+        ),
+    >,
+    pub code_lenses: HashMap<String, Vec<(ServerName, CodeLens)>>,
+    pub completion_items: Vec<(ServerName, CompletionItem)>,
```
krobelus, line 64: Have you given thought to moving these two into `ServerSettings`?
I guess that would be less convenient because of the way we forward it to the editor.

src/context.rs
```diff
@@ -111,58 +127,91 @@ impl Context {
...
+                    ops
+                })
+                .collect(),
+        };
         self.batch_call::<R, _>(
             meta,
             ops,
             Box::new(
-                move |ctx: &mut Context, meta: EditorMeta, mut results: Vec<R::Result>| {
-                    if let Some(result) = results.pop() {
```
krobelus, line 130: The old results.pop() seems super weird. Why would we not consume all of results?
Maybe it's just an `Option<R::Result>` in disguise...
Anyway, the new version seems fine though it would be good to understand this.
Maybe it's because we have `batch_sizes()` now.

src/controller.rs
```diff
@@ -93,8 +121,44 @@ pub fn start(
...
+                    }
+                },
+            )
+            .unwrap_or_else(never);
+
+        let mut sel = Select::new();
+        let server_rxs: Vec<&Receiver<ServerMessage>> = language_servers
+            .values()
+            .map(|(transport, _)| transport.from_lang_server.receiver())
+            .collect();
```
krobelus, line 148: I think `server_rxs` is loop-invariant, so we could initialize it before the loop?
I'd need to check.

src/controller.rs
```diff
@@ -104,65 +168,137 @@
                 // capabilities also serve as a marker of completing initialization
                 // we park all requests from editor before initialization is complete
                 // and then dispatch them
-                if ctx.capabilities.is_some() {
+                let parked: Vec<_> = ctx
```
krobelus, line 171: something for the future: if one server is parked and another one isn't, then we can already talk to some servers

src/controller.rs
```diff
@@ -171,69 +307,97 @@
...
+                                            == ErrorCode::ServerError(CONTENT_MODIFIED)
+                                            || method == request::CodeActionRequest::METHOD =>
+                                        {
                                             // Nothing to do, but sending command back to the editor is required to handle case when
                                             // editor is blocked waiting for response via fifo.
                                             ctx.exec(meta, "nop".to_string());
-                                        },
+                                        }
                                         code => {
+                                            // TODO: Detail messages for different failing servers.
```
krobelus, line 371: Is this TODO about the case where multiple servers return errors?

src/controller.rs
```diff
@@ -404,8 +568,12 @@
             goto::text_document_references(meta, params, ctx);
         }
         notification::Exit::METHOD => {
-            ctx.notify::<notification::Exit>(());
```
krobelus, line 407: nit: maybe we should have a `ctx.notify_all` helper method.

src/diagnostics.rs
```diff
@@ -11,12 +11,24 @@ use lsp_types::*;
...
+        .into_iter()
+        .filter(|(name, _)| name != server_name)
+        .collect();
+    let params: Vec<_> = params
+        .diagnostics
+        .into_iter()
+        .map(|d| (server_name.clone(), d))
+        .collect();
+    diagnostics.extend(params);
+    ctx.diagnostics.insert(buffile.to_string(), diagnostics);
```
krobelus, line 31: I guess this would be more elegant if we used `ctx.diagnostics.entry().or_insert()...` instead of remove() followed by insert(). Didn't try it though

src/diagnostics.rs
```diff
@@ -204,14 +221,16 @@
         write_response_to_fifo(meta, &ctx.diagnostics);
         return;
     }
+    let (_, main_settings) = ctx.language_servers.first_key_value().unwrap();
```
krobelus, line 224: if we compute a "most general root path" then we don't need to fall back to the main settings.
That can be done later though

src/diagnostics.rs
```diff
@@ -257,13 +283,14 @@
...
                     let p = get_kakoune_position_with_fallback(
+                        server,
                         filename,
                         info.location.range.start,
                         ctx,
                     );
                     format!(
-                        "{}:{}:{}: {}",
-                        short_file_path(filename, &ctx.root_path),
+                        "{}:{}:{}: ({server_name}) {}",
```
krobelus, line 292: I think we should only add the server name if there are multiple servers configured

src/language_features/call_hierarchy.rs
```diff
@@ -32,56 +44,77 @@ fn request_call_hierarchy(
     meta: EditorMeta,
     ctx: &mut Context,
     incoming_or_outgoing: bool,
-    result: Option<Vec<CallHierarchyItem>>,
+    results: Vec<(ServerName, Option<Vec<CallHierarchyItem>>)>,
 ) {
+    let result = results
+        .into_iter()
+        .find(|(_, response)| response.is_some())
+        .and_then(|(server_name, item)| Some((server_name, item.unwrap())));
```
krobelus, line 52: I don't think we should call `unwrap()` here, at least the previous code didn't.
Of course in practice, most language servers will probably send an empty vector but I don't think the protocol guarantees that.
I'd write something like

	and_then(|...| item.into_iter().map(|item| (server_name, item)));

src/language_features/call_hierarchy.rs
```diff
@@ -32,56 +44,77 @@
...
+                partial_result_params: PartialResultParams::default(),
+            }],
+        );
 
         ctx.call::<CallHierarchyIncomingCalls, _>(
             meta,
-            params,
-            move |ctx: &mut Context, meta, result| {
-                format_call_hierarchy_calls(meta, ctx, incoming_or_outgoing, &item, &result);
+            RequestParams::Each(params),
```
krobelus, line 75: Can we not use `RequestParams::All` here?

src/language_features/call_hierarchy.rs
```diff
@@ -32,56 +44,77 @@
...
+            move |ctx: &mut Context, meta, results| {
+                if let Some(result) = results.first() {
```
krobelus, line 77: ok so this shows the results of whichever server responds first.
This will probably need correction (what if one server always returns `[]`?).
It can be done in a follow-up. Maybe a priority, or maybe skip empty reslts..

src/language_features/code_action.rs
```diff
@@ -33,85 +41,144 @@ pub fn text_document_code_action(meta: EditorMeta, params: EditorParams, ctx: &m
             return;
         }
     };
-    let range = kakoune_range_to_lsp(
-        &parse_kakoune_range(&params.selection_desc).0,
-        &document.text,
-        ctx.offset_encoding,
-    );
-    code_actions_for_range(meta, params, ctx, range)
```
krobelus, line 41: this should be renamed to `code_actions_for_ranges`

src/language_features/code_action.rs
```diff
@@ -33,85 +41,144 @@
...
+            attempt_server_capability(
+                (server_name, server_settings),
+                &meta,
+                CAPABILITY_CODE_ACTIONS_RESOLVE,
+            )
+        })
+        .map(|(server_name, _)| server_name)
+        .collect();
 
+    // TODO: Should pattern contain the server's name?
```
krobelus, line 181: we can try that in future

src/language_features/formatting.rs
```diff
@@ -1,31 +1,77 @@
...
+    if eligible_servers.len() > 1 {
+        let choices = eligible_servers
+            .into_iter()
+            .map(|(server_name, _)| {
+                let cmd = if meta.fifo.is_some() {
+                    "lsp-formatting-sync"
+                } else {
+                    "lsp-formatting"
+                };
+                let cmd = format!("{} {}", cmd, server_name);
```
krobelus, line 40: should editor_quote `server_name`, just in case

src/language_features/rename.rs
```diff
@@ -11,26 +11,54 @@ use super::super::workspace;
...
+                    work_done_progress_params: Default::default(),
+                }],
+            )
+        })
+        .collect();
+    ctx.call::<Rename, _>(
+        meta,
+        RequestParams::Each(req_params),
+        move |ctx: &mut Context, meta, results| {
+            let result = match results.into_iter().find(|(_, v)| v.is_some()) {
```
krobelus, line 43: ok so it's safe to broadcast the rename request because we'll apply only the first response

krobelus:
This is really fantastic, well designed and executed. Thanks!
I'll merge soon. If you want to address some review comments, go ahead, else I'll do the important ones.

There's a broken system test, here's the fix:

```diff
diff --git a/test/gopls-dynamic-settings.sh b/test/gopls-dynamic-settings.sh
index 454b55b..e0da1e3 100755
--- a/test/gopls-dynamic-settings.sh
+++ b/test/gopls-dynamic-settings.sh
@@ -29,7 +29,7 @@ test_tmux capture-pane -p
 
 echo '
 set global lsp_config %{
-	[language.go.settings.gopls]
+	[language_server.gopls.settings.gopls]
 	"formatting.gofumpt" = true
 }' | kak -p "$test_kak_session"
 test_sleep
```

README.asciidoc
```diff
@@ -304,22 +304,52 @@ Please let us know if you have any ideas about how to make the default config mo
 ==== Server-specific configuration
 
 Many servers accept configuration options that are not part of the LSP spec.  The TOML table
-`[language.<filetype>.settings]` holds those configuration options.  It has the same structure
+`[language_server.<server_name>.settings]` holds those configuration options.  It has the same structure
```
krobelus, line 307: I believe that if the old format (`[language.c_cpp]`) is used, then we should keep sending `c_cpp` as languageId, not the filetype. This is to avoid breaking existing configs. Unless we find that all filetypes work fine but I doubt that (for example with typescript-language-server, languageId javascript didn't work for jsx files last time I checked).
To implement this, we probabl need a custom deserializer.

src/context.rs
krobelus: > For goto specifically, I think it's better UX to present all possible results.

yes let's show all of them for now, but remove duplicates

krobelus:
I'm actually using this in production now, it feels awesome (I have a silly setup where I want to use two python language servers at the same time).
One other thing I noticed is that we should send codeLens requests only to `eligible_servers`.

raiguard:
Initial impressions are very good, thanks for working on this! Only problem I've run into is related to completions sorting. When using [FMTK](https://github.com/justarandomgeek/vscode-factoriomod-debug) and [LuaLS](https://github.com/LuaLS/lua-language-server/) together in a Lua file, the FMTK completions are buried under a mountain of plain text completions. 

LuaLS doesn't return any completions in this situation, so I assume that kak-lsp is erroneously letting the plaintext completions show up because of that. If I remove LuaLS, then the FMTK completions work properly and no plaintext completions are shown.

image

gbrlsnchs:
Hey @krobelus, I'm a little busy lately and I haven't been able to give this PR some love. I just enabled maintainers to modify it as well, so please feel free to make adjustments as you seem fit!

gbrlsnchs:
> Initial impressions are very good, thanks for working on this! I hope we can add a combined diagnostics view to see all diagnostics for all servers in the future, but that can be a separate PR.

Hey @raiguard, thanks! I'm glad to help. Multiple diagnostics is already a thing. Each diagnostic is labeled by their respective language server. Is there a use case where they didn't properly show up to you?

> Turns out LuaLS returns text completions for strings, so this is theoretically working as intended. Something I would definitely do is sort text completions after all other completions, since more specific completions will undoubtedly be more useful.

This is something I also wish we had for kak-lsp. Maybe we can work on this as a future improvement, @krobelus? To let users choose their sorting priorities by item kind, maybe.

krobelus:
I'll tackle this next. Want to finish another thing first..

> Turns out LuaLS returns text completions for strings, so this is theoretically working as intended. Something I would definitely do is sort text completions after all other completions, since more specific completions will undoubtedly be more useful.

that first needs sorting support (as defined by LSP), which is added in https://github.com/kak-lsp/kak-lsp/pull/657 (needs a Kakoune patch as well)
Chances are it already works fine with that PR in most cases
If not, then it would be interesting to know how other editors solve it

README.asciidoc
krobelus: ok I think you're right, `settings_section` is optional

kak-lsp.toml
krobelus: ok looks like you're right; I've read elsewhere that `javascriptreact` is the proper language ID and not `jsx`

krobelus:
@gbrlsnchs thanks for your patience, this is just awesome.
Merging now; let's keep our eyes open for things that don't work properly when using multiple language servers.

PR #696, opened 2023-11-06
Add PureScript to the config

postsolar:


kak-lsp.toml
```diff
@@ -267,6 +267,12 @@ filetypes = ["protobuf"]
 roots = [".git", ".hg"]
 command = "pls" # https://github.com/lasorda/protobuf-language-server
 
+[language_server.purescript]
+filetypes = ["purescript"]
+roots = [".git", "package.json", "spago.dhall", "spago.yaml"]
```
krobelus, line 272: roots are processed from left to right, so if there is any parent directory (including grandparent etc.) with a `.git` dir, that parent will be used as root. That's why we usually put .git/.hg last.
Do package.json/spago.dhall/spago.yaml have the same risk of false positives that `.git` has?

postsolar: Spago files are uniquely PureScript projects roots. `package.json` is not unique to PS so I guess it should come third?

krobelus:
thanks! I made some cosmetic changes

postsolar:
Thank you!

Issue #297, opened 2020-03-16
support for scala metals ?


joeitu:
Does kak-lsp support scala metals ? 
I have installed metals through vscode, but it doesn't seems to work with kak-lsp.
I didn't found any doc on metals or kak-lsp pages.

joeitu:
related:
https://github.com/scalameta/metals-feature-requests/issues/91

topisani:
It appears from that, that the issue is `kak-lsp` not supporting `window/showMessageRequest`. That should be fairly easy to implement

ckipp01:
Thought I'd pipe in here as well after my comment in https://github.com/scalameta/metals-feature-requests/issues/91. I'd love to get kak documented and "supported", so I've gone ahead and started to test various features. I'm sure I'll find more as I play around, but I've got some initial thoughts and questions. I'll also add that I really enjoyed playing around with this so far :)

1. As was originally mentioned, `window/showMessageRequest` is important. This is what kicks off the entire process for metals to dump out the build definition that is needed to provide literally anything. We also use this for several other requests such as asking the user to add in a scalafmt version if they are missing one. Without it for example, `lsp-formatting` won't work. There is actually an issue for this already [here](https://github.com/ul/kak-lsp/issues/183), which is pretty old.

2. Metals relies pretty heavily on either `workspace/showMessage` being shown to the user. From what I can tell [here](https://github.com/ul/kak-lsp/blob/17fc6a6573251a77811ec7e75bf3807e24277cbb/src/controller.rs#L299-L311) the message should be shown via `lsp-show-message`, but I haven't seen any messages shown yet. Keep in mind I'm a noob with kak and kak-lsp, so I did the bare minimum to get everything up and running. Is there something specific I need to enable or add to see messages?

3. We also rely on the user being able to pass in configuration values. It looks like that's exactly what `lsp_server_configuration` is for, but I'm unable to get it to work. For example, I have the following in my `kakrc`.

```
hook global WinSetOption filetype=(scala) %{
  lsp-enable-window
  set global lsp_server_configuration metals.superMethodLenses=true
}
```
I would expect this to get set under `metals.superMethodLenses` which is what is illustrated in the readme, but the settings object is just empty for me. Is there something I'm missing here?

```json
[Trace - 09:34:03 PM] Received notification 'workspace/didChangeConfiguration'
Params: {
  "settings": {}
}
```

[...]

PR #697, opened 2023-11-07
Add Scala Metals LSP

igor-ramazanov:
Closes https://github.com/kak-lsp/kak-lsp/issues/297.

Created by following the [Scala Metals guide](https://scalameta.org/metals/docs/integrations/new-editor) and existing entries in the `kak-lsp.toml`.

### DIFF
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -77,7 +77,6 @@ roots = ["build.sbt", ".scala-build"]
 command = "metals"
 args = ["-Dmetals.extensions=false"]
 settings_section = "metals"
-
 [language_server.metals.settings.metals]
 icons = "unicode"
 isHttpEnabled = true
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -69,6 +69,21 @@ roots = ["package.json", ".git", ".hg"]
 command = "vscode-css-languageserver"
 args = ["--stdio"]
 
+
+# See https://scalameta.org/metals/docs/integrations/new-editor
+[language_server.metals]
+filetypes = ["scala", "sbt", "sc"]
+roots = ["build.sbt", ".scala-build"]
+command = "metals"
+args = ["-Dmetals.extensions=false"]
+settings_section = "metals"
+
+[language_server.metals.settings.metals]
+icons = "unicode"
+isHttpEnabled = true
+statusBarProvider = "log-message"
+compilerOptions = { overrideDefFormat = "unicode" }
+
 [language_server.scss-language-server]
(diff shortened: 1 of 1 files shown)


kak-lsp.toml
```diff
@@ -69,6 +69,21 @@ roots = ["package.json", ".git", ".hg"]
 command = "vscode-css-languageserver"
 args = ["--stdio"]
 
+
+# See https://scalameta.org/metals/docs/integrations/new-editor
+[language_server.metals]
+filetypes = ["scala", "sbt", "sc"]
```
krobelus, line 75: I'm not sure what `sbt` and `sc` are.
In general we only take filetypes that are also in Kakoune.
So if the plan is to add them to Kakoune's `scala.kak` then this is fine.
(Or the languages are the same maybe they should all get the `scala` filetype).

By default we send the filetype as LSP languageId.
So if metals works better/worse with "sbt" instead of "scala" that would be interesting to know.

kak-lsp.toml
```diff
@@ -69,6 +69,21 @@
...
+roots = ["build.sbt", ".scala-build"]
+command = "metals"
+args = ["-Dmetals.extensions=false"]
+settings_section = "metals"
+
```
krobelus, line 80: can we delete the empty line? To help `<a-i>p` ;)

kak-lsp.toml
igor-ramazanov: Aha, got it, didn't know `filetypes` are in the Kakoune terminology.

Assumed this field is for general `*.<file extension>` file extensions.

`scala`, `sbt` and `sc` are all Scala files just with different extension names (`sbt` is for [sbt](https://www.scala-sbt.org/index.html) and `sc` for [ScalaCLI](https://scala-cli.virtuslab.org/)).

Done!

kak-lsp.toml
igor-ramazanov: Sure, done

kak-lsp.toml
krobelus: ok then we might want

```diff
diff --git a/rc/filetype/scala.kak b/rc/filetype/scala.kak
index 0822a016b..05d7b28ee 100644
--- a/rc/filetype/scala.kak
+++ b/rc/filetype/scala.kak
@@ -4,7 +4,7 @@
 # Detection
 # ‾‾‾‾‾‾‾‾‾
 
-hook global BufCreate .*[.](scala) %{
+hook global BufCreate .*[.](scala|sbt|sc) %{
     set-option buffer filetype scala
 }
 ```

PR #704, opened 2023-12-26
Don't send didSave notifications if server doesn't want it

natsukagami:
This crashes [nil](https://github.com/oxalica/nil), since it defaults to crashing upon getting a notification it doesn't handle.


krobelus:
thanks; if this is the best-in-class language server we should add it to the default config

PR #706, opened 2023-12-28
Tree-like document symbols

igor-ramazanov:
Renders document symbols in a tree-like manner.

Example:

Before
```
links-fixer.scala:17:12:     Class    Link
links-fixer.scala:24:8:      Module   LinksFixer
  links-fixer.scala:26:7:    Method   main
  links-fixer.scala:29:15:   Method   fixFile
    links-fixer.scala:30:9:  Constant file
  links-fixer.scala:34:15:   Method   fix
  links-fixer.scala:37:15:   Method   parse
    links-fixer.scala:38:9:  Constant trimmed
    links-fixer.scala:39:15: Constant prefix
    links-fixer.scala:39:23: Constant title
    links-fixer.scala:40:9:  Constant fragment
    links-fixer.scala:44:9:  Constant fragmentless
    links-fixer.scala:45:10: Constant path
    links-fixer.scala:45:16: Constant file
  links-fixer.scala:51:15:   Method   render
```

After
```
links-fixer.scala:17:12:     Link                 Class
links-fixer.scala:24:8:      LinksFixer           Module
  links-fixer.scala:26:7:    ├── main             ├── Method
  links-fixer.scala:29:15:   ├── fixFile          ├── Method
    links-fixer.scala:30:9:  │   └── file         │   └── Constant
  links-fixer.scala:34:15:   ├── fix              ├── Method
  links-fixer.scala:37:15:   ├── parse            ├── Method
    links-fixer.scala:38:9:  │   ├── trimmed      │   ├── Constant
    links-fixer.scala:39:15: │   ├── prefix       │   ├── Constant
    links-fixer.scala:39:23: │   ├── title        │   ├── Constant
    links-fixer.scala:40:9:  │   ├── fragment     │   ├── Constant
    links-fixer.scala:44:9:  │   ├── fragmentless │   ├── Constant
    links-fixer.scala:45:10: │   ├── path         │   ├── Constant
    links-fixer.scala:45:16: │   └── file         │   └── Constant
  links-fixer.scala:51:15:   └── render           └── Method
```

**Story:**
So, I wanted to try out this output style to see how it would work for myself.
And it turned out to be good, it made it easier to read for me.
Definitely, going to continue using it.

**Concerns:**
1. Perhaps, previous style was there for a reason, e.g. to make it easy to parse the buffer in case if users would want to customize it using Kakoune itself. And probably, it already could be implemented with highlighters. I'm still learning Kakoune, so not 100% sure about this.
2. Visual cluttering?
3. Use of UTF-8 symbols instead of ASCII.

### DIFF
--- a/src/language_features/document_symbol.rs
+++ b/src/language_features/document_symbol.rs
@@ -9,6 +9,7 @@ use crate::position::{
 use crate::types::*;
 use crate::util::*;
 use crate::{context::*, position::get_file_contents};
+use core::fmt;
 use indoc::{formatdoc, indoc};
 use itertools::Itertools;
 use lsp_types::request::*;
@@ -206,6 +207,24 @@ fn editor_document_symbol(
     ctx.exec(meta, command);
 }
 
+enum Tree {
+    Child,
+    LastChild,
+    Pipe,
+    Empty,
+}
+
+impl fmt::Display for Tree {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Tree::Child => write!(f, "{}", "├── "),
+            Tree::LastChild => write!(f, "{}", "└── "),
+            Tree::Pipe => write!(f, "{}", "│   "),
+            Tree::Empty => write!(f, "{}", "    "),
+        }
+    }
+}
+
 /// Represent list of symbols as filetype=grep buffer content.
 /// Paths are converted into relative to project root.
 pub fn format_symbol<T: Symbol<T>>(
@@ -216,14 +235,41 @@ pub fn format_symbol<T: Symbol<T>>(
     ctx: &Context,
 ) -> String {
     fn format_symbol_at_depth<'a, T: Symbol<T>>(
-        output: &mut Vec<(String, String, &'a str)>,
+        output: &mut Vec<(String, String, String)>,
         items: &'a [T],
         meta: &EditorMeta,
         server: &ServerSettings,
         ctx: &Context,
         depth: usize,
+        prefix: &Vec<bool>,
     ) {
-        for symbol in items {
+        let length = items.len();
+        let is_root = depth == 0;
+        for (index, symbol) in items.iter().enumerate() {
+            let is_last = index + 1 == length;
+
+            let mut new_prefix = prefix.clone();
+            new_prefix.push(is_last);
+
+            let hierarchy = if is_root {
+                "".to_string()
+            } else {
+                let last_hierarchy_symbol = if is_last {
+                    Tree::LastChild
+                } else {
+                    Tree::Child
+                };
+                let prefixing_hierarchy_symbols = prefix
+                    .iter()
+                    .rev()
+                    .take(depth - 1)
+                    .map(|flag| if *flag { Tree::Empty } else { Tree::Pipe })
+                    .rev()
+                    .join("");
+                format!("{}{}", prefixing_hierarchy_symbols, last_hierarchy_symbol)
+            };
+            let description = format!("{}{}", hierarchy, symbol.name());
+            let kind = format!("{}{:?}", hierarchy, symbol.kind());
             let mut filename_path = PathBuf::default();
             let filename = symbol_filename(meta, symbol, &mut filename_path);
             let position = get_kakoune_position_with_fallback(
@@ -240,14 +286,15 @@ pub fn format_symbol<T: Symbol<T>>(
                     position.line,
                     position.column,
                 ),
-                format!("{:?}", symbol.kind()),
-                symbol.name(),
+                description,
+                kind,
             ));
-            format_symbol_at_depth(output, symbol.children(), meta, server, ctx, depth + 1)
+            let children = symbol.children();
+            format_symbol_at_depth(output, children, meta, server, ctx, depth + 1, &new_prefix)
         }
     }
     let mut columns = vec![];
-    format_symbol_at_depth(&mut columns, &items, meta, server, ctx, 0);
+    format_symbol_at_depth(&mut columns, &items, meta, server, ctx, 0, &vec![]);
     if align {
         let Some(width1) = columns
             .iter()
@@ -258,23 +305,29 @@ pub fn format_symbol<T: Symbol<T>>(
         };
         let width2 = columns
             .iter()
-            .map(|(_, kind, _)| UnicodeWidthStr::width(kind.as_str()))
+            .map(|(_, description, _)| UnicodeWidthStr::width(description.as_str()))
+            .max()
+            .unwrap();
+        let width3 = columns
+            .iter()
+            .map(|(_, _, kind)| UnicodeWidthStr::width(kind.as_str()))
             .max()
             .unwrap();
         columns
             .into_iter()
-            .map(|(position, kind, description)| {
+            .map(|(position, description, kind)| {
                 format!(
-                    "{position:width1$} {kind:width2$} {description}\n",
+                    "{position:width1$} {description:width2$} {kind:width3$}\n",
                     width1 = width1,
-                    width2 = width2
+                    width2 = width2,
+                    width3 = width3,
                 )
             })
             .join("")
     } else {
         columns
             .into_iter()
-            .map(|(position, kind, description)| format!("{position} {kind} {description}\n"))
+            .map(|(position, description, kind)| format!("{position} {description} {kind}\n"))
             .join("")
     }
 }
--- a/src/language_features/document_symbol.rs
+++ b/src/language_features/document_symbol.rs
@@ -235,16 +235,15 @@ pub fn format_symbol<T: Symbol<T>>(
     ctx: &Context,
 ) -> String {
     fn format_symbol_at_depth<'a, T: Symbol<T>>(
-        output: &mut Vec<(String, String, String)>,
+        output: &mut Vec<(String, String)>,
         items: &'a [T],
         meta: &EditorMeta,
         server: &ServerSettings,
         ctx: &Context,
-        depth: usize,
         prefix: &Vec<bool>,
     ) {
         let length = items.len();
-        let is_root = depth == 0;
+        let is_root = prefix.len() == 0;
         for (index, symbol) in items.iter().enumerate() {
             let is_last = index + 1 == length;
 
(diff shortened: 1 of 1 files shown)


src/language_features/document_symbol.rs
```diff
@@ -206,6 +207,24 @@ fn editor_document_symbol(
...
 }
 
+enum Tree {
+    Child,
+    LastChild,
+    Pipe,
+    Empty,
+}
+
+impl fmt::Display for Tree {
```
krobelus, line 217: > 1. Perhaps, previous style was there for a reason, e.g. to make it easy to parse the buffer in case if users would want to customize it using Kakoune itself.

Not specifically.

> And probably, it already could be implemented with highlighters. I'm still learning Kakoune, so not 100% sure about this.

Highlighters can work but are not the best choice (you can't navigate or search for the virtual text as created by replace-ranges).
Users can indeed override `lsp-show-document-symbol` but barely anyone does.

> 3. Use of UTF-8 symbols instead of ASCII.

That's fine because Kakoune already uses box drawing characters unconditionally for `:info`.

---

The reason why symbol kinds are in a separate column is because their length is effectively bounded (the longest one is "TypeParameter") so it is cheap -- in terms of screen real estate -- to align them.
I think with the tree, the alignment is no longer important enough, so I'd add the kind directly after the symbol name (or if we have universally-understood icons we can also put them before).

Also the new tree clearly outshines the indented approach in the first column. I don't think that one adds value anymore, especially because it's all the same file.
The correspondence between nesting structure and line number makes no sense. Only between nesting structure and symbol, or symbol and line number.

So how about

```
links-fixer.scala:17:12: Link (Class)
links-fixer.scala:24:8:  LinksFixer (Module)
links-fixer.scala:26:7:  ├── main (Method)
links-fixer.scala:29:15: ├── fixFile (Method)
links-fixer.scala:30:9:  │   └── file (Constant)
links-fixer.scala:34:15: ├── fix (Method)
links-fixer.scala:37:15: ├── parse (Method)
links-fixer.scala:38:9:  │   ├── trimmed (Constant)
links-fixer.scala:39:15: │   ├── prefix (Constant)
links-fixer.scala:39:23: │   ├── title (Constant)
links-fixer.scala:40:9:  │   ├── fragment (Constant)
links-fixer.scala:44:9:  │   ├── fragmentless (Constant)
links-fixer.scala:45:10: │   ├── path (Constant)
links-fixer.scala:45:16: │   └── file (Constant)
links-fixer.scala:51:15: └── render (Method)
```

WDYT?

(the file path is still noisy but it's a reasonable compromise to make this look and feel like a grep-buffer)

src/language_features/document_symbol.rs
```diff
@@ -216,14 +235,41 @@
...
 ) -> String {
     fn format_symbol_at_depth<'a, T: Symbol<T>>(
-        output: &mut Vec<(String, String, &'a str)>,
+        output: &mut Vec<(String, String, String)>,
         items: &'a [T],
         meta: &EditorMeta,
         server: &ServerSettings,
         ctx: &Context,
         depth: usize,
+        prefix: &Vec<bool>,
```
krobelus, line 244: This should mean that `depth` is redundant (it's `prefix.len()`) so we should remove it

src/language_features/document_symbol.rs
```diff
@@ -216,14 +235,41 @@
...
     ) {
-        for symbol in items {
+        let length = items.len();
+        let is_root = depth == 0;
+        for (index, symbol) in items.iter().enumerate() {
+            let is_last = index + 1 == length;
+
+            let mut new_prefix = prefix.clone();
+            new_prefix.push(is_last);
```
krobelus, line 252: Should not need to clone the entire Vec here, instead push to the existing one and pop it at the end of the loop iteration

src/language_features/document_symbol.rs
```diff
@@ -216,14 +235,41 @@
...
+                    Tree::LastChild
+                } else {
+                    Tree::Child
+                };
+                let prefixing_hierarchy_symbols = prefix
+                    .iter()
+                    .rev()
+                    .take(depth - 1)
+                    .map(|flag| if *flag { Tree::Empty } else { Tree::Pipe })
+                    .rev()
```
krobelus, line 267: `.rev().take().rev()` can be replaced by `skip(1)`.
It makes sense because we don't want to indent top-level symbols. Could also try to not push if root but that'd probably be uglier.

src/language_features/document_symbol.rs
```diff
@@ -258,23 +305,29 @@
         };
         let width2 = columns
             .iter()
-            .map(|(_, kind, _)| UnicodeWidthStr::width(kind.as_str()))
+            .map(|(_, description, _)| UnicodeWidthStr::width(description.as_str()))
+            .max()
+            .unwrap();
+        let width3 = columns
```
krobelus, line 311: I don't think the last column needs alignment because it's left-justified (and we don't want to add trailing spaces, right?).
With my suggestion, we can drop the alignment logic altogether, because its only use is document symbols where the filename is always the same.
For `lsp-workspace-symbol` the alignment is disabled because it might cause confusion if some file paths are really long (not sure, I never use it)

igor-ramazanov: I think we have to keep it because positions and line numbers can have different length.

src/language_features/document_symbol.rs
igor-ramazanov: Done

src/language_features/document_symbol.rs
igor-ramazanov: Done

src/language_features/document_symbol.rs
igor-ramazanov: Agree, pushed a commit.

src/language_features/document_symbol.rs
krobelus: right, my bad. I was misled by the fact that we don't need UnicodeWidthStr for that but it doesn't hurt of course

src/language_features/document_symbol.rs
igor-ramazanov: Sorry, I didn't understand what you meant.

The algorithm build a prefix string of booleans as _"is this symbol at the depth level (marked by a vector index) is a last child of its parent?"_. For some symbols it will be `111`, for some `010`, etc. It's probably possible, but I'm not sure how can I reuse the same vector.

src/language_features/document_symbol.rs
igor-ramazanov: I think [Visual Studio Code icons](https://github.com/microsoft/vscode-icons/) are probably the most recognizable?

Although, they are in .svg format.

Another option would be to highlight the `SymbolKind` with the same color as in the editor, but syntax highlighting is out of the scope of the project. 

Or maybe just coloring them with the same built-in terminal color per-type?

src/language_features/document_symbol.rs
krobelus: we can reuse it because we're doing a [depth-first traversal](https://en.wikipedia.org/wiki/Depth-first_search).
The prefix vector is a stack of all ancestors of the current node, which is exactly what we need to compute the prefixed pipes.

igor-ramazanov: Yeah, [I see now](https://github.com/kak-lsp/kak-lsp/commit/4dd8e2a2a0222827a43808be9ed3c7483a6b1ffb#diff-aea8cfa1bdea7837f8efd39fe43fd6a66d98a9724b536736b10a97206d98478dR289-R291), cool!

src/language_features/document_symbol.rs
krobelus: > I think Visual Studio Code icons are probably the most recognizable?
> 
> Although, they are in .svg format.

Interesting, yeah.

I personally don't find them very intuitive (a method is a cube?) but it's
probably hard to do better with a small icon, and perhaps people who used
graphical IDEs recognize the pattern.

> 
> Another option would be to highlight the SymbolKind with the same color as in the editor, but syntax highlighting is out of the scope of the project.
> 
> Or maybe just coloring them with the same built-in terminal color per-type?

If the language server supports semantic tokens, then we can definitely apply the mapped face to the symbol name.
Our default configuration includes

```toml
[semantic_tokens]
faces = [
    {face="function", token="function"},
    {face="module", token="namespace"},
    {face="type", token="type"},
]
```

which means that given

```rust
mod some_module {
	fn some_function() {
	}
	struct SomeType {}
}
```

`some_module` has LSP semantic token kind "namespace", so we would add a highlighter that gives it the "module" Kakoune face.
The Kakoune theme sets the face. These are the available faces

```kak
# For Code
face global value red
face global type yellow
face global variable green
face global module green
face global function cyan
face global string magenta
face global keyword blue
face global operator yellow
face global attribute green
face global comment cyan
face global documentation comment
face global meta magenta
face global builtin default+b
```

For every SymbolKind we'd have to find a Semantic Token type and a fitting face.
This sounds like something we can do.
Not sure whether to keep the `(Function)` annotations, I guess so (I usually don't use syntax highlighting).

---

Instead of

	src/text_edit.rs:16:11:  TextEditish (Interface)
	src/text_edit.rs:17:8:   ├── text_edit (Function)
	src/text_edit.rs:18:8:   ├── as_ref (Function)
	src/text_edit.rs:19:8:   └── from_text_edit (Function)
	src/text_edit.rs:492:8:  KakouneTextEdit (Struct)
	src/text_edit.rs:493:5:  ├── range (Field)
	src/text_edit.rs:494:5:  ├── new_text (Field)
	src/text_edit.rs:495:5:  └── command (Field)

for Rust, I'd expect:

	src/text_edit.rs:16:11:  trait TextEditish
	src/text_edit.rs:17:8:   ├── fn text_edit
	src/text_edit.rs:18:8:   ├── fn as_ref
	src/text_edit.rs:19:8:   └── fn from_text_edit
	src/text_edit.rs:492:8:  struct KakouneTextEdit
	src/text_edit.rs:493:5:  ├── range: Range
	src/text_edit.rs:494:5:  ├── new_text: String
	src/text_edit.rs:495:5:  └── command: KakouneTextEditCommand

but that'd require language-specific info, it would be better if the LSP server did that

igor-ramazanov: > Faces

Yes, that is a good idea. It will cause integration problems with non-standard highlighters like https://github.com/phaazon/kak-tree-sitter, but such issues are better to be handled on their side, I think.

> Language-dependent outline

That would be a really good UX/UI improvement! To show the basic signature:
`pub fn as_ref`, `private val a: String`, `override def toString: String`.

Unfortunately, it seems like an LSP limitation.

I came to Kakoune after Intellij IDEA and that's how it looks there:
image


VS Code:
image


So, VS Code shows the same amount of information as we do, as it also uses LSP.

IntelliJ IDEA doesn't use LSP and instead has its own internal way of integrating with languages.

PR #711, opened 2024-01-22
Indentation guides

igor-ramazanov:
Adds indentation guides highlighting the scope of the functions/methods/classes/etc.

The implementation depends on the document symbols LSP API.

* Adds 2 new commands similar to existing commands.
    - `lsp-indent-guides-(enable|disable) {buffer|global|window}`.
* Adds 2 new faces.
    - `IndentGuide` and `IndentGuideFocused`, defaults `comment` and `MatchingChar`.
* Nested characters can be customized according to their level.
Picks the last one if the scope is deeper than the amount of characters.
    - `lsp_indent_guides_characters`, defaults to `│ ┆ ┊ ⸽`.
* Adds an option to skip specified levels of nestedness.
    - `lsp_indent_guides_skip`, defaults to `0`.
* Adds an option to disable highlighting of the current scope.
    - `lsp_indent_guides_highlight_focused`, defaults to `true`.

**NOTE**:
Unfortunately, it can not render on black lines due to
Kakoune `replace-range` being unable to render on blank spaces.
It's possible to make it render on blank lines by keeping track and merging specs on the same line,
however it's not trivial to implement.

I would really like to render on blank lines as it improves the UX quite well.
I was able to implement tracking and merging, but performance-wise it was pretty horrible,
so I decided to publish a simpler solution with a potential for further improvements.

And also likely will break if indentation does not follow the scope,
e.g. with multiline strings in some languages.

Reviews and advices are welcomed.

Screenshot before:
image

Screenshot after:
image

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -33,6 +33,9 @@ set-face global ReferenceBind +u@Reference
 # Face for inlay hints.
 set-face global InlayHint cyan+d
 set-face global InlayCodeLens cyan+d
+# Face for indent guides
+set-face global IndentGuide +f@comment
+set-face global IndentGuideFocused +f@MatchingChar
 
 # Options for tuning LSP behaviour.
 
@@ -161,6 +164,9 @@ Capture groups must be:
     3: optional column
     4: optional message
 } regex lsp_location_format ^\h*\K([^:\n]+):(\d+)\b(?::(\d+)\b)?(?::([^\n]+))
+
+declare-option -docstring "Number of indent levels to skip" int lsp_indent_guides_skip 1
+declare-option -docstring "Indent guides characters" str-list lsp_indent_guides_characters "│" "┆" "┊" "⸽"
 
 # Callback functions. May override these.
 
@@ -340,6 +346,7 @@ declare-option -hidden range-specs cquery_semhl
 declare-option -hidden int lsp_timestamp -1
 declare-option -hidden range-specs lsp_references
 declare-option -hidden range-specs lsp_semantic_tokens
+declare-option -hidden range-specs lsp_indent_guides
 declare-option -hidden range-specs lsp_inlay_hints
 declare-option -hidden range-specs lsp_inlay_code_lenses
 declare-option -hidden line-specs lsp_code_lenses 0 '0| '
@@ -1415,6 +1422,37 @@ incomingOrOutgoing = $1
 " | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }
 
+define-command -hidden lsp-indent-guides-request -docstring "request updating lsp_indent_guides for the buffer" %{
+  nop %sh{
+    eval set -- "$kak_quoted_opt_lsp_indent_guides_characters"
+    # also remove the trailing comma
+    indent_guides_characters="[ $(printf "'%s', " "$@" | rev | cut -c2- | rev) ]"
+
+    (printf %s "
+session  = \"${kak_session}\"
+buffile  = \"${kak_buffile}\"
+filetype = \"${kak_opt_filetype}\"
+version  = ${kak_timestamp:-0}
+method   = \"kakoune/indent-guides\"
+$([ -z ${kak_hook_param+x} ] || echo hook = true)
+[params]
+skip = $kak_opt_lsp_indent_guides_skip
+characters = $indent_guides_characters
+position_line = $kak_cursor_line
+" | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
+}
+
+define-command lsp-indent-guides-enable -params 1 -docstring "lsp-indent-guides-enable <scope>: enable indent guides for <scope>" %{
+  add-highlighter -override "%arg{1}/lsp_indent_guides" replace-ranges lsp_indent_guides
+  hook -group lsp-indent-guides %arg{1} BufReload .* lsp-indent-guides-request
+  hook -group lsp-indent-guides %arg{1} NormalIdle .* lsp-indent-guides-request
+  hook -group lsp-indent-guides %arg{1} InsertIdle .* lsp-indent-guides-request
+} -shell-script-candidates %{ printf '%s\n' buffer global window }
+
+define-command lsp-indent-guides-disable -params 1 -docstring "lsp-indent-guides-disable <scope>: disable indent guides for <scope>" %{
+  remove-highlighter "%arg{1}/lsp_indent_guides"
+  remove-hooks %arg{1} lsp-indent-guides
+} -shell-script-candidates %{ printf '%s\n' buffer global window }
 
 define-command -hidden lsp-inlay-hints -docstring "lsp-inlay-hints: request inlay hints" %{
     declare-option -hidden int lsp_inlay_hints_timestamp -1
--- a/src/language_features/document_symbol.rs
+++ b/src/language_features/document_symbol.rs
@@ -3,8 +3,9 @@ use crate::language_features::goto::edit_at_range;
 use crate::language_features::hover::editor_hover;
 use crate::markup::escape_kakoune_markup;
 use crate::position::{
-    get_kakoune_position_with_fallback, get_kakoune_range_with_fallback, get_lsp_position,
-    kakoune_position_to_lsp, lsp_range_to_kakoune, parse_kakoune_range,
+    get_kakoune_position_with_fallback, get_kakoune_range, get_kakoune_range_with_fallback,
+    get_lsp_position, kakoune_position_to_lsp, lsp_position_to_kakoune, lsp_range_to_kakoune,
+    parse_kakoune_range,
 };
 use crate::types::*;
 use crate::util::*;
@@ -15,7 +16,7 @@ use lsp_types::request::*;
 use lsp_types::*;
 use serde::Deserialize;
 use std::any::TypeId;
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::convert::TryInto;
 use std::fmt;
 use std::fmt::Write;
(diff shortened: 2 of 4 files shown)


igor-ramazanov:
@krobelus 
I would really like to be able to render them on blank lines too, but it's been difficult to achieve.

igor-ramazanov:
I was able to implement tracking and merging, but performance-wise it was pretty horrible.

krobelus:
This is awesome! Maybe we can fix the empty-line problem in Kakoune - this is not the only use case where virtual text should be shown after the actual line ends (`lsp-inlay-diagnostics` is another one which doesn't work properly today because of that).

igor-ramazanov:
Yeah, it's quite limiting, would be wonderful to fix it there, it'll unblock lots of possibilities.

I've tried to create code folding (using document symbols again :sweat_smile:), but it can't render the buffer properly, unfortunately.

igor-ramazanov:
There's something I don't understand: sometimes, the feedback is instantaneous and indent guides are updated very quickly, but sometimes I need to do something like changing modes to trigger the change.

It also happens sometimes with different code/features. I don't know why it happens. Something around `NormalIdle` not triggering, perhaps.

rc/lsp.kak
```diff
@@ -162,6 +165,10 @@ Capture groups must be:
     4: optional message
 } regex lsp_location_format ^\h*\K([^:\n]+):(\d+)\b(?::(\d+)\b)?(?::([^\n]+))
 
+declare-option -docstring "Number of indent levels to skip" int lsp_indent_guides_skip 0
```
krobelus, line 168: The purpose of indent guides is so we can see more easily which lines belong together.
At level 0 we already have such a guide, namely the edge of the terminal window.
These two vertical lines don't look good here and I think that's a common pattern.
Also, in most languages it's very obvious if we are inside a function or not (usually we are).

So I think the default should be 1.

rc/lsp.kak
```diff
@@ -162,6 +165,10 @@
...
+declare-option -docstring "Highlight current indent scope" bool lsp_indent_guides_highlight_focused true
```
krobelus, line 169: Is `lsp_indent_guides_highlight_focused` really necessary? Can't we control that already by setting IndentGuideFocused to the same as IndentGuid

rc/lsp.kak
```diff
@@ -1415,6 +1423,49 @@
 " | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }
 
+define-command -hidden lsp-indent-guides-request -docstring "lsp-indent-guides-request: request updating lsp_indent_guides for the buffer" %{
```
krobelus, line 1426: as a convention, we only add the command to the docstring if there are any arguments

```suggestion
define-command -hidden lsp-indent-guides-request -docstring "request updating lsp_indent_guides for the buffer" %{
```

rc/lsp.kak
```diff
@@ -1415,6 +1423,49 @@
...
+  declare-option -hidden int lsp_indent_guides_timestamp -1
+  nop %sh{
+    if [ $kak_opt_lsp_indent_guides_timestamp -eq $kak_timestamp ]
+    then
+      exit
+    fi
+
+    indent_guides_characters="[ $(
+      DELIM=""
+      for CHARACTER in $(echo $kak_quoted_opt_lsp_indent_guides_characters)
```
krobelus, line 1436: > echo $kak_quoted_opt_lsp_indent_guides_characters

this kind of expansion has some pitfalls, for example it won't work on spaces.
It doesn't matter here usually but it's user input so better use this pattern:

	eval set -- "$kak_quoted_opt_lsp_indent_guides_characters"
	for CHARACTER # in "$@" (but that's implied)
	do
		...
	done

while at it, we can maybe simplify it (not sure if the trailing comma is okay, didn't test it)

	eval set -- "$kak_quoted_opt_lsp_indent_guides_characters"
	printf %s, "$@"

rc/lsp.kak
```diff
@@ -1415,6 +1423,49 @@
...
+$([ -z ${kak_hook_param+x} ] || echo hook = true)
+[params]
+skip = $kak_opt_lsp_indent_guides_skip
+highlight_focused = $kak_opt_lsp_indent_guides_highlight_focused
+characters = $indent_guides_characters
+position_line = $kak_cursor_line
+" | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
+}
+
+define-command lsp-indent-guides-enable -params 1 -docstring "lsp-indent-guides-enable <scope>: enable indent guides for <scope>" %{
```
krobelus, line 1458: I wonder if we should make it the default.
Either way, maybe add a line in the Usage section of our README.

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,128 @@ use std::path::{Path, PathBuf};
...
+                         set-option buffer lsp_indent_guides_timestamp {version}"
+                    );
+                    let command = format!(
+                        "evaluate-commands -buffer {} -- {}",
+                        editor_quote(&meta.buffile),
+                        editor_quote(&command)
+                    );
+                    ctx.exec(meta, command);
+                }
+                _ => {
```
krobelus, line 79: maybe better to explicitly ignore `DocumentSymbolResponse::Flat` here, so the compiler will tell us if there's a new enum variant

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,128 @@
...
+                .get(adjusted_level)
+                .or(params.characters.last())
+                .unwrap();
+
+            // We want to ignore putting indentation on signatures and comments,
+            // thus starting from `selection_range`.
+            for line in (symbol.selection_range.end.line + 2)..(range.end.line) {
+                let spec = format!(
+                    "{}.{}+0|{{{}}}{} ",
+                    line, range.start.column, face, indent_character
```
krobelus, line 139: `indent_character` should be escaped using `escape_tuple_element()`

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,128 @@
...
+                    range.start.line <= params.position_line
+                        && params.position_line <= range.end.line
+                })
+                .is_none();
+
+            let is_inside =
+                range.start.line <= params.position_line && params.position_line <= range.end.line;
+            let is_focused = params.highlight_focused && is_not_in_child && is_inside;
+            let face = if is_focused {
+                "IndentGuideFocused"
```
krobelus, line 122: The `is_focused` logic doesn't seem to work for me.
I tried opening `src/text_edit.rs` and scroll around; indent guides work fine but they all have `IndentGuide` face.

In general, can we simplify this tree-walking so every node is only visited once?
We should be able to do this with a depth-first post-order traversal.
We always look at the innermost child first; and once we have found the symbol that "contains" the cursor, we ignore that check for all parents (or even for all other nodes).

```
def walk(node, position) -> bool:
    descendant_contains_position = False
    for child in node.children:
	descendant_contains_position |= walk(child, position)
    is_focused = not descendant_contains_position and node.contains(position)
    face = "IndentGuideFocused" if is_focused else "IndentGuide"
    return descendant_contains_position or is_focused
```

krobelus:
> Maybe we can fix the empty-line problem in Kakoune

I started a thread on this at https://lists.sr.ht/~mawww/kakoune/%3CZbTHsBlDwwv-TYmN%40gmail.com%3E

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,128 @@ use std::path::{Path, PathBuf};
 use unicode_width::UnicodeWidthStr;
 use url::Url;
 
+pub fn indent_guides(meta: EditorMeta, editor_params: EditorParams, ctx: &mut Context) {
```
krobelus, line 26: This doesn't work for blocks that are not indented.
For example this

```c++
namespace foo {
namespace bar {
namespace baz {

void foo() {
  int a;
  int b;
}

} // namespace baz
} // namespace bar
} // namespace foo
```

renders as

```
 namespace foo {
 namespace bar {
 │namespace baz {
 ┆│
 ┆│void foo() {
 ┊┆│  int a;
 ┊┆│  int b;
 ┆│}
 ┆│
 │} // namespace baz
 } // namespace bar
 } // namespace foo 
```

It's not critical to fix this if we don't enable it by default,
but I think it will be fine if we simply avoid adding an indent guide to the same column twice.

rc/lsp.kak
igor-ramazanov: Agree, changed to 1

rc/lsp.kak
igor-ramazanov: Yeah we can, and I like it - seems more elegant and simpler. Done!

src/language_features/document_symbol.rs
igor-ramazanov: Yes, done!

igor-ramazanov:
> I started a thread on this at [lists.sr.ht/~mawww/kakoune/%3CZbTHsBlDwwv-TYmN%40gmail.com%3E](https://lists.sr.ht/~mawww/kakoune/%3CZbTHsBlDwwv-TYmN%40gmail.com%3E)

Agree, btw, there's a hacky way to render on blank lines if that line has at least the newline character - by first copying the line's content, then updating it and replacing the whole line using `replace-ranges`.

It's not very convenient, but seems doable, at least.

Why with the newline only - because `replace-range` doesn't work well if it changes the amount of visible lines in the `window`, as I understand.

krobelus:
> Agree, btw, there's a hacky way to render on blank lines if that line has at least the newline character - by first copying the line's content, then updating it and replacing the whole line using replace-ranges.

Yeah actually that would work. Inside Kakoune, every line has a trailing newline.
I think we can do this workaround. Mawww off-list suggested the same workaround but also that it should maybe be part of a built-in highlighter.
The glitches happen for example when deleting a bunch of lines; the indentation guides briefly appear all on one line, separated a single space each.
So either we fix `replace-ranges` or we need a new highlighter.
It's not a blocker but I probably wouldn't use it until we fix that.

(Also I'd prefer if you rewrote your commits instead of adding new ones see https://www.bitsnbites.eu/git-history-work-log-vs-recipe/)

igor-ramazanov:
> (Also I'd prefer if you rewrote your commits instead of adding new ones see [bitsnbites.eu/git-history-work-log-vs-recipe](https://www.bitsnbites.eu/git-history-work-log-vs-recipe/))

Yeah, no problem, I'll cleanup the history :+1:

krobelus:
I think it's a bug in replace-ranges, reported as https://lists.sr.ht/~mawww/kakoune/%3C20240128002359.1846988-2-mawww%40kakoune.org%3E#%3CZbjOpENyllZnPayq@gmail.com%3E

src/language_features/document_symbol.rs
igor-ramazanov: Could it be something else? I'll try to simplify it anyway, yes, but does it work for you if try switching modes while being inside the innermost scope?

Could it be something with `NormalIdle` hook?

rc/lsp.kak
igor-ramazanov: Done, thanks

src/language_features/document_symbol.rs
igor-ramazanov: Done

rc/lsp.kak
igor-ramazanov: Yeah, agree that it should be reflected in the README. Will do

src/language_features/document_symbol.rs
igor-ramazanov: Removed the timestamp logic, so it requests it all the time. Seems ok with the `--release` build. Wanted to have a smarter filtering logic like you've described and still can implement it, but decided to try the simpler solution first. Seems ok?

src/language_features/document_symbol.rs
igor-ramazanov: Fixed by keeping lines and columns as keys in `BTree`s.

igor-ramazanov:
@krobelus

1. Cleaned up the git history.
1. Render on blank lines too.
1. Ignore one-line indentations, e.g. for single-line functions.
1. Override existing indent character if it already exists on the same column.

What's left:
1. Update the README.
1. Optimize traversal algorithm.

igor-ramazanov:
UI changed slightly, due to using `line.column+0` and starting the indentation guide on the symbol start line (before it was below).

The reason is if I specify the length, even if it's proper, then rendering breaks and becomes ugly when the cursor moves inside the replaced whitespace characters.

And then, if I don't start on the same line as symbol declaration, then it visually breaks the code indentation and becomes ugly again, as with length 0 it shifts affected lines to the right.

Also, decided to ignore identation for single line symbols, again, because it looked ugly.

Here's the new look, you can see that it renders on empty lines:
image

krobelus:
> UI changed slightly, due to using line.column+0 and starting the indentation guide on the symbol start line (before it was below).
> 
> The reason is if I specify the length, even if it's proper, then rendering breaks and becomes ugly when the cursor moves inside the replaced whitespace characters.

Thanks for exploring this. So this will add indentation to the second line
even though it should be at the same level as the first line:

```
let newv: Vec<RopeSlice<'_>> = new.lines().collect::<Vec<_>>();
│    struct BuildEditScript<'a, T: TextEditish<T>> {
```

This is quite weird; it sounds like implementing the expected behavior is
too complicated with `replace-ranges`.  Let's not overcomplicate things. This
is what mawww wrote:

> But overall I wonder if indentation guides should be a built-in, either
> standalone or part of the whitespaces highlighter.

So I think we should add to Kakoune support for something like

	declare-option line-specs lsp_indent_guides 
	# highlight column 8 on two lines
	set-option lsp_indent_guides %val{timestamp} 3|8,\| 4|8,\|
	add-highlighter buffer/lsp_indent_guides indent-guides lsp_indent_guides

igor-ramazanov:
>it sounds like implementing the expected behavior is
too complicated with replace-ranges

Unfortunately, yes, I've explored several solutions, and none of them is ideal.

Agree that it should be best done in Kakoune, and not the identations themselves, but enhancing/fixing highlighters.

Time to learn some C/C++ :)

krobelus:
> Time to learn some C/C++ :)

If needed I can try implementing it as well.
But we should find a design that fits Kakoune and is easy to use.

> set-option lsp_indent_guides %val{timestamp} 3|8,\| 4|8,\| ...

What if instead of telling Kakoune every single coordinate,
we could get away with only listing the lines where indent guides should start?
Then they can go until the first non-whitespace character below. E.g.

```
 1 namespace { namespace {
 2 	class foo {
 3 		void bar() {
 4 			;
 5 		}
 6 		void baz()
 7 		{
 8 			;
 9 		}
10 		void baz()
11 		  {
12 		  	;
13 		  }
14	}
15 } }
```

```kakrc
declare-option line-specs lsp_indent_guides 
# highlight the class and its three methods
set-option lsp_indent_guides %val{timestamp} 2| 3| 6| 10| # last case breaks, maybe use 11|
add-highlighter buffer/lsp_indent_guides indent-guides lsp_indent_guides
```

I wonder how the approaches compare:
- Specifying every coordinate would avoid false positives.
- Specifying end coordinate in addition to start would improve things but would still have transient false positives.
- Specifying every coordinate means that, when a line is changed, its indentation guide highlight is invalidated. This may cause flickering when we don't refill the option instantly. OTOH if we only put a flag-lines on the start (and maybe end) line, then it will only be invalidated when those lines change. Then the entire range will be invalidated.

This is mostly guesswork; once we look into the highlighter implementation we'll know more.

---

Forexample, I'd like to try using it in my kakrc which has patterns like

```
	evaluate-commands -save-regs ontab| %{
		set-register | %{
			perl -we '{
				$state = "header";
				while (<STDIN>) {
					if (m{^diff\b}) {
						$state = "header";
						next;
					}
		...
```

The structure is enforced by a [simple formatter](https://git.sr.ht/~krobelus/dotfiles/tree/master/item/bin/kakfmt) that indents after lines that end with one of "{([" and unindents on lines starting with "})]".

With a small change I could make it emit the `lsp_indent_guides` values.
(not sure what's my point, I guess I'm making the case for a dedicated highlighter instead of `replace-ranges` which would require weird workarounds for empty lines etc.)

igor-ramazanov:
Closing, replaced by https://github.com/kakoune-lsp/kakoune-lsp/pull/720.

PR #715, opened 2024-02-03
Sticky Contexts

igor-ramazanov:
Adds sticky contexts feature.

Example:
image

* new face `StickyContexts`, defaults to `default+Fbuid`.
* new option `lsp_sticky_contexts_skip`, defaults to 0.
* new commands `lsp-sticky-contexts-(enable|disable)` similar to existing commands.

**UPDATE:**
* added additional option `lsp_sticky_contexts_max`, defaults to 4.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -33,6 +33,8 @@ set-face global ReferenceBind +u@Reference
 # Face for inlay hints.
 set-face global InlayHint cyan+d
 set-face global InlayCodeLens cyan+d
+# Face for sticky contexts
+set-face global StickyContexts +Fiud@Default
 
 # Options for tuning LSP behaviour.
 
@@ -161,6 +163,9 @@ Capture groups must be:
     3: optional column
     4: optional message
 } regex lsp_location_format ^\h*\K([^:\n]+):(\d+)\b(?::(\d+)\b)?(?::([^\n]+))
+
+declare-option -docstring "Number of sticky contexts to skip from the top" int lsp_sticky_contexts_skip 0
+declare-option -docstring "Max number of sticky contexts to show" int lsp_sticky_contexts_max 4
 
 # Callback functions. May override these.
 
@@ -340,6 +345,7 @@ declare-option -hidden range-specs cquery_semhl
 declare-option -hidden int lsp_timestamp -1
 declare-option -hidden range-specs lsp_references
 declare-option -hidden range-specs lsp_semantic_tokens
+declare-option -hidden range-specs lsp_sticky_contexts
 declare-option -hidden range-specs lsp_inlay_hints
 declare-option -hidden range-specs lsp_inlay_code_lenses
 declare-option -hidden line-specs lsp_code_lenses 0 '0| '
@@ -1415,6 +1421,37 @@ incomingOrOutgoing = $1
 " | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }
 
+define-command -hidden lsp-sticky-contexts-request -docstring "request updating lsp_sticky_contexts for the buffer" %{
+  nop %sh{
+    window_start_line="$(echo $kak_window_range | cut -d' ' -f1)"
+
+    (printf %s "
+session  = \"${kak_session}\"
+buffile  = \"${kak_buffile}\"
+filetype = \"${kak_opt_filetype}\"
+version  = ${kak_timestamp:-0}
+method   = \"kakoune/sticky-contexts\"
+$([ -z ${kak_hook_param+x} ] || echo hook = true)
+[params]
+skip = $kak_opt_lsp_sticky_contexts_skip
+max = $kak_opt_lsp_sticky_contexts_max
+position_line = $kak_cursor_line
+window_start_line = $window_start_line
+window_width = $kak_window_width
+" | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
+}
+
+define-command lsp-sticky-contexts-enable -params 1 -docstring "lsp-sticky-contexts-enable <scope>: enable sticky contexts for <scope>" %{
+  add-highlighter -override "%arg{1}/lsp_sticky_contexts" replace-ranges lsp_sticky_contexts
+  hook -group lsp-sticky-contexts %arg{1} BufReload .* lsp-sticky-contexts-request
+  hook -group lsp-sticky-contexts %arg{1} NormalIdle .* lsp-sticky-contexts-request
+  hook -group lsp-sticky-contexts %arg{1} InsertIdle .* lsp-sticky-contexts-request
+} -shell-script-candidates %{ printf '%s\n' buffer global window }
+
+define-command lsp-sticky-contexts-disable -params 1 -docstring "lsp-sticky-contexts-disable <scope>: disable sticky contexts for <scope>" %{
+  remove-highlighter "%arg{1}/lsp_sticky_contexts"
+  remove-hooks %arg{1} lsp-sticky-contexts
+} -shell-script-candidates %{ printf '%s\n' buffer global window }
 
 define-command -hidden lsp-inlay-hints -docstring "lsp-inlay-hints: request inlay hints" %{
     declare-option -hidden int lsp_inlay_hints_timestamp -1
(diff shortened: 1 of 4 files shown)


igor-ramazanov:
The README for this feature also should be updated.

raiguard:
I've been using this for a few days and it's pretty great! My only gripes are the lack of highlighting (impossible to fix?) and the fact that it's not synchronous with the viewport. Neither are show-stoppers though so I will continue to use this in my daily work. Thank you!

igor-ramazanov:
@raiguard Thanks for the feedback!

Agree it would be nice to have a syntax highlighting, and I am not really sure to achieve it, unfortunately, maybe @krobelus has some ideas.

> and the fact that it's not synchronous with the viewport

Could you elaborate? I didn't fully get it.

raiguard:
The ranges update on idle, so the context lines don't always stay exactly at the top of the viewport - they jump around, especially if you're moving very quickly.

krobelus:
I did not look at the code yet but tried it briefly.

Apparently this uses the full symbol range, so comments are included, like

```
/// Apply text edits to the file pointed by uri either by asking Kakoune to modify corresponding     
```

probably `selectionRange` gives better results. Or only show the function name? That would be more robust (multi-line functions).
 
> Agree it would be nice to have a syntax highlighting, and I am not really sure to achieve it

Highlighting information is not really exposed I don't think. (On a related note I'd probably not make it bold by default but Default or dim)
But maybe Kakoune could have a custom `always-on-top` highlighter to "transplant" a given line to the top of the viewport.
That would solve the glitches when scrolling mentioned above.

When pressing `<c-d>` repeatedly to scroll, the cursor will be on the first line, hiding the sticky context, which is unfortunate.
Not sure how to fix this even with a custom highlighter. Would be great to fix but not a blocker.

Nesting doesn't seem to work properly. If I move around in `apply_text_edits_to_file_impl`, it keeps displaying the parent `apply_text_edits_to_file`.
I remember it sometimes displayed two lines, which was a bit too much I think.

igor-ramazanov:
> Apparently this uses the full symbol range, so comments are included, like

Indeed, didn't notice that. Using `selection_range` now for that.

> Or only show the function name?

Yeah, that would work, but it's not always a function, sometimes it's a package/module/class/trait. Didn't try, but feels like keywords like `fn`, `class`, `package`, etc. are important to show. It'd would be difficult to do that for each language.

> On a related note I'd probably not make it bold by default but Default or dim

Removed the boldness and switched to `@Default`.

> But maybe Kakoune could have a custom always-on-top highlighter to "transplant" a given line to the top of the viewport.
That would solve the glitches when scrolling mentioned above.

Yeah, it's not something we can do in this PR, I think.

> Nesting doesn't seem to work properly. If I move around in apply_text_edits_to_file_impl, it keeps displaying the parent apply_text_edits_to_file.
I remember it sometimes displayed two lines, which was a bit too much I think.

Yes, actually, it was intentional - to stack contexts on each other. E.g. if cursor in the nested function of another function, and both signatures are outside of the window position, then those 2 signature lines will be rendered.

I've added another option called `lsp_sticky_contexts_max` to limit max amount of sticky contexts. Combined with `lsp_sticky_contexts_skip` it makes it flexible to choose what do you want to see.

krobelus:
> When pressing <c-d> repeatedly to scroll, the cursor will be on the first line, hiding the sticky context,

actually we can do `set scrolloff 1,0`. Or `"%opt{lsp_sticky_contexts_max},1" if sticky context takes more lines.

I think `lsp-sticky-context-enable` should set scrolloff to that if the top-scrollof one is still at the default 0.

Another glitch is that when scrolling the viewport with the mouse such
that the cursor goes off-screen, then sticky context will go stale. I don't
off-hand know a solution but it's not a big deal.

> it makes it flexible to choose what do you want to see.

I feel strongly that configuration options are not a substitute for figuring out a good default behavior.
It's fine to take them for now but in future I'd probably opt to kill your darlings and replace them with a more robust UI, see below.

> declare-option int lsp_sticky_contexts_max 4

My feeling is that 4 lines is way too much noise for the default behavior.
Even 2 lines I don't really like though I didn't try for long, maybe I'm wrong.

My use case for this feature is to quickly see which function I'm viewing.
One thing that helps a lot is if it's always in the same place.
Take the modeline as example; there is a static location I can glance at to find out how many selections I have etc.
So I'd prefer one line with minimal clutter, like:

	... → subnamespace → someclass → somefunction() → innerfunction()

which I believe is what VSCode does.

I'd like to try putting the sticky context in the modeline (which shows up in the terminal title). That would happen to work well with my setup. No color support but I don't care about that.

It's probably fine to merge these features in a less-than-perfect state but
I'd really like to avoid what happened with inlay hints, they had horrible
glitches for years until we got a very simple patch from upstream adding
`flag-lines -after` which fixed it.

krobelus:
> Didn't try, but feels like keywords like fn, class, package, etc. are important to show. It'd would be difficult to do that for each language.

in the projects I work on it's obvious from the naming convention whether something is a type or a function, and the remaining symbol kinds are probably insignificant. If needed we can add the LSP-provided kind like `Function` and wait for more standardization.

igor-ramazanov:
@krobelus if I'm not mistaken then you're referring [to this one](https://code.visualstudio.com/Docs/editor/editingevolved#_breadcrumbs)?

igor-ramazanov:
@krobelus Pushed a new commit to replace sticky contexts with breadcrumbs in the modeline. It's a quick prototype to see how it works.

What perhaps nice to have:
* Highlight the last element using markup, e.g. `{Information}`
* Render on the left side?

igor-ramazanov:
Linked to this: https://github.com/mawww/kakoune/issues/4507

krobelus:
There are two options on what two show:
a. the sticky line, possibly with some markup
b. breadcrumbs

and several options how to show it:

1. top line using replace-ranges
2. modeline (which shows up as terminal title)
3. top line using a new highlighter

I think breadcrumbs with modeline or new highlighter have potential to be enabled by default.
I propose we merge modeline breadcrumbs enabled by default and watch out for any problems[*].

Does that sound good to you?

Some comments on the breadcrumbs MVP:
- the modeline change should be window-scoped, not buffer scoped because you can have two clients showing a buffer with different cursor position.
- I prefer `>` to `->`, to save space

> * Render on the left side?

The problem is that that would be ambiguous with `echo`. Not sure if that's a problem in practice... maybe something to try for the future though I wouldn't bet on it.

[*]: If anyone actually needs it we can also add the sticky line using replace-ranges but with plans to replace it with a new highlighter to get rid of glitches.
The new highlighter would also be useful for breadcrumbs -- modeline breadcrumbs work best in envs like i3 where the terminal title is clearly visible; a custom in-buffer line would be better for tmux.

igor-ramazanov:
@krobelus Thanks for the feedback!

I would enable breadcrumbs in the modeline by default, yes. It's simple and useful.

Also I think that even in the current implementation sticky contexts also provide a value, but they aren't that stable, and indeed the best thing to do is have highlighters fixed to provide stable, non-glitchy and a pleasant experience.

I don't really know what is the best course of actions here, as I want both to merge it right away as they still seem to be useful in the current form, but also it would be even better if those de-synchronizations would go away. And even better, would be if that highlighter could "copy" the highlighting of the source lines as @raiguard mentioned - that would simply be a perfect solution.

It's a somewhat complex decision, but I would like to avoid the situation of merging them, and then they would be in a _good enough to ignore the problem and not to improve_ state.

I'm also concerned if the code isn't getting too bloated and repetitive where each feature is implemented "in itself" without regards with the rest of the codebase.

So, I don't have a strong opinion over sticky contexts.Would be really cool if we would have improved highlighters in Kakoune right away, but yeah, they need to be implemented first.

krobelus:
If we merge the sticky contexts, I suggest we don't document it and make it a hidden command; then it's fine to remove it without warning.

> I'm also concerned if the code isn't getting too bloated

Adding more requests doesn't really cost us anything.
Though it can be a problem if we add competing features and let one of them bitrot.

(Oh and we do have bloat - our debug binary is 177M (!) should look into why)

I'd say don't merge sticky contexts until someone has used them for a while and vouches for them.

igor-ramazanov:
@krobelus Yeah, let's keep sticky contexts and indent guides open.

I've extracted the breadcrumbs PR: https://github.com/kakoune-lsp/kakoune-lsp/pull/719

igor-ramazanov:
Closing in favour of https://github.com/kakoune-lsp/kakoune-lsp/pull/721.

PR #717, opened 2024-02-08
Add `lsp_hover_max_info_lines` and `lsp_hover_max_diagnostic_lines`

### LINKED ISSUE #716: `Diagnostics` message ignore `lsp_hover_max_lines`
For some reason `lsp_hover_max_lines` seems to effect all hover messages but the `Diagnostics` ones. And as a result I am getting `Diagnostics` messages the size of my kak window maybe a common issue for Haskell programmers.

Rosuavio:
Fixes #716


krobelus:
sorry about the broken CI, I'll push a fix in a minute

Rosuavio:
I forgot to add the bit about prioritizing diagnostics over info in the commit message, is it worth re-triggering CI?

krobelus:
it's fine to trigger CI but I'll have some comments on the changes, so you can also wait until the next push, and maybe put it in a comment here

Rosuavio:
So I just discovered that awk has functions, so on-top of addressing any feedback I can probably clean this up big time with some functions if is appropriate.

CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
 ## Unreleased
 
+Breaking changes:
+- `lsp_hover_max_lines` now limits the size of the hover box spawned by `lsp-hover`, limiting lines of diagnostics as well as lines of information.
+
 Additions:
 - `lsp-document-symbols` now renders symbols in a tree.
 - `lsp-code-lens` can now run the test at cursor with `rust-analyzer`, provided https://gitlab.com/Screwtapello/kakoune-cargo is installed.
+- `lsp-hover-max_info_lines` defaults to 20, which limit the information in the lsp hover output box spawned by `lsp-hover`.
```
krobelus, line 9: typo in the option name:

```suggestion
- `lsp_hover_max_info_lines` defaults to 20, which limit the information in the lsp hover output box spawned by `lsp-hover`.
```

rc/lsp.kak
```diff
@@ -77,6 +77,8 @@ declare-option -docstring "Automatically highlight references with Reference fac
 declare-option -docstring "Show available code actions (default: a 💡 in the modeline)" bool lsp_auto_show_code_actions false
 # Set it to a positive number to limit the size of the lsp-hover output. Use 0 to disable the limit.
 declare-option -docstring "Set it to a positive number to limit the size of the lsp hover output. Use 0 to disable the limit" int lsp_hover_max_lines 20
+declare-option -docstring "Set it to a positive number to limit the information in the lsp hover output. Use 0 to disable the limit" int lsp_hover_max_info_lines 20
+declare-option -docstring "Set it to a positive number to limit the diagnostics in the lsp hover output. Use 0 to disable the limit" int lsp_hover_max_diagnostic_lines 20
```
krobelus, line 81: So the fix to `lsp_hover_max_lines` is already a good improvement worth taking.

I noticed that if we have a ton of diagnostics, this will show 20 lines of diagnostics and no line of hover info.
I think it would be better to show a bit of both, maybe 10 lines each.

I'd expect the behavior to be: if truncation is happening, include all
diagnostics up to a maximum of say 10 lines, and fill up what's left of the
20 line budget with hover info. After that, if there's still something left
in the budget, fill it up with the remaining diagnostics.

We can almost get there with your new options by setting
`lsp_hover_max_diagnostic_lines` to 10; however, if there is no hover info
then we do want to show up to 20 lines worth of diagnostics.

I don't think this edge case is very important.  But I do think making it
work should be the goal because that would mean we might get away without
the two new config options (or at least, most users wouldn't ever need to
use it to fix this case). As an intermediate solution we could default
`lsp_hover_max_diagnostic_lines` to something like 10?

rc/lsp.kak
```diff
@@ -113,40 +115,180 @@
...
+                printf "\n"
+                printf diagnostic_data_line[1]
+                for (i = 2; i <= diagnostic_lines; i++) {
+                    printf "%s%s", "\n", diagnostic_data_line[i]
+                }
+            }
+
+            if (diagnostics_truncated == 1) {
+                printf "\n"
+                printf "{+i@InfoDefault}Hover diagnostics truncated, use lsp-hover-buffer (shortcut H) for full hover info"
```
krobelus, line 271: not sure if it's a good idea to show two messages about truncation.
I'd expect only one at the bottom. 
At the bottom, it's right next to the other hints about code actions etc. that show which shortcuts to use.

rc/lsp.kak
```diff
@@ -113,40 +115,180 @@
...
+                # "Diagnostics truncated" message, update the
+                # available_dynamic_lines and extra_dynamic_lines.
+                if (diagnostics_truncated == 0) {
+                    diagnostics_truncated = 1
+                    static_lines++
+
+                    available_dynamic_lines = available_lines > static_lines ?
+                        available_lines - static_lines : 0
+                    extra_dynamic_lines = requested_dynamic_lines > available_dynamic_lines ?
+                        requested_dynamic_lines - available_dynamic_lines : 0
```
krobelus, line 227: haven't read the code yet but it looks like we can extract a function here

Rosuavio:
Okay after some thinking this can go on of two ways.

1. Predictability, control and simplicity in the different max heights of the contents of the box with a loss of predictability and control (keeping only simplicity) in the max height of the box.
2. Predictability, control and simplicity in the "max height of the box" (like a `lsp_max_hover_lines` that is absolute), with a battle between predictability, control and simplicity in how the space in the box is used.

1 would be the simplest solution, but would require that we deprecate  `lsp_max_hover_lines` and only provide `lsp_max_info_lines` and `lsp_max_diagnostic_lines`. They directly control the max lines for those portions of the hover box and nothing more. It would be dead simple to use and implement and very predictable for the user, but it would not provide the user with direct control over what I believe they care the most about, which is that the box never exceeds a certain height.

2 would keep and "fix" `lsp_max_hover_lines` but how the contents of the box is split can either be 
- Simple and predictable with little control - If the space is split with a ratio or percentage that would be simple for the user to configure and underhand and very predictable in its results, but provide little control over how the space is split
- Precisely controllable, but complicated and less predictable  - If we provide the user with a way to specify a small list of "fill objectives" that we will try to meet in the order they specify then there control is only limited by the a lack of proper "fill objectives" for them to included in there configuration. For these fill objectives to meet many possible user needs they writing them might be like writing a small configuration language.

If I take your a bit from your feedback
> I think it would be better to show a bit of both, maybe 10 lines each.
> 
> I'd expect the behavior to be: if truncation is happening, include all diagnostics up to a maximum of say 10 lines, and fill up what's left of the 20 line budget with hover info. After that, if there's still something left in the budget, fill it up with the remaining diagnostics.
> 
> We can almost get there with your new options by setting `lsp_hover_max_diagnostic_lines` to 10; however, if there is no hover info then we do want to show up to 20 lines worth of diagnostics.

and translate it into a "fill objective" list I think it would look like...

1. 10 diagnostic
2. rest of the info
3. rest of the diagnostic

But from you saying "I think it would be better to show a bit of both, maybe 10 lines each." I think that something like this would be closer to what you are asking for...

1. 10-info:10-diagnostic (together)
2. rest of the info
3. rest of the diagnostic

The first "fill objective" would essentially alternate so that it scale better with different amount of `lsp_max_hover_lines`.

It would obviously be way easier enough to hard code one of those patters, and if it makes this PR can hardcore one of this patterns, but I can make a fallow up PR after that allows the user to configure their "fill objective" list. Some example values of different types of fill objectives I thought of are...

- Rest of info
- 10 diagnostic lines
- 10% max lines of info lines
- 10-info and 5-diagnostic lines together - starting with diagnostic
- rest of info and diagnostic together - starting with info
- rest of info and diagnostic together with a 3:4 ratio - starting with diagnostics
- 50% of max lines of info and diagnostics together with a ratio of 2:1 - starting with info
- 10% of max lines of info and 5 lines of diagnostics together - starting with info

krobelus:
> 1 would be the simplest solution, but would require that we
> deprecate lsp_max_hover_lines and only provide lsp_max_info_lines and
> lsp_max_diagnostic_lines

That sounds reasonable.

My primary concern was to make the default experience a little bit better.
If there is both a ton of hover info and diagnostics, neither should eclipse the other.

Other than that, I don't yet have a strong opinion on which approach to
take. If anything I prefer the simplest one.
(That will also make it easier for users to copy and tweak `lsp_show_hover_format` for max flexibility.  We can also provide an overridable callback that receives as arguments what we want to render and outputs the full box. Though I don't expect anyone would want to do that.)

> The first "fill objective" would essentially alternate so that it scale
> better with different amount of lsp_max_hover_lines.

Interesting, yeah so if there are 8 diagnostics and 8 info lines, then a
limit of 10 meaning we show 5 of both would make sense.. but we can avoid
this complexity if we use two options.

PR #718, opened 2024-02-16
fix: return the whole settings when section is empty

Yukaii:
When working with the `vscode-eslint-language-server`, it request `workspace/configuration` with a empty section, while kak-lsp respond with `null`.

It's caused by line the following line in `kak-lsp`:

https://github.com/kak-lsp/kak-lsp/blob/117db92b515ea8e826a9bfbda98432a7fd345c25/src/workspace.rs#L92

While the same config works for helix, helix just return the whole settings object when the section is empty.

### Working screenshot

スクリーンショット 2024-02-16 朝9 37 24

~~It works on my machine™~~

### Configuration for kak-lsp

```toml
[language_server.eslint]
filetypes = ["javascript", "typescript"]
roots = [".eslintrc", ".eslintrc.json"]
command = "vscode-eslint-language-server"
args = ["--stdio"]
settings_section = "_"

[language_server.eslint.settings._]
codeActionsOnSave = { mode = "all", "source.fixAll.eslint" = true }
format = { enable = true }
nodePath = ""
quiet = false
rulesCustomizations = []
run = "onType"
validate = "on"
experimental = {}
problems = { shortenToSingleLine = false }

[language_server.eslint.settings._.codeAction]
disableRuleComment = { enable = true, location = "separateLine" }
showDocumentation = { enable = false }

```

### Configuration for helix

```toml
[language-server.eslint]
command = "vscode-eslint-language-server"
args = ["--stdio"]

[language-server.eslint.config]
codeActionsOnSave = { mode = "all", "source.fixAll.eslint" = true }
format = { enable = true }
nodePath = ""
quiet = false
rulesCustomizations = []
run = "onType"
validate = "on"
experimental = {}
problems = { shortenToSingleLine = false }

[language-server.eslint.config.codeAction]
disableRuleComment = { enable = true, location = "separateLine" }
showDocumentation = { enable = false }
```

### DIFF
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -32,8 +32,7 @@ command = "clangd"
 filetypes = ["clojure"]
 roots = ["project.clj", ".git", ".hg"]
 command = "clojure-lsp"
-settings_section = "_"
-[language_server.clojure-lsp.settings._]
+[language_server.clojure-lsp.settings]
 # See https://clojure-lsp.io/settings/#all-settings
 # source-paths-ignore-regex = ["resources.*", "target.*"]
 
@@ -76,8 +75,7 @@ filetypes = ["scala"]
 roots = ["build.sbt", ".scala-build"]
 command = "metals"
 args = ["-Dmetals.extensions=false"]
-settings_section = "metals"
-[language_server.metals.settings.metals]
+[language_server.metals.settings]
 icons = "unicode"
 isHttpEnabled = true
 statusBarProvider = "log-message"
@@ -113,8 +111,7 @@ command = "jdtls"
 filetypes = ["elixir"]
 roots = ["mix.exs"]
 command = "elixir-ls"
-settings_section = "elixirLS"
-[language_server.elixir-ls.settings.elixirLS]
+[language_server.elixir-ls.settings]
 # See https://github.com/elixir-lsp/elixir-ls/blob/master/apps/language_server/lib/language_server/server.ex
 # dialyzerEnable = true
 
@@ -123,8 +120,7 @@ filetypes = ["elm"]
 roots = ["elm.json"]
 command = "elm-language-server"
 args = ["--stdio"]
-settings_section = "elmLS"
-[language_server.elm-language-server.settings.elmLS]
+[language_server.elm-language-server.settings]
 # See https://github.com/elm-tooling/elm-language-server#server-settings
 runtime = "node"
 elmPath = "elm"
@@ -148,7 +144,7 @@ command = "erlang_ls"
 filetypes = ["go"]
 roots = ["Gopkg.toml", "go.mod", ".git", ".hg"]
 command = "gopls"
-[language_server.gopls.settings.gopls]
+[language_server.gopls.settings]
 # See https://github.com/golang/tools/blob/master/gopls/doc/settings.md
 # "build.buildFlags" = []
 
@@ -157,8 +153,7 @@ filetypes = ["haskell"]
 roots = ["hie.yaml", "cabal.project", "Setup.hs", "stack.yaml", "*.cabal"]
 command = "haskell-language-server-wrapper"
 args = ["--lsp"]
-settings_section = "_"
-[language_server.haskell-language-server.settings._]
+[language_server.haskell-language-server.settings]
 # See https://haskell-language-server.readthedocs.io/en/latest/configuration.html
 # haskell.formattingProvider = "ormolu"
 
@@ -167,8 +162,7 @@ filetypes = ["html"]
 roots = ["package.json"]
 command = "vscode-html-languageserver"
 args = ["--stdio"]
-settings_section = "_"
-[language_server.html-language-server.settings._]
+[language_server.html-language-server.settings]
 # quotePreference = "single"
 # javascript.format.semicolons = "insert"
 
@@ -177,8 +171,7 @@ filetypes = ["php"]
 roots = [".htaccess", "composer.json"]
 command = "intelephense"
 args = ["--stdio"]
-settings_section = "intelephense"
-[language_server.intelephense.settings.intelephense]
+[language_server.intelephense.settings]
 storagePath = "/tmp/intelephense"
 
 [language_server.json-language-server]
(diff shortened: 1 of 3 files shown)


Yukaii:
Oh I found that I didn't use `settings_section` variable to get setting. I'll fix that first.

krobelus:
Ok so one solution is to add a `settings_section`.

The idea behind the `settings_section` is that it allows us to use a structure equivalent to VSCode's `settings.json` [*].

As far as I could tell, `textDocument/initialize` and `workspace/didChangeConfiguration` only send a server-specific section but `workspace/configuration` can request anything from `settings.json`.
For example, `gopls` might ask for settings owned by your Git plugin.

I didn't check but it looks like with the Helix approach, anything that is within reach of `workspace/configuration` will unconditionally be included in `textDocument/initialize` and `workspace/didChangeConfiguration`.
That should be the only functional difference as far as I can tell.
Of course this difference is usually harmless.

I think we should simplify our config in the common case and match the Helix model.
For servers (if any) that break if they get unknown keys in `textDocument/initialize` and `workspace/didChangeConfiguration`, we can still use `settings_section`.
It will be more obscure but my guess is that essentially no one will need it.

So if you or someone else wants to work on this, the change would entail
- in `kak-lsp.toml`, remove `settings_section`
- in `kak-lsp.toml`, change `[language_server.foo.settings.bar]` to just `[language_server.foo.settings]`
- in Rust, if `settings_section` is unset
  - send the entirety of `[language_server.foo.settings]`
  - return all settings when server requests the empty section (as this PR already does). With the above change this makes things more consistent instead of less.

[*]: It should be possible to write a script that converts `kak-lsp.toml` into a valid `settings.json` with the equivalent server configs. Converting from `settings.json` to `kak-lsp.toml` also works except that we need to get `settings_section` from somewhere else because that info is provided to VSCode by server-specific plugins, unfortunately.

Yukaii:
@krobelus Thanks for the detailed explanation and guidance for modifying the codebase! I'm unfamiliar with Rust and this PR is basically changed with the help of copilot haha.

I've pushed your suggested changes. Feel free to pick it up or make any modification to my commit!

krobelus:
Thanks. I did some testing: language servers like lua-language-server and texlab ask for specific sections via `workspace/configuration`.
This means that we need to have the section name in our config, like:

```toml
[language_server.texlab.settings]
texlab.forwardSearch.executable = "zathura"
```

but including the `texlab` prefix in either `initializationOptions` or `didChangeConfiguration` is garbage and some servers (like `gopls`) reply by showing an error message to the user. So that would be a bad regression.

I don't think the Helix approach works for configuring `texlab`. I'd be happy to hear otherwise. Similarly, I'm not sure how they configure `gopls`, probably they can't

Yukaii:
So would there be any chance to get `vscode-eslint-language-server` working? Maybe we keep the commit ca014ba and update `kak-lsp.toml` for example eslint config, other lsp config stay the same.

```toml
[language_server.eslint]
filetypes = ["javascript", "typescript"]
roots = [".eslintrc", ".eslintrc.json"]
command = "vscode-eslint-language-server"
args = ["--stdio"]

[language_server.eslint.settings]
codeActionsOnSave = { mode = "all", "source.fixAll.eslint" = true }
format = { enable = true }
nodePath = ""
quiet = false
rulesCustomizations = []
run = "onType"
validate = "on"
experimental = {}
problems = { shortenToSingleLine = false }

[language_server.eslint.settings.codeAction]
disableRuleComment = { enable = true, location = "separateLine" }
showDocumentation = { enable = false }
```

krobelus:
sorry I'll look into it. I need to find out what VSCode does here, it would be absurd if it sent all of `settings.json`

krobelus:
@Yukaii do you have a minimal where eslint server shows diagnostics? I tried creating `test.js` with this contents: `const a = {};`
But it doesn't show diagnostics, even with https://github.com/kakoune-lsp/kakoune-lsp/commit/ca014bac2d910982b2e20bfc15bfdc43a5031595

krobelus:
Oh my debug buffer has

```
LSP: log from eslint: Uncaught exception received.
LSP: log from eslint: TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at Object.isAbsolute (node:path:1160:5)
    at /usr/lib/node_modules/vscode-langservers-extracted/lib/eslint-language-server/eslint.js:1191:19
```

krobelus:
ok that goes away if I remove the settings_section.. but diagnostics still don't show. Apparently it crashes when the client doesn't support their LSP extensions (?!):

```
Feb 17 11:05:26.915 DEBG From server eslint: {"jsonrpc":"2.0","id":4,"method":"eslint/noLibrary","params":{"source":{"uri":"file:///home/johannes/git/kakoune-lsp/t/test.js"}}}, module: kak_lsp::language_server_transport:168
Feb 17 11:05:26.915 WARN Unsupported method: eslint/noLibrary, module: kak_lsp::controller:760
Feb 17 11:05:26.916 DEBG To server eslint: {"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found"},"id":4}, module: kak_lsp::language_server_transport:198
Feb 17 11:05:26.916 DEBG From server eslint: {"jsonrpc":"2.0","method":"window/logMessage","params":{"type":1,"message":"Uncaught exception received."}}, module: kak_lsp::language_server_transport:168
Feb 17 11:05:26.916 DEBG From server eslint: {"jsonrpc":"2.0","method":"window/logMessage","params":{"type":1,"message":"Error: Method not found\n    at handleResponse (/usr/lib/node_modules/vscode-langservers-extracted/node_modules/vscode-jsonrpc/lib/common/connection.js:565:48)\n    at handleMessage (/usr/lib/node_modules/vscode-langservers-extracted/node_modules/vscode-jsonrpc/lib/common/connection.js:345:13)\n    at processMessageQueue (/usr/lib/node_modules/vscode-langservers-extracted/node_modules/vscode-jsonrpc/lib/common/connection.js:362:17)\n    at Immediate.<anonymous> (/usr/lib/node_modules/vscode-langservers-extracted/node_modules/vscode-jsonrpc/lib/common/connection.js:334:13)\n    at process.processImmediate (node:internal/timers:478:21)"}}, module: kak_lsp::language_server_transport:168
```

we render it as

```
LSP: log from eslint: ESLint server running in node v21.6.1
LSP: log from eslint: Uncaught exception received.
LSP: log from eslint: Error: Method not found
    at handleResponse (/usr/lib/node_modules/vscode-langservers-extracted/node_modules/vscode-jsonrpc/lib/common/connection.js:565:48)
    at handleMessage (/usr/lib/node_modules/vscode-langservers-extracted/node_modules/vscode-jsonrpc/lib/common/connection.js:345:13)
    at processMessageQueue (/usr/lib/node_modules/vscode-langservers-extracted/node_modules/vscode-jsonrpc/lib/common/connection.js:362:17)
    at Immediate.<anonymous> (/usr/lib/node_modules/vscode-langservers-extracted/node_modules/vscode-jsonrpc/lib/common/connection.js:334:13)
    at process.processImmediate (node:internal/timers:478:21)
```

krobelus:
I don't get diagnostics in Helix or VSCode either. I don't know which file extension to use

krobelus:
nevermind I didn't have eslint installed.. anything else I migh need?

krobelus:
> Maybe we keep the commit https://github.com/kakoune-lsp/kakoune-lsp/commit/ca014bac2d910982b2e20bfc15bfdc43a5031595 and update kak-lsp.toml for example eslint config, other lsp config stay the same.

I have done some digging. The eslint language server does something really weird here.
I don't want to add a server-specific workaround that's enabled unconditionally. If upstream does not fix it we could add a special option to `kak-lsp.toml` to enable a workaround.

Upstream bug report is at https://github.com/microsoft/vscode-eslint/issues/1788

Yukaii:
Huge appreciation for your taking the time to look into this!

krobelus:
sure, I hope it works for you with the `workaround_eslint` config option. As I reported upstream, the configuration format that VSCode sends to the server is subtly different from what the `settings.json` looks like so you might need some tweaking.

Of course it would be best if instead of populating `language_server.eslint.settings` you could write `.eslintrc` or similar, because that should then work reliably, in any editor or CLI. Though I don't know if the eslint LS supports this.

PR #719, opened 2024-02-17
Breadcrumbs in the modeline

igor-ramazanov:
Extracted from https://github.com/kakoune-lsp/kakoune-lsp/pull/715

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -165,6 +165,7 @@ Capture groups must be:
 # Callback functions. May override these.
 
 declare-option -hidden str lsp_code_action_indicator
+
 define-command -hidden lsp-show-code-actions -params 1.. -docstring "Called when code actions are available for the main cursor position" %{
     set-option buffer lsp_modeline_code_actions %opt{lsp_code_action_indicator}
 }
@@ -345,10 +346,11 @@ declare-option -hidden range-specs lsp_inlay_code_lenses
 declare-option -hidden line-specs lsp_code_lenses 0 '0| '
 declare-option -hidden str lsp_project_root
 
+declare-option -hidden str lsp_modeline_breadcrumbs ""
 declare-option -hidden str lsp_modeline_code_actions
 declare-option -hidden str lsp_modeline_progress ""
 declare-option -hidden str lsp_modeline_message_requests ""
-declare-option -hidden str lsp_modeline '%opt{lsp_modeline_code_actions}%opt{lsp_modeline_progress}%opt{lsp_modeline_message_requests}'
+declare-option -hidden str lsp_modeline '%opt{lsp_modeline_breadcrumbs}%opt{lsp_modeline_code_actions}%opt{lsp_modeline_progress}%opt{lsp_modeline_message_requests}'
 set-option global modelinefmt "%opt{lsp_modeline} %opt{modelinefmt}"
 
 ### Requests ###
@@ -1415,6 +1417,20 @@ incomingOrOutgoing = $1
 " | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }
 
+define-command -hidden lsp-breadcrumbs-request -docstring "request updating modeline breadcrumbs for the window" %{
+  nop %sh{
+    (printf %s "
+session  = \"${kak_session}\"
+client   = \"${kak_client}\"
+buffile  = \"${kak_buffile}\"
+filetype = \"${kak_opt_filetype}\"
+version  = ${kak_timestamp:-0}
+method   = \"kakoune/breadcrumbs\"
+$([ -z ${kak_hook_param+x} ] || echo hook = true)
+[params]
+position_line = $kak_cursor_line
+" | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
+}
 
 define-command -hidden lsp-inlay-hints -docstring "lsp-inlay-hints: request inlay hints" %{
     declare-option -hidden int lsp_inlay_hints_timestamp -1
@@ -2327,6 +2343,9 @@ define-command -hidden lsp-enable-impl -params 1 %{
     hook -group lsp %arg{1} NormalKey (<a-i>|<a-a>|\[|\]|\{|\}|<a-\[>|<a-\]>|<a-\{>|<a-\}>) %{
         set-option window lsp_object_mode %val{hook_param}
     }
+    hook -group lsp-breadcrumbs %arg{1} BufReload .* lsp-breadcrumbs-request
+    hook -group lsp-breadcrumbs %arg{1} NormalIdle .* lsp-breadcrumbs-request
+    hook -group lsp-breadcrumbs %arg{1} InsertIdle .* lsp-breadcrumbs-request
     define-command -override lsp-on-exit lsp-exit
 }
 
@@ -2340,6 +2359,7 @@ define-command -hidden lsp-disable-impl -params 1 %{
     try %{ set-option -remove %arg{1} completers option=lsp_completions }
     set-option %arg{1} lsp_fail_if_disabled fail
     remove-hooks %arg{1} lsp
+    remove-hooks %arg{1} lsp-breadcrumbs
     remove-hooks global lsp-auto-hover
     remove-hooks global lsp-auto-hover-insert-mode
     remove-hooks global lsp-auto-signature-help
--- a/src/language_features/document_symbol.rs
+++ b/src/language_features/document_symbol.rs
@@ -3,8 +3,8 @@ use crate::language_features::goto::edit_at_range;
 use crate::language_features::hover::editor_hover;
 use crate::markup::escape_kakoune_markup;
 use crate::position::{
-    get_kakoune_position_with_fallback, get_kakoune_range_with_fallback, get_lsp_position,
-    kakoune_position_to_lsp, lsp_range_to_kakoune, parse_kakoune_range,
+    get_kakoune_position_with_fallback, get_kakoune_range, get_kakoune_range_with_fallback,
+    get_lsp_position, kakoune_position_to_lsp, lsp_range_to_kakoune, parse_kakoune_range,
 };
 use crate::types::*;
 use crate::util::*;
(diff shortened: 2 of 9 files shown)


rc/lsp.kak
```diff
@@ -1415,6 +1417,26 @@ incomingOrOutgoing = $1
 " | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
 }
 
+define-command -hidden lsp-breadcrumbs-request -docstring "request updating lsp_modeline_breadcrumbs for the buffer" %{
```
krobelus, line 1420: I've addressed all issues I found important enough and pushed over it (previous version was 9b36bb56ae1526fdaf0de922c7953cf398c94825).
If you have feedback, let me know.

Otherwise I'll merge later today.
I'm really happy about this feature, I've wanted it for years.
I think it makes a big difference in my workflow.
Sure do do have commands to jump to symbol start but I don't want to clobber my selectin/jump history just for finding the function name.

rc/lsp.kak
```diff
@@ -1415,6 +1417,26 @@
...
```
krobelus, line 1420: I'd avoid implementation details (variable name) in comments. Also it should be window scope

```suggestion
define-command -hidden lsp-breadcrumbs-request -docstring "request updating modeline breadcrumbs for the window" %{
```

rc/lsp.kak
```diff
@@ -1415,6 +1417,26 @@
...
+  nop %sh{
+    window_start_line="$(echo $kak_window_range | cut -d' ' -f1)"
```
krobelus, line 1422: unused, removing. It's not clear whether cursor or viewport is better, and cursor is definnitely simpler

rc/lsp.kak
```diff
@@ -1415,6 +1417,26 @@
...
+filetype = \"${kak_opt_filetype}\"
+version  = ${kak_timestamp:-0}
+method   = \"kakoune/breadcrumbs\"
+$([ -z ${kak_hook_param+x} ] || echo hook = true)
+[params]
+position_line = $kak_cursor_line
+" | eval "${kak_opt_lsp_cmd} --request") > /dev/null 2>&1 < /dev/null & }
+}
+
+hook -group lsp-breadcrumbs %arg{1} BufReload .* lsp-breadcrumbs-request
```
krobelus, line 1436: we don't want to enable these at load-time (i.e. `source`)
Only when `lsp-enable{,-window}` is called, we should launch servers etc.

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,125 @@ use std::path::{Path, PathBuf};
 use unicode_width::UnicodeWidthStr;
 use url::Url;
 
+pub fn breadcrumbs(meta: EditorMeta, editor_params: EditorParams, ctx: &mut Context) {
+    let eligible_servers: Vec<_> = ctx
+        .language_servers
+        .iter()
+        .filter(|srv| attempt_server_capability(*srv, &meta, CAPABILITY_DOCUMENT_SYMBOL))
```
krobelus, line 30: unrelated: one of these days we should go through all the requests that are not yet gated on a capability.
Otherwise we might get errors, especially from "mixin" servers that support only a handful of things like diagnostics

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,125 @@
...
+                }],
+            )
+        })
+        .collect();
+    let params = BreadcrumbsParams::deserialize(editor_params).unwrap();
+    ctx.call::<DocumentSymbolRequest, _>(
+        meta,
+        RequestParams::Each(req_params),
+        move |ctx: &mut Context, meta, results| {
+            match results.into_iter().find(|(_, v)| v.is_some()) {
```
krobelus, line 52: our usual style is to not have large callbacks but instead immediately sipatch to a function named `editor_breadcrumbs`.
It doesn't matter much so I'll not try that.

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,125 @@
...
+                Some((server_name, Some(DocumentSymbolResponse::Nested(symbols))))
+                    if !symbols.is_empty() =>
+                {
+                    let server = &ctx.language_servers[&server_name];
+                    let document = if let Some(document) = ctx.documents.get(&meta.buffile) {
+                        document
+                    } else {
+                        return;
```
krobelus, line 60: We can use `let-else` nowadays (`let Some(document) = ctx.documents.get(&meta.buffile) else { return }`).
Also this seems like a very exceptional case, I'll add a warning. No need to forward it to the user I guess since this is always called from a hook.

actually, clippy informed me that this is unused, so we can remove it altogether

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,125 @@
...
+                    };
+                    let mut filename_path = PathBuf::default();
+                    let symbol = &symbols[0];
```
krobelus, line 63: better inline this variable to make it abundantly clear it's only used in the next line

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,125 @@
...
+                        &symbols,
+                        &params,
+                        ctx,
+                        server,
+                        document,
+                        filename.as_str(),
+                        &mut breadcrumbs,
+                    );
+                    if let Some(last) = breadcrumbs.last_mut() {
+                        *last = format!("{last} ");
```
krobelus, line 76: I think it'd more clear to branch on `if breadcrumbs.is_empty()`

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,125 @@
...
+                    };
+                    let mut breadcrumbs = breadcrumbs.join(" > ");
+                    breadcrumbs = editor_quote(breadcrumbs.as_str());
```
krobelus, line 79: I think instead of using `mut` with reassignment it's better to declare a new var (can use the same name just fine)

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,125 @@
...
+                    let command =
+                        format!("set-option buffer lsp_modeline_breadcrumbs {breadcrumbs}");
```
krobelus, line 81: this should be window scope because Kakoune supports showing the same file in two different windows (unlike VSCode) and they can have different cursor positions
To implement this we need a client in context

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,125 @@
...
+            )
+        })
+        .collect();
+    let params = BreadcrumbsParams::deserialize(editor_params).unwrap();
+    ctx.call::<DocumentSymbolRequest, _>(
+        meta,
+        RequestParams::Each(req_params),
+        move |ctx: &mut Context, meta, results| {
+            match results.into_iter().find(|(_, v)| v.is_some()) {
+                Some((server_name, Some(DocumentSymbolResponse::Nested(symbols))))
```
krobelus, line 53: you match the `Some` pattern in several places.
Better use the early return pattern so this is dealt with in one place

src/language_features/document_symbol.rs
krobelus, line 79: I wonder if using `.as_str()` is better than what we usually do (which is add a `&`)

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,125 @@
...
+                        "evaluate-commands -buffer {} -- {}",
+                        editor_quote(&meta.buffile),
+                        editor_quote(&command)
+                    );
+                    ctx.exec(meta, command);
+                }
+                Some((_, Some(DocumentSymbolResponse::Nested(_)))) => {
+                    return;
+                }
+                Some((_, Some(DocumentSymbolResponse::Flat(_)))) => {
```
krobelus, line 92: It will be very easy to support Flat symbol responses.
We have a `trait Symbol<T: Symbol<T>>` that is an umbrella for both flat and nested symbols
that can absolutely be done post-merge but I did it already

src/language_features/document_symbol.rs
```diff
@@ -23,6 +23,125 @@
...
+        let symbol_range = get_kakoune_range(server, filename, &symbol.range, ctx).unwrap();
+        let is_inside = {
+            let is_after_symbol_start = symbol_range.start.line <= params.position_line;
+            let is_before_symbol_end = params.position_line <= symbol_range.end.line;
+            is_after_symbol_start && is_before_symbol_end
+        };
+        if is_inside {
+            match symbol.kind() {
+                SymbolKind::FUNCTION | SymbolKind::METHOD => {
+                    acc.push(symbol.name.clone() + "()");
```
krobelus, line 125: not sure if the `()` make sense everywhere -- it's not really kakoune-lsp's job to be creative about this kind of thing -- but I guess we can leave it and improve this later

krobelus, line 125: can get rid of this clone

src/language_features/document_symbol.rs
krobelus, line 76: This causes a double-space in the (depending on the language, common) case where there are no code actions.
We should probably get rid of it later, not sure

rc/lsp.kak
```diff
@@ -345,10 +346,11 @@
 declare-option -hidden line-specs lsp_code_lenses 0 '0| '
 declare-option -hidden str lsp_project_root
 
+declare-option -hidden str lsp_modeline_breadcrumbs ""
 declare-option -hidden str lsp_modeline_code_actions
 declare-option -hidden str lsp_modeline_progress ""
 declare-option -hidden str lsp_modeline_message_requests ""
-declare-option -hidden str lsp_modeline '%opt{lsp_modeline_code_actions}%opt{lsp_modeline_progress}%opt{lsp_modeline_message_requests}'
+declare-option -hidden str lsp_modeline '%opt{lsp_modeline_breadcrumbs}%opt{lsp_modeline_code_actions}%opt{lsp_modeline_progress}%opt{lsp_modeline_message_requests}'
```
krobelus, line 353: fixed the system tests (hope it's not flaky)

The system tests are woefully undocumented but they are fairly simple to use.

Each system test is an executable shell script.
When run by itself it does not check the output.
Since it doesn't need to capture any output it can take control of the TTY.
This means you can insert a `test_tmux attach` line to attach to a running test and debug interactively.

For running the test with checks use something like `sh -x test/run test/path-to-test.sh`.
Probably we can improve the interface.

rc/lsp.kak
igor-ramazanov: @krobelus Oh, sorry, I think I've force pushed as I didn't notice your messages. Can you force push over again?

krobelus:
Sorry I had to revert to the old version to be able to post my comments that's why I pushed thrice.
I think you pushed a stale version, see `git range-diff 80afe65...46c30a6 --creation-factor=100`

igor-ramazanov:
@krobelus I also did a force push before I've read your messages. Thanks for the changes! Helps me to learn Rust.

igor-ramazanov:
@krobelus Maybe it'd be easier to collaborate if I move branches from https://github.com/kakoune-lsp/kakoune-lsp/pull/711 and https://github.com/kakoune-lsp/kakoune-lsp/pull/715 to this repo? Although, not sure if the review history won't be lost.

Would nice perhaps to have them around as sort-of official kakoune-lsp WIP features.

src/language_features/document_symbol.rs
igor-ramazanov: Yes, in some languages it's possible to define parentheses-less methods which doesn't look consistent.

Parentheses are nice for languages where function or method indeed always defined with `foo()`.

I'm voting for removing this logic and to make consistent across all languages.

If we look at VS Code they assign a specific emoji/symbol to all `SymbolKind`s. Honestly, I don't see a connection, e.g. between a cube and `method`, but I guess it's a repetition and consistency which helps to quickly see where is what.

igor-ramazanov:
@krobelus Hm, maybe, not sure too. Is it because you want to save the space?

There are a number of possibilities:
1. Skip N top most symbols
2. Trim long/multiline symbols (`impl SomeTrait...`)
3. Silence one-line symbols
4. ...?

krobelus:
> Is it because you want to save the space?

No, it's because it might be distracting that the modeline updates when I use `j` to step through struct fields.
Normally only line/column numbers change and often it's the same amount of digits so I'd not notice unless I'm actively looking.
Will test some more

PR #720, opened 2024-02-17
WIP (Blocked): Indentation guides

igor-ramazanov:
Adds indentation guides highlighting the scope of the functions/methods/classes/etc.

The implementation depends on the document symbols LSP API.

* Adds 2 new commands similar to existing commands.
    - `lsp-indent-guides-(enable|disable) {buffer|global|window}`.
* Adds 2 new faces.
    - `IndentGuide` and `IndentGuideFocused`, defaults `comment` and `MatchingChar`.
* Nested characters can be customized according to their level.
Picks the last one if the scope is deeper than the amount of characters.
    - `lsp_indent_guides_characters`, defaults to `│ ┆ ┊ ⸽`.
* Adds an option to skip specified levels of nestedness.
    - `lsp_indent_guides_skip`, defaults to `1`.

Screenshot before:
image

Screenshot after:
image

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -33,6 +33,9 @@ set-face global ReferenceBind +u@Reference
 # Face for inlay hints.
 set-face global InlayHint cyan+d
 set-face global InlayCodeLens cyan+d
+# Face for indent guides
+set-face global IndentGuide +f@comment
+set-face global IndentGuideFocused +f@MatchingChar
 
 # Options for tuning LSP behaviour.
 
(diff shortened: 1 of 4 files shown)


igor-ramazanov:
@krobelus Answering your comment in a previous PR: https://github.com/kakoune-lsp/kakoune-lsp/pull/711#issuecomment-1937596204

I like the idea with specifying only start of the indent.

In my opinion, it's good to mirror the existing combination of `line-specs` option + `flag-lines` highlighter for columns: `column-specs` + `flag-columns`.

It'll be intuitive to understand and use through an analogy, I think.

Not exactly what you proposed, but a development of your idea.

Hopefully, this similarity can also help in implementation.

igor-ramazanov:
Ah, I've forgot that `flag-lines` is only for the gutter.

The only API we have to replace/render characters is `replace-ranges`.

PR #722, opened 2024-02-18
lsp-auto-hover: Only run lsp-hover if the selection changed

topisani:
This helps the auto hover info box not instantly kill other info boxes

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -2181,6 +2181,15 @@ define-command lsp-inlay-diagnostics-disable -params 1 -docstring "lsp-inlay-dia
     remove-hooks %arg{1} lsp-inlay-diagnostics
 } -shell-script-candidates %{ printf '%s\n' buffer global window }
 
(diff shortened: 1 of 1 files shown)


rc/lsp.kak
```diff
@@ -2181,14 +2181,23 @@ define-command lsp-inlay-diagnostics-disable -params 1 -docstring "lsp-inlay-dia
     remove-hooks %arg{1} lsp-inlay-diagnostics
 } -shell-script-candidates %{ printf '%s\n' buffer global window }
 
+declare-option -hidden range-specs lsp_auto_hover_selection
+define-command -hidden lsp-check-auto-hover -params 1 %{
+  eval %sh{
```
krobelus, line 2186: Nice, I think this change makes a lot of sense.
In theory it's possible that a user has different expectations but any friction is super tiny.

I have never really used auto hover because I used to work on fairly small terminals.
And also because of glitches like this one I guess.
I'll try using it again.

I'll fix `range-specs -> str` and make some small style changes.

PR #723, opened 2024-02-20
Hover box: better control over info and diagnostic lines

Rosuavio:
Simplifying the behavior of the controls for limiting the amount of output going into the hover box, while making them behave more as they closely to how the user expects them to behave.

- Change `lsp_hover_max_lines` to more directly control the info lines in the hover box.
- Deprecated `lsp_hover_max_lines` for `lsp_hover_max_info_lines`, hiding `lsp_hover_max_lines` defaulting it to `-1` to represent that it was not set by the user and using `lsp_hover_max_info_lines` to control the amount of info lines in the hover box only when `lsp_hover_max_lines` is `-1`.
- Added `lsp_hover_max_diagnostic_lines` to control the amount of diagnostic lines in the hover box.

Fix https://github.com/kakoune-lsp/kakoune-lsp/issues/716


Rosuavio:
Not sure if this is the best way to rename `lsp_hove_max_lines`to `lsp_hover_max_info_lines`. The sentinel value of `-1` feels like a hack but there seems to be no way to tell the difference between a user setting the default value of a option and an unset option.

rc/lsp.kak
```diff
@@ -113,40 +115,93 @@ info=$lsp_info \
...
+
+        info_lines = split(ENVIRON["info"], info_line, /\n/)
+
+        info_truncated = 0
+
+        # Will the info lines need to be truncated
+        if(is_truncated(max_info_lines, info_lines)) {
+            # Only output max_info_lines amount of info lines
+            info_lines = max_info_lines
+            info_truncated = 1
```
krobelus, line 134: > printf info_line[1]

should not use user input as format string, should be

	printf "%s" info_line[1]

I fixed it

rc/lsp.kak
```diff
@@ -113,40 +115,93 @@
...
+        # new lsp_hover_max_info_lines.
+        if (max_info_lines < 0)
+            max_info_lines = ENVIRON["kak_opt_lsp_hover_max_info_lines"]
+
+        info_lines = split(ENVIRON["info"], info_line, /\n/)
+
+        info_truncated = 0
+
+        # Will the info lines need to be truncated
+        if(is_truncated(max_info_lines, info_lines)) {
```
krobelus, line 131: I think something like `has_unterminated_line` would be more descriptive.
Anyway, can't we get rid of this variable and always newline-terminate every line as soon as we print it?

```diff
diff --git a/rc/lsp.kak b/rc/lsp.kak
index ddeb4db..c576738 100644
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -127,19 +127,14 @@ info=$lsp_info \
         info_truncated = 0
 
         # Will the info lines need to be truncated
-        if(is_truncated(max_info_lines, info_lines)) {
+        if (is_truncated(max_info_lines, info_lines)) {
             # Only output max_info_lines amount of info lines
             info_lines = max_info_lines
             info_truncated = 1
         }
 
-        # Track if there is anything in the hover, so we know when to insert
-        # a new-line character.
-        output_in_hover = 0
-
         if (info_lines) {
             print_at_least_one_line(info_line, info_lines)
-            output_in_hover = 1
         }
 
         diagnostics = ENVIRON["diagnostics"]
@@ -147,55 +142,37 @@ info=$lsp_info \
         diagnostics_truncated = 0
 
         if (diagnostics) {
-            if (output_in_hover == 1) printf "\n"
-            printf "{+b@InfoDefault}Diagnostics{InfoDefault} (shortcut e):"
+            print "{+b@InfoDefault}Diagnostics{InfoDefault} (shortcut e):"
 
             max_diagnostic_lines = ENVIRON["kak_opt_lsp_hover_max_diagnostic_lines"]
 
             diagnostic_lines = split(diagnostics, diagnostic_line, /\n/)
 
             # Will the diagnostic lines need to be truncated
-            if(is_truncated(max_diagnostic_lines, diagnostic_lines)) {
+            if (is_truncated(max_diagnostic_lines, diagnostic_lines)) {
                 # Only output max_diagnostic_lines amount of diagnostic lines
                 diagnostic_lines = max_diagnostic_lines
                 diagnostics_truncated = 1
             }
 
             if (diagnostic_lines) {
-                printf "\n"
                 print_at_least_one_line(diagnostic_line, diagnostic_lines)
             }
-
-            output_in_hover = 1
         }
 
-        if (info_truncated == 1 || diagnostics_truncated == 1) {
-            if (output_in_hover == 1) printf "\n"
-            printf "{+i@InfoDefault}Hover info truncated, use lsp-hover-buffer (shortcut H) for full hover info"
-            output_in_hover = 1
-        }
-        if (ENVIRON["code_lenses"]) {
-            if (output_in_hover == 1) printf "\n"
-            printf "Code Lenses available (shortcut l)"
-            output_in_hover = 1
-        }
-        if (ENVIRON["kak_opt_lsp_modeline_code_actions"]) {
-            if (output_in_hover == 1) printf "\n"
-            printf "Code Actions available (shortcut a)"
-            output_in_hover = 1
-        }
-
-        if (ENVIRON["kak_opt_lsp_modeline_message_requests"]) {
-            if (output_in_hover == 1) printf "\n"
-            printf "There are unread messages (use lsp-show-message-request-next to read)"
-            output_in_hover = 1
-        }
+        if (info_truncated == 1 || diagnostics_truncated == 1)
+            print "{+i@InfoDefault}Hover info truncated, use lsp-hover-buffer (shortcut H) for full hover info"
+        if (ENVIRON["code_lenses"])
+            print "Code Lenses available (shortcut l)"
+        if (ENVIRON["kak_opt_lsp_modeline_code_actions"])
+            print "Code Actions available (shortcut a)"
+        if (ENVIRON["kak_opt_lsp_modeline_message_requests"])
+            print "There are unread messages (use lsp-show-message-request-next to read)"
     }
 
     function print_at_least_one_line(data, lines) {
-        printf "%s", data[1]
-        for (i = 2; i <= lines; i++) {
-            printf "%s%s", "\n", data[i]
+        for (i = 1; i <= lines; i++) {
+            printf "%s\n", data[i]
         }
     }
 
```

krobelus:
Thanks so much, this is a great fix to a long-standing issue.

The original limit of 20 lines was chosen to fit into a "standard" 80x24 terminal.

In future we can dynamically adapt to the terminal height with something like this:

```
# set smart defaults for max hover lines
set -- $kak_window_range
window_height=$3
# subtract the trailing 4 lines in hover info, also 1 for the Diagnostics header, and 2 for the info box borders
half_height=$(( ($window_height - 4 - 1 - 2) / 2 )) 
echo "set-option lsp_hover_max_info_lines $half_height"
echo "set-option lsp_hover_max_diagnostic_lines $half_height"
```

(of course that's still not a complete solution, because if there are overlong lines, they will be softwrapped..)

PR #728, opened 2024-02-29
Use chain-loading in language server wrapper scripts

deviant:
These shell processes have no need to exist after having launched a language server, and clutter up the process hierarchy unpleasantly.

### DIFF
--- a/kak-lsp.toml
+++ b/kak-lsp.toml
@@ -99,7 +99,7 @@ command = "dls"
 filetypes = ["dart"]
 roots = ["pubspec.yaml", ".git", ".hg"]
 command = "sh"
-args = ["-c", "dart $(dirname $(command -v dart))/snapshots/analysis_server.dart.snapshot --lsp"]
+args = ["-c", "exec dart $(dirname $(command -v dart))/snapshots/analysis_server.dart.snapshot --lsp"]
 
 [language_server.jdtls]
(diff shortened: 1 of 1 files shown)


kak-lsp.toml
```diff
@@ -99,7 +99,7 @@ command = "dls"
 filetypes = ["dart"]
 roots = ["pubspec.yaml", ".git", ".hg"]
 command = "sh"
-args = ["-c", "dart $(dirname $(command -v dart))/snapshots/analysis_server.dart.snapshot --lsp"]
+args = ["-c", "exec dart $(dirname $(command -v dart))/snapshots/analysis_server.dart.snapshot --lsp"]
```
krobelus, line 102: thanks - I don't think it's necessary for dart, the shell already uses exec semantics here

deviant:
Huh, I guess that may have changed at some point, then. I tried out some other POSIX-compatible shells as well: osh and zsh appear to share this behaviour, but the ksh family (including mksh, and oksh), do not. So whilst there probably aren't a lot of them, this would affect *ksh users (some Gentoo users, users of some smaller Linux distros, and all OpenBSD users).

krobelus:
I wonder why `ksh` doesn't do this optimization, it seems very safe to simply not fork and directly exec

PR #735, opened 2024-03-15
Update Arch pkg name

maximbaz:
Hello, I have renamed the Arch package, as per the encouragement in the latest changelog :slightly_smiling_face: 

https://archlinux.org/packages/extra/x86_64/kakoune-lsp/


krobelus:
thanks! Arch handles renames very well

PR #749, opened 2024-05-18
Trim end of diagnostic message on hover

gbrlsnchs:
I don't know why, but servers like clangd provide
diagnostic messages with trimming whitespace,
which messes up with the formatting a little
when seeing the message on hover.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -273,32 +273,35 @@ pub fn format_related_information(
     main_settings: &ServerSettings,
     ctx: &Context,
 ) -> Option<String> {
-    d.related_information.as_ref().map(|infos| {
-        "\n".to_string()
-            + &infos
-                .iter()
-                .map(|info| {
-                    let path = info.location.uri.to_file_path().unwrap();
-                    let filename = path.to_str().unwrap();
-                    let p = get_kakoune_position_with_fallback(
-                        server,
-                        filename,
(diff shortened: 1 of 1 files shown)


gbrlsnchs:
Converting to draft since I probably need to do this only on last message.

krobelus:
I think the main problem is that we add indentation/newlines even if related information is empty:

```diff
diff --git a/src/diagnostics.rs b/src/diagnostics.rs
index e73638f..3ab5514 100644
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -273,7 +273,13 @@ pub fn format_related_information(
     main_settings: &ServerSettings,
     ctx: &Context,
 ) -> Option<String> {
-    d.related_information.as_ref().map(|infos| {
+    let Some(infos) = d.related_information.as_ref() else {
+        return None;
+    };
+    if infos.is_empty() {
+        return None;
+    }
+    Some(
         "\n".to_string()
             + &infos
                 .iter()
@@ -299,6 +305,6 @@ pub fn format_related_information(
                         info.message
                     )
                 })
-                .join("\n")
-    })
+                .join("\n"),
+    )
 }
```

gbrlsnchs:
Thanks for the input, @krobelus! How about now? I've added `.filter(|infos| !infos.is_empty())` and it seems to do the job.

PR #755, opened 2024-06-30
When computing diagnostics, ignore servers without input ranges

natsukagami:


src/language_features/code_action.rs
```diff
@@ -72,7 +72,11 @@ fn code_actions_for_ranges(
     {
         buff_diags
             .iter()
-            .filter(|(server_name, d)| ranges_overlap(d.range, ranges[server_name]))
+            .filter(|(server_name, d)| {
+                ranges
+                    .get(server_name)
+                    .is_some_and(|r| ranges_overlap(d.range, *r))
```
krobelus, line 78: I don't understand how this crash can happen.
`ranges` should always contain all eligible servers.
How to reproduce the problem?

natsukagami:
I ran into this crash while running both texlab and ltex-ls on the same
latex file. Logging seems to show that at some point `ranges` only contain
one entry for `ltex-ls` when requesting code action, so kak-lsp crashes.

I will try to dig deeper later if needed.

krobelus:
I had missed that we're looking at diagnostics here, which may include servers that don't support code actions

PR #764, opened 2024-08-16
chore(deps): run cargo update to fix Rust 1.80 build

philiptaron:
In Rust 1.80.0, the version of the `time` crate included in `kakoune-lsp` no longer builds. These links have more detail.

- https://github.com/rust-lang/rust/issues/127343
- https://internals.rust-lang.org/t/type-inference-breakage-in-1-80-has-not-been-handled-well/21374

I've run `cargo update`, then `cargo build` and `cargo test` and done a smoketest.

I think everything's looking OK after this update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -52,9 +52,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.14"
+version = "0.6.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "418c75fa768af9c03be99d17643f93f79bbba589895012a80e3452a19ddda15b"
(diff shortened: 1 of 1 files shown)


philiptaron:
cc @krobelus

philiptaron:
I confirmed that this fixes the NixOS `kakounePlugins.kakoune-lsp` build as well. 🎉 

Would you mind tagging 17.1.2, @krobelus?

krobelus:
done - it's just a matter of running `ci/commit-new-release.sh 17.1.2` :)
Usually I try to avoid super small releases (to avoid noise, though it's hard to tell if that's a real problem) but this fix seems important enough.
Thanks for taking care of the package!

philiptaron:
Thank you!

Issue #702, opened 2023-11-29
Feature Request: conditional language server selection


aThorp96:
Forgive me if there is already a mechanism to do this.

# The problem
I have multiple Typescript applications I work on. One of which uses Node and the other uses Deno, using `typescript-language-server` and `deno-lsp` respectively. Neither language server works with the other application type, though both projects use `filetype: typescript`. 

Currently I can get around this by using both language servers (config below). This gives me language support, completion, etc, however it also means that I always get erroneous error diagnostics from the wrong server. E.g. In a deno project `typescript-language-server` will complain that my imports cannot be resolved but in a node application instead `deno-lsp` will complain about the imports. 

## Current config:
```
[language_server.typescript-language-server]
filetypes = ["typescript", "javascript"]
roots = ["package.json"]
command = "typescript-language-server"
args = ["--stdio"]

[language_server.deno-lsp]
filetypes = ["typescript", "javascript"]
roots = ["deno.json"]
command = "deno"
args = ["lsp"]

[language_server.deno-lsp.settings.deno]
enable = true
lint = true
```

# Proposed Solution
Currently (I believe) `filetype` is the only mechanism used to determine which language server(s) is selected. However we also have `roots` to define the project root. If it doesn't present a technical challenge or a specification divergence, I believe it would be possible to select the language server based first on `filetype`, then, if multiple servers are selected use only those which have at least one `root` resolved. If no `roots` are supplied, the server is considered to match. As to not break the current config API, it be better to specify a `require_root_match` field, which would cause the server selection to behave as described. 

I'm very open to alternative mechanisms, so long as there would be a way for me to infer (or specify somehow) which language server to use when multiple are matched.

krobelus:
duplicate of https://github.com/kak-lsp/kak-lsp/issues/537 which links to an approach to change the `lsp_cmd` option per project to include different `--config` file arguments

I appreciate this isn't super convenient. I think in general we should try to move away from the toml and allow setting everything via Kakoune options, then we solve this for free.

krobelus:
https://github.com/kakoune-lsp/kakoune-lsp/pull/765 will make it more obvious how to use language servers conditionally.
The config syntax is not final yet

PR #765, opened 2024-08-19
Get rid of kak-lsp.toml in favor of native config options

### LINKED ISSUE #537: Recommended way to select language servers depending on project
[This comment](https://discuss.kakoune.com/t/kak-lsp-tips-for-correctly-setting-up-deno-lsp/1685/10) explains how to select the path to `kak-lsp.toml` based on the project.
Another idea from that thread is to extend `%opt{lsp_config}`, to allow overriding everything in `kak-lsp.toml`.

Not very high priority but a nice improvement.

krobelus:
(current status: this has been a long time coming but the
implementation is still fresh. We should try to simplify the config
syntax even more, and shake out bugs)

Static configuration files like our `kak-lsp.toml` are increasingly
popular, since readily available language bindings allow to quickly
make something that works. On the other hand, many traditional Unix
configuration files are written in a DSL which allow for better UX,
more concise syntax, etc. but requires a custom parser.

As it stands today I don't think `kak-lsp.toml` is a great fit for us:
1. It doesn't allow shell expansion and other expansions familiar to
   Kakoune users. (Of course you could generate the config file but
   that's backwards.)
2. Changing it dynamically (within a single Kakoune session) is
   awkward. It is possible to set `lsp_cmd` at buffer scope but few
   people do. That also forces you to generate more `kak-lsp.toml`
   files.
3. We don't offer a nice way of overriding only parts of the default
   `kak-lsp.toml`. It's either all or nothing. Users who copy the entire
    default config will end up with lots of stale defaults.
4. It encourages users to maintain a separate file just to configure
   the plugin. This is very unorthodox for Kakoune plugins. I
   personally prefer to keep all my Kakoune configuration in one file
   (except for plugins).

Some of these we could probably address independently: For example,
for 3, we could implement some kind of config merging; IIRC that's
what Helix does.

Let's go a different route, removing `kak-lsp.toml` and replacing
it with Kakoune options `lsp_servers`, `lsp_semantic_tokens` and a
few others.

This means we replace

```toml
[language_server.clangd]
filetypes = ["c", "cpp"]
roots = ["compile_commands.json", ".clangd", ".git", ".hg"]
```

with

```kakrc
hook -group lsp-filetype-clangd global WinSetOption filetype=(?:c|cpp) %{
        [language_server.clangd]
        roots = ["compile_commands.json", ".clangd", ".git", ".hg"]
    }
}
```

Which allows the user to easily disable/remove `lsp-filetype-*` hooks selectively.

For backwards compatibility, the new options are ignored if a
`~/.config/kak-lsp/kak-lsp.toml`.

Closes #537, #702

- Maybe allow arbitrary logic for specifying the root directory, see
  https://discord.com/channels/717163870724161556/717165646181564498/1200809299283288094 (cc @topisani) and https://github.com/kakoune-lsp/kakoune-lsp/issues/762


README.asciidoc
```diff
@@ -404,14 +378,15 @@ hook global WinSetOption filetype=<language> %{
 }
 ----
 
-The faces used for semantic tokens and modifiers can be modified in `kak-lsp.toml`, using the `semantic_tokens.faces` array, for example:
+The faces used for semantic tokens and modifiers can be modified by changing the `lsp_semantic_tokens` option, for example:
```
sclu1034, line 381: > modifiers can be modified by changing

Three words with the same meaning right next to each other are probably going to trip up people reading this. I'd drop one of them:

```suggestion
The faces used for semantic tokens and modifiers can be changed through the `lsp_semantic_tokens` option, for example:
```

README.asciidoc
```diff
@@ -423,14 +398,16 @@ kakoune-lsp will find the most specific matching configuration to apply, where s
...
+[source,kak]
 ----
-[semantic_tokens]
-faces = [
-    {face="const_variable_declaration", token="variable", modifiers=["constant","declaration"]},
-    {face="const_variable", token="variable", modifiers=["constant"]},
-    {face="variable", token="variable"},
-]
+set-option global lsp_semantic_tokens %{
+    [semantic_tokens]
```
sclu1034, line 404: You didn't have this header in the snippet above. It's probably fine if the daemon can accept both with minimal complexity, but examples should probably be consistent.

README.asciidoc
```diff
@@ -404,14 +378,15 @@ hook global WinSetOption filetype=<language> %{
...
 ----
-[semantic_tokens]
-faces = [
-    {face="const_variable_declaration", token="variable", modifiers=["constant", "declaration"]},
-]
+set-option global lsp_semantic_tokens %{
+    faces = [
+        {face="const_variable_declaration", token="variable", modifiers=["constant", "declaration"]},
+    ]
+}
```
sclu1034, line 389: Maybe this snippet should be wrapped with another `filetype`-based hook, to demonstrate and encourage that faces can be matched to what the individual servers support.

krobelus:
> Is it possible to override for example just the command part of the config for a filetype?

Not yet but I think we should try to make that work.
Today Kakoune doesn't have arbitrarily nested objects. One option is to add that.
Or fake it with `declare-option str-list lsp_server_clangd_roots; declare-option str-list lsp_sever_clangd_args` which doesn't seem great.
Maybe it's better to tell users to use `set-option buffer lsp_servers %sh{printf %s "$lsp_servers" | sed /roots/cfoo}`, I'm not sure.
We're still not fully off TOML here but it's not clear how to get away from it for server-specific configuration

For `roots` specifically, we will set `.git` and `.hg` once globally; the filetype hooks can prepend to it.
Adding more flexible ways of root detection is also on the table but this is currently done on every request so efficiency is a plus

README.asciidoc
krobelus: Yeah the `[semantic_tokens]` header is gone; users shouldn't need to see it.
`faces = ` is still here; I tried making it just

```
[
   <face-spec1>,
   <face-spec2>,
]
```

but that on it's own is not valid TOML.

Perhaps we should make it a `str-list` option

sclu1034: > that on it's own is not valid TOML

You could have the daemon add the `faces = ` before parsing, I guess.
But a `str-list` would also enable `-add` and `-remove`, so that's probably more useful anyways.

krobelus: Hmm might as well revive https://github.com/kakoune-lsp/kakoune-lsp/pull/558
It looks like this now:

```
set g lsp_semantic_tokens %{
    # face        token     modifiers...
    documentation comment   documentation
    comment       comment
    function      function
    keyword       keyword
    module        namespace
    operator      operator
    string        string
    type          type
    default+d     variable  readonly
    default+d     variable  constant
    variable      variable
}
```

krobelus:
Note that this deprecates/hides `lsp_config`, since `lsp_servers` etc. are already dynamic .
Not sure if this is worth mentioning in the changelog since not too many people use this.
(I still need to add a change hook though it's not super important, the user can always restart servers)

For root detection I'll try something like

	declare-option str-to-str-map lsp_server_root

	define-command lsp-find-root -params 1.. %{
		evaluate-commands %sh{
			server_name=$1
			shift
			for glob; do; 
				cd "$(dirname "$kak_buffile")"
				until [ "$(pwd -P)" = / ]; do
					[ -e $glob ] && break 2
					cd ..
				done
			done
			printf 'set-option buffer lsp_server_root %s=%s' "${server_name}" $(pwd)
		}
	}

	hook global WinSetOption filetype=(?:c|cpp) %{
		lsp-find-root clangd compile_commands.json .clangd .git .hg
	}

This should deprecate our environment variables like `KAK_LSP_PROJECT_ROOT_RUST` and `KAK_LSP_FORCE_PROJECT_ROOT`.
The first is triival to replace with a filetype hook, and so is the second (remove all `lsp-filetype-root` hooks and set your global value).

The root needs to be discovered only once on `BufCreate`, instead of over and over again for every LSP request as it's done today.

sclu1034:
> This should deprecate our environment variables like `KAK_LSP_PROJECT_ROOT_RUST` and `KAK_LSP_FORCE_PROJECT_ROOT`.

I feel like the "overrule all clever detections and just force a value"-like hammer could still be useful to keep around for debugging or special use cases.

krobelus:
I don't plan to remove support for the environment variables in the foreseeable future but I think it's valuable to remove it from the documentation, to make the system easier to learn. Removing `lsp-filetype-.*` hooks should get rid of all automatic root detection. Similarly, your own hook can fairly easily overwrite what the standard hooks did (though I guess only for selected language servers)

I recognize that there is the danger that we're building something that's too much catered to experts.
Everyone can use environment variables and TOML files so having to read `:doc expansions` might pose a small barrier. But I'm fine with encouraging people to learn

---

Looking at the root detection I realized that we can maybe finally add multi-root support.
Today, if a single  Kakoune session opens two projects (let's say `../kakoune-lsp` and `../lsp-types`), there will be two instances of `rust-analyzer`, which is not ideal because of startup time.
Our workaround is `KAK_LSP_FORCE_PROJECT_ROOT` which is bad.

We could have every kakoune-lsp session only ever spawn one instance of each
language server.  We'd implement `DidChangeWorkspaceFolders` (so whenever
we open a file from another project, we add that project's root to the
language server).

Of course that might break
1. language servers that don't support `DidChangeWorkspaceFolders`
2. people who use one Kakoune session for everything and the multiple roots
   somehow interfere with each other (hard to come up with a good scenario but I
   guess a language server like rust-analyzer might run out of memory easier)

   Note that one workaround is to use use multiple kakoune-lsp sessions
   within a single Kakoune session. Maybe we can add an `lsp_session` option
   to control the `kak-lsp -s` argument. So if you really want to isolate
   them within a single Kakoune session you can still do that. We're still
   much more flexible than VSCode, but we should also be user-friendly :)

So I think we can add multi-root for the servers that support it.

Somehow I previously thought multi-root is not feasible because we only have one `$PWD` but that doesn't seem relevant.

krobelus:
For now I broke the option to use different project roots for two servers in a single buffer.
I'm not sure why this would be necessary but i guess we can add it back if needed.

krobelus:
If I do something like `set-option -add buffer lsp_servers %{ [language_server.pyright-langserver] args = ["--stdio"] }`,
the additional language server will be started as needed.
If I remove that snippet from the option again, we'll stop talking to `pyright` but leave it running.
I guess this is acceptable. In future we could also shut it down though that might complicate the implementation.

sclu1034:
I guess the question for that is: Do people have use cases where they use a single, continuous Kakoune instance that jumps between projects?
If so, then they might pile up language servers in the background still running from projects they worked on over the lifetime of that instance.

krobelus:
Right, I forgot about that. I think the one-session-for-everything use case is an important one (although I have never used it).

Before multi-root support, every project opened from the same kakoune-lsp session got its own kak-lsp controller (which can talk to multiple language servers). The controllers are pretty much completely independent of each other. So with that setup, we could have added a timeout to each controller to address this (existing) issue. Maybe the reason why it hasn't been a problem is that we've had a session timeout, so after 30 minutes of inactivity all controllers will be shut down.

With multi-root support it doesn't make sense to compartmentalize servers like that, so I removed the multiple-controllers code. All language servers share one controller, which allows the controller to easily check who supports `workspaceFolders` and reuse those instead of launching a new one. To get rid of unused language servers, we could add a per-server timeout.

nvim-lspconfig does a bunch of extra stuff for rust-analyzer; they only reuse the rust-analyzer instance when visiting libraries, and they open stdlib files in "single file mode".

krobelus:
planning to merge soon but LMK if you want to test/review first.
I wonder if we should loudly stop supporting `kak-lsp.toml`, to reduce confusion from having different config flavors in the wild.. I guess there's no rush

topisani:
> I wonder if we should loudly stop supporting `kak-lsp.toml`, to reduce confusion from having different config flavors in the wild.. I guess there's no rush

An option could be to use this chance to also rename the binary to kakoune-lsp, and make one big breaking release. I think this fundamental config change could be a chance to clean up some backwards compatibility in a few places

README.asciidoc
krobelus: @topisani on discord pointed out that

> i think the token.modifier.modifier is the standard syntax for textmate scopes which the standard is based on, so the only thing is it would make it more in line with other systems

so how about we change the order, since the user already needs to do make manual changes to the config (once they chose to delete their `kak-lsp.toml`). I think putting the face first is weird.

sclu1034: But putting it last would mean that you can't rely on its position. Since the amount modifiers is variable, putting a token after them could land at column 30 in one row, and at column 60 in the next.
You wouldn't be able to compare faces across lines anymore.

krobelus: I'd align them (using `&`) like the `lsp_semantic_tokens` above

krobelus:
> An option could be to use this chance to also rename the binary to kakoune-lsp, and make one big breaking release. I think this fundamental config change could be a chance to clean up some backwards compatibility in a few places

There are definitely a bunch of breaking changes that we could do:
- making language servers opt-in instead of opt-out.
- reject arguments set in `lsp_cmd` (especially `-s` which is no longer necessary)
- remove a bunch of options/commands marked as DEPRECATED

but that can be a separate decision, until then we can keep things working as before.

krobelus:
Merged as 5593ce5e

evannjohnson:
I arrived here from [`fe4b481`](https://github.com/kakoune-lsp/kakoune-lsp/commit/fe4b481), and I'm seeking a way to disable the "reuse server if it supports workspaceFolders" behavior. My use case is that I set certain LSP settings when opening some files, to include external libraries within the workspace. However, if I then navigate to a file in the external library within that Kakoune session, it sets the `lsp_servers` config without the additional settings of the main project I am working in. Since the server I'm using supports workspaceFolders, the same server is used but its configuration is changed to drop the project-specific config I had set.

I've tried a few ways to get this configuration to be applied again when switching back to the buffer in the project, or for the config not to be overridden when opening files from the library, but I haven't been successful. Disabling this "reuse server" feature, while maybe not the most elegant solution, does make it behave nicely with how I'm trying to use it. It seems like might be fairly simple to allow some configuration in `lsp_servers` that disables this behavior.

This is how I'm setting this project-specific config:
```
hook -group norns global BufCreate %opt{norns_hook_regex} %{
    set-option buffer lsp_settings %{
        diagnostics.unusedLocalExclude = ["i"]
        workspace.library = ["/Users/evanjohnson/repos/norns-things/norns/lua/core/"]
    }
}
```

```
hook global WinSetOption filetype=(lua) %{
    lsp-enable-window
}

hook -group lsp-filetype-lua global BufSetOption filetype=lua %{
        [lua-language-server]
        args = ["lsp"]
        root_globs = [".git", ".hg"]
        settings_section = "Lua"
        [lua-language-server.settings.Lua]
        diagnostics.enable = true
    }
    set-option buffer lsp_servers "
        %opt{lsp_servers}
        %opt{lsp_settings}
    "
}

I'm open to just going about things in my configuration in a different way, but simply allowing a different server for each project root within a session if the user wants it seems like useful flexibility.

krobelus:
pushed commit 642ca2b4 with a config knob to override the magic
single-instance behavior.

I'm not yet sure what we should add to LSP (if anything) to let
servers decide this for us. It's a bit complicated..

> Or, to find a method to set certain LSP settings based on the filepath of a buffer (a project, really),

Today there are three different ways to communicate settings like `diagnostics.unusedLocalExclude`:
1. `initializationOption`
2. `workspace/didChangeConfiguration`
3. `workspace/configuration`

The first two can't be scoped to a specific project (when multiple
projects are opened in a single session).

`workspace/configuration` does have a scope argument (the path to the project).

But kakoune-lsp doesn't really expose to the user a list of active
projects (as detailed in 642ca2b4).

I guess we could answer `workspace/configuration` with whatever is
set in the first buffer that belongs to the given project project.

But if separate instances work fine, I'd wait with adding complexity.
But it would be good if we could figure out a more reliable way to detect it.

evannjohnson:
Working well for me. Thank you for this, and for all the work on this plugin - it's such a useful tool.

PR #771, opened 2024-09-08
gopls: ui.completion prefix is not required

vbauerster:
Tested with gopls 0.16.1


rc/servers.kak
```diff
@@ -139,7 +139,7 @@ hook -group lsp-filetype-go global User LSPDefaultConfig=go %{
             hints.functionTypeParameters = true
             hints.parameterNames = true
             hints.rangeVariableTypes = true
-            ""ui.completion.usePlaceholders"" = true
+            usePlaceholders = true
```
krobelus, line 142: Thank you! I think when I added this I was looking at gopls source code and probably got confused

PR #772, opened 2024-09-08
lsp_semantic_tokens: defaults for method and builtinType

vbauerster:
At least both gopls and rust-analyzer declare method and builtinType semantic tokens, so why not cover them by default.


rc/lsp.kak
```diff
@@ -58,6 +58,8 @@ declare-option -docstring %{
         {face="documentation", token="comment", modifiers=["documentation"]},
         {face="comment", token="comment"},
         {face="function", token="function"},
+        {face="function", token="method"},
+        {face="builtin", token="builtinType"},
```
krobelus, line 62: Thanks!

Looks like there is a lot of overlap between `gopls` and `rust-analyzer`. (I mixed up the order but that's okay).

```
$ comm gopls-tokens rust-analyzer-tokens
		angle
		arithmetic
		attribute
		attributeBracket
		bitwise
		boolean
		brace
		bracket
		builtinAttribute
		builtinType
		character
		colon
		comma
		comment
		comparison
const
		constParameter
		decorator
		derive
		deriveHelper
		dot
		enum
		enumMember
		escapeSequence
		formatSpecifier
		function
		generic
		interface
		invalidEscapeSequence
		keyword
		label
		lifetime
		logical
		macro
		macroBang
		method
		namespace
		number
		operator
		parameter
		parenthesis
procMacro
		property
		punctuation
		selfKeyword
		selfTypeKeyword
		semicolon
static
		string
		struct
		toolModule
		typeAlias
		typeParameter
		union
		unresolvedReference
		variable
```

There will always be servers-specific token types but for the shared ones we could define our own faces if there is no builtin one

krobelus (approved):

rc/lsp.kak
vbauerster: I was confused by gopls' capabilities info box. It doesn't declare token `builtinType` so I removed that.
https://github.com/golang/tools/blob/master/gopls/doc/semantictokens.md

sclu1034: The LSP spec also added `decorator` in 3.17

krobelus: sorry I messed up my `comm` command, I must have been comparing two logs from `rust-analyzer`.

My gopls 0.16.1 also doesn't declare `method`:

` lsp-semantic-tokens: types: [comment, function, namespace, variable, keyword, string, operator, type] [gopls]`

But still, things like `method` seems like good idea to add (or does `rust-analyzer` already tell us that every `method` is a `function`?)

sclu1034: I think the concept of a "method" is common enough across many languages that we can expect some servers to make use of it. The Typescript server has it, too.

vbauerster: > My gopls 0.16.1 also doesn't declare method:

gopls' `:lsp-capabilities` info box lists whatever is set by client, not what it actually declares. That's why I was confused. Source of truth for gopls is [documentation](https://github.com/golang/tools/blob/master/gopls/doc/semantictokens.md#semantic-tokens-for-go-files).

krobelus: > gopls' :lsp-capabilities info box lists whatever is set by client, not what it actually declares.

Wow this is extremely confusing. 

Since `lsp_semantic_tokens` is global (well, at least it applies to all servers associated with the current buffer),
we might want to change the `lsp-capabilities` output, to make it less confusing. Maybe it's a feature that language servers ignore unknown token types...

Is it possible to want conflicting semantic tokens configurations for different servers?
If yes, then we could move this into `lsp_servers`.
But TBH the same applies to a lot of options from `lsp.kak`.
If we could link [arbitrary scopes](https://github.com/mawww/kakoune/pull/4152) to each buffer that might work.
Anyway, not super important.

sclu1034: Separating token configuration per server would likely make it a lot harder for the user to create uniform themes across the servers.
I don't see anything specific about this in the spec, so I guess as long as servers are fine with just ignoring tokens they don't support, we should keep it global.

The capabilities could be built from the `SemanticTokensLegend` that each server sends.

sclu1034: Actually, `SemanticTokensLegend` probably isn't ideal either, as it would only show what the client and server agree on. That would make it useless for something like "which tokens does the server support that I don't have in my config yet?".

krobelus: well `rust-analyzer`'s legend contains all the token types that `rust-analyzer` supports (independent of whether the client supports them)
and `gopls`'s legend contains exactly the tokens that the client requests, whether they are supported or not.

If servers behave like rust-analyzer then we can compute the intersection etc. and display that; that would be preferrable to us

PR #775, opened 2024-09-09
Remove lsp-if-no-servers hook ordering workaround

krobelus:
`lsp-if-no-servers` exists because we want the user to be able to
override servers using familiar hooks:

	eval %sh{kak-lsp --kakoune}
	hook global BufSetOption filetype=(?:c|cpp) %{
	        [clangd]
	        root = "/path/to/project"
	    }
	}

without regard to hook ordering issues. The hook ordering issues should
only arise when the user's hook is defined before kak-lsp is loaded
(when the "eval %sh{kak-lsp --kakoune}" happens after the hook).

We can probably assume this doesn't happen, simplifying the default config.

---

The LSPDefaultConfig hook exists to be able to delay
root detection etc. until after `lsp-enable`.
This has two reasons:
1. it also works for `kak main.c -e 'eval %sh{kak-lsp --kakoune}'`
2. no shell calls until `lsp-enable`

I realized that reason 1 is not so important because one can use
`kak -E`.  Reason 2 is probably not a big deal. In future we could
change root detection again to be declarative in most cases, removing
this reason.

Remove it to reduce complexity for users looking at the default config.

Alternatively we could hide the default config better.

---

cc @alexherbo2 @topisani

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -584,7 +584,7 @@ Please also try to reproduce the issue with a minimal configuration. Sometimes t
 
 [source,sh]
 ----
-env -u XDG_CONFIG_HOME HOME=$(mktemp -d) kak -e '
+env -u XDG_CONFIG_HOME HOME=$(mktemp -d) kak -E '
     eval %sh{kak-lsp --kakoune}
     set global lsp_debug true       # Check the *debug* buffer for debug logs.
     lsp-enable'
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -504,16 +504,6 @@ declare-option -hidden -docstring %{
     }
     find_root}
 
-define-command -hidden lsp-if-no-servers -docstring %{
-    Run the given commands if the 'lsp_servers' option is empty
-} -params 1 %{
-    evaluate-commands %sh{
-        if [ -z "$kak_opt_lsp_servers" ]; then
-            echo evaluate-commands %arg{1}
-        fi
-    }
-}
-
 ### Requests ###
 
 define-command lsp-start -docstring "Start kakoune-lsp session" %{ nop %sh{
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -2652,12 +2652,8 @@ map global goto y '<esc>:lsp-type-definition<ret>' -docstring 'type definition'
 ### Default integration ###
 
 define-command lsp-enable -docstring "Default LSP integration" %{
-    hook -group lsp global BufSetOption filetype=(.*) %{
-        trigger-user-hook "LSPDefaultConfig=%val{hook_param_capture_1}"
-    }
     lsp-enable-impl global
     evaluate-commands -buffer * %{
-        trigger-user-hook "LSPDefaultConfig=%opt{filetype}" # For kak somefile -e '%eval{kak-lsp --kakoune}'
         hook -group lsp buffer BufSetOption (?:lsp_servers|lsp_config|lsp_server_configuration)=.* lsp-did-change-config
     }
     hook -group lsp global BufClose .* lsp-did-close
@@ -2678,7 +2674,6 @@ define-command lsp-disable -docstring "Disable LSP" %{
 }
 
 define-command lsp-enable-window -docstring "Default LSP integration in the window scope" %{
-    trigger-user-hook "LSPDefaultConfig=%opt{filetype}"
     lsp-enable-impl window
     hook -group lsp window WinClose .* lsp-did-close
     hook -group lsp window WinSetOption (?:lsp_servers|lsp_config|lsp_server_configuration)=.* lsp-did-change-config
--- a/rc/servers.kak
+++ b/rc/servers.kak
@@ -2,14 +2,14 @@
 #
 # For help, see the docstring of the 'lsp_servers' option.
 
-hook -group lsp-filetype-c-family global User LSPDefaultConfig=(?:c|cpp|objc) %{
+hook -group lsp-filetype-c-family global BufSetOption filetype=(?:c|cpp|objc) %{
     set-option buffer lsp_servers "
         [clangd]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" compile_commands.json .clangd .git .hg $(: kak_buffile)}""
     "
 }
 
-hook -group lsp-filetype-clojure global User LSPDefaultConfig=clojure %{
+hook -group lsp-filetype-clojure global BufSetOption filetype=clojure %{
     set-option buffer lsp_servers "
         [clojure-lsp]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" project.clj .git .hg $(: kak_buffile)}""
@@ -20,21 +20,21 @@ hook -group lsp-filetype-clojure global User LSPDefaultConfig=clojure %{
     "
 }
 
-hook -group lsp-filetype-cmake global User LSPDefaultConfig=make %{
+hook -group lsp-filetype-cmake global BufSetOption filetype=make %{
     set-option buffer lsp_servers "
         [cmake-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" CMakeLists.txt .git .hg $(: kak_buffile)}""
     "
 }
 
-hook -group lsp-filetype-crystal global User LSPDefaultConfig=crystal %{
+hook -group lsp-filetype-crystal global BufSetOption filetype=crystal %{
     set-option buffer lsp_servers "
         [crystalline]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" shard.yml $(: kak_buffile)}""
     "
 }
 
-hook -group lsp-filetype-css global User LSPDefaultConfig=(?:css|less|scss) %{
+hook -group lsp-filetype-css global BufSetOption filetype=(?:css|less|scss) %{
     set-option buffer lsp_servers "
         [vscode-css-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" package.json .git .hg $(: kak_buffile)}""
@@ -42,14 +42,14 @@ hook -group lsp-filetype-css global User LSPDefaultConfig=(?:css|less|scss) %{
     "
 }
 
-hook -group lsp-filetype-d global User LSPDefaultConfig=(?:d|di) %{
+hook -group lsp-filetype-d global BufSetOption filetype=(?:d|di) %{
     set-option buffer lsp_servers "
         [dls]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .git dub.sdl dub.json $(: kak_buffile)}""
     "
 }
 
-hook -group lsp-filetype-dart global User LSPDefaultConfig=dart %{
+hook -group lsp-filetype-dart global BufSetOption filetype=dart %{
     set-option buffer lsp_servers "
         [dart-lsp]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" pubspec.yaml .git .hg $(: kak_buffile)}""
@@ -59,7 +59,7 @@ hook -group lsp-filetype-dart global User LSPDefaultConfig=dart %{
     "
 }
 
-hook -group lsp-filetype-elixir global User LSPDefaultConfig=elixir %{
+hook -group lsp-filetype-elixir global BufSetOption filetype=elixir %{
     set-option buffer lsp_servers "
         [elixir-ls]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" mix.exs $(: kak_buffile)}""
@@ -70,7 +70,7 @@ hook -group lsp-filetype-elixir global User LSPDefaultConfig=elixir %{
     "
 }
 
-hook -group lsp-filetype-elm global User LSPDefaultConfig=elm %{
+hook -group lsp-filetype-elm global BufSetOption filetype=elm %{
     set-option buffer lsp_servers "
         [elm-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" elm.json $(: kak_buffile)}""
@@ -85,7 +85,7 @@ hook -group lsp-filetype-elm global User LSPDefaultConfig=elm %{
     "
 }
 
-hook -group lsp-filetype-elvish global User LSPDefaultConfig=elvish %{
+hook -group lsp-filetype-elvish global BufSetOption filetype=elvish %{
     set-option buffer lsp_servers "
         [elvish]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .git .hg $(: kak_buffile)}""
@@ -93,7 +93,7 @@ hook -group lsp-filetype-elvish global User LSPDefaultConfig=elvish %{
     "
 }
 
-hook -group lsp-filetype-erlang global User LSPDefaultConfig=erlang %{
+hook -group lsp-filetype-erlang global BufSetOption filetype=erlang %{
     set-option buffer lsp_servers "
         [erlang_ls]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" rebar.config erlang.mk .git .hg $(: kak_buffile)}""
@@ -102,7 +102,7 @@ hook -group lsp-filetype-erlang global User LSPDefaultConfig=erlang %{
     "
 }
 
-hook -group lsp-filetype-go global User LSPDefaultConfig=go %{
+hook -group lsp-filetype-go global BufSetOption filetype=go %{
     set-option buffer lsp_servers "
         [gopls]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" Gopkg.toml go.mod .git .hg $(: kak_buffile)}""
@@ -120,7 +120,7 @@ hook -group lsp-filetype-go global User LSPDefaultConfig=go %{
     "
 }
 
-hook -group lsp-filetype-haskell global User LSPDefaultConfig=haskell %{
+hook -group lsp-filetype-haskell global BufSetOption filetype=haskell %{
     set-option buffer lsp_servers "
         [haskell-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" hie.yaml cabal.project Setup.hs stack.yaml '*.cabal' $(: kak_buffile)}""
@@ -133,7 +133,7 @@ hook -group lsp-filetype-haskell global User LSPDefaultConfig=haskell %{
     "
 }
 
-hook -group lsp-filetype-html global User LSPDefaultConfig=html %{
+hook -group lsp-filetype-html global BufSetOption filetype=html %{
     set-option buffer lsp_servers "
         [vscode-html-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" package.json $(: kak_buffile)}""
@@ -145,7 +145,7 @@ hook -group lsp-filetype-html global User LSPDefaultConfig=html %{
     "
 }
 
-hook -group lsp-filetype-javascript global User LSPDefaultConfig=(?:javascript|typescript) %{
+hook -group lsp-filetype-javascript global BufSetOption filetype=(?:javascript|typescript) %{
     set-option -add buffer lsp_servers "
         [typescript-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" package.json tsconfig.json jsconfig.json .git .hg $(: kak_buffile)}""
@@ -188,7 +188,7 @@ hook -group lsp-filetype-javascript global User LSPDefaultConfig=(?:javascript|t
     # "
 }
 
-hook -group lsp-filetype-java global User LSPDefaultConfig=java %{
+hook -group lsp-filetype-java global BufSetOption filetype=java %{
     set-option buffer lsp_servers "
         [jdtls]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" mvnw gradlew .git .hg $(: kak_buffile)}""
@@ -198,7 +198,7 @@ hook -group lsp-filetype-java global User LSPDefaultConfig=java %{
     "
 }
 
-hook -group lsp-filetype-json global User LSPDefaultConfig=json %{
+hook -group lsp-filetype-json global BufSetOption filetype=json %{
     set-option buffer lsp_servers "
         [vscode-json-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" package.json $(: kak_buffile)}""
@@ -206,7 +206,7 @@ hook -group lsp-filetype-json global User LSPDefaultConfig=json %{
     "
 }
 
-hook -group lsp-filetype-julia global User LSPDefaultConfig=julia %{
+hook -group lsp-filetype-julia global BufSetOption filetype=julia %{
     set-option buffer lsp_servers "
         # Requires Julia package ""LanguageServer""
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" Project.toml .git .hg $(: kak_buffile)}""
@@ -240,7 +240,7 @@ hook -group lsp-filetype-julia global User LSPDefaultConfig=julia %{
     "
 }
 
-hook -group lsp-filetype-latex global User LSPDefaultConfig=latex %{
+hook -group lsp-filetype-latex global BufSetOption filetype=latex %{
     set-option buffer lsp_servers "
         [texlab]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .git .hg $(: kak_buffile)}""
@@ -272,7 +272,7 @@ hook -group lsp-filetype-latex global User LSPDefaultConfig=latex %{
     "
 }
 
-hook -group lsp-filetype-lua global User LSPDefaultConfig=lua %{
+hook -group lsp-filetype-lua global BufSetOption filetype=lua %{
     set-option buffer lsp_servers "
         [lua-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .git .hg $(: kak_buffile)}""
@@ -283,7 +283,7 @@ hook -group lsp-filetype-lua global User LSPDefaultConfig=lua %{
     "
 }
 
-hook -group lsp-filetype-markdown global User LSPDefaultConfig=markdown %{
+hook -group lsp-filetype-markdown global BufSetOption filetype=markdown %{
     set-option -add buffer lsp_servers "
         [marksman]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .marksman.toml $(: kak_buffile)}""
@@ -296,21 +296,21 @@ hook -group lsp-filetype-markdown global User LSPDefaultConfig=markdown %{
     # "
 }
 
-hook -group lsp-filetype-nim global User LSPDefaultConfig=nim %{
+hook -group lsp-filetype-nim global BufSetOption filetype=nim %{
     set-option buffer lsp_servers "
         [nimlsp]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" '*.nimble' .git .hg $(: kak_buffile)}""
     "
 }
 
-hook -group lsp-filetype-nix global User LSPDefaultConfig=nix %{
+hook -group lsp-filetype-nix global BufSetOption filetype=nix %{
     set-option buffer lsp_servers "
         [nil]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" flake.nix shell.nix .git .hg $(: kak_buffile)}""
     "
 }
 
-hook -group lsp-filetype-ocaml global User LSPDefaultConfig=ocaml %{
+hook -group lsp-filetype-ocaml global BufSetOption filetype=ocaml %{
     set-option buffer lsp_servers "
         [ocamllsp]
         # Often useful to simply do a `touch dune-workspace` in your project root folder if you have problems with root detection
@@ -321,7 +321,7 @@ hook -group lsp-filetype-ocaml global User LSPDefaultConfig=ocaml %{
     "
 }
 
-hook -group lsp-filetype-php global User LSPDefaultConfig=php %{
+hook -group lsp-filetype-php global BufSetOption filetype=php %{
     set-option buffer lsp_servers "
         [intelephense]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .htaccess composer.json $(: kak_buffile)}""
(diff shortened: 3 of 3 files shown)


vbauerster:
How about following approach? Sort of dependency injection style :) 
**On plugin side:**
```kak
hook global BufSetOption filetype=go %{
  try %{
    require-module lsp-server-go
    set-option buffer lsp_servers %opt{lsp_server_go}
  } catch %{
    set-option buffer lsp_servers "default gopls config"
  }
}
hook global BufSetOption filetype=rust %{
  try %{
    require-module lsp-server-rust
    set-option buffer lsp_servers %opt{lsp_server_rust}
  } catch %{
    set-option buffer lsp_servers "default rust-analyzer config"
  }
}
```
**On user side:**
If user wants to provide server-specific configuration then he/she needs to provide-module with config declaration like:
```kak
provide-module lsp-server-go %{
  declare-option str lsp_server_go "user defined [gopls]"
}
provide-module lsp-server-rust %{
  declare-option str lsp_server_rust "user defined [rust-analyzer]"
}
```

Naming convention for module: `lsp-server-{filetype}` and for option: `lsp_server_{filetype}`, of course it can be changed.

krobelus:
I forgot to mention that the goal of this change is to make the default config copy-paste friendly.
The user should be able to copy and modify the default config and it will just work.

I realized there is a way to achieve this goal, while maintaining today's behavior, see below.
I'm not saying that this behavior is so important to warrant all this complexity; I don't know how many people would trip over hook ordering issues.
But I do have a preference for delaying the BufSetOption shell-calls until `lsp-enable` because that's what I expect as user. Loading should happen eagerly but the rest should be lazy.

> provide-module lsp-server-go %{
  declare-option str lsp_server_go "user defined [gopls]"
}

Here we'd want to make sure this works for filetypes that kakoune-lsp doesn't know about, else it's too hard to define LSP integration for those.

The advantage of `provide-module` is that by defining a module like `lsp-server-go`,
the default config would be completely disabled (instead of being set but immediately overridden by the user hook).
This is great.

For the same reason, it seems a bit less copy-paste friendly. Given

	```
	hook global BufSetOption filetype=(.*) %{
		try %{
			require-module "lsp-server-%val{hook_param_capture_1}"
			evaluate-commands set-option buffer lsp_servers "%%opt{lsp_server_%val{hook_param_capture_1}}"
		} catch %{
			try %{
				require-module "lsp-default-server-%val{hook_param_capture_1}"
				evaluate-commands set-option buffer lsp_servers "%%opt{lsp_server_%val{hook_param_capture_1}}"
			}
		}
	}
	provide-module lsp-default-server-go %{
		declare-option str lsp_server_go "[gopls] ...."
	}
	```

How do we make it obvious that the user should `provide-module lsp-server-go`?

---

```
index 6a1f58f5..7d040f3d 100644
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -514,6 +514,27 @@ define-command -hidden lsp-if-no-servers -docstring %{
     }
 }
 
+define-command -hidden lsp-if-no-language-id -docstring %{
+    Run the given commands if the 'lsp_language_id' option is empty
+} -params 1 %{
+    evaluate-commands %sh{
+        if [ -z "$kak_opt_lsp_language_id" ]; then
+            echo evaluate-commands %arg{1}
+        fi
+    }
+}
+
+define-command -hidden lsp-load-default-config -params 1 -docstring %{
+    Register the given BufSetOption filetype hooks in a way that they only execute after LSP
+    is enabled.
+} %{
+    evaluate-commands %sh{
+        printf %s "$1" |
+        sed -e '/-group-lsp-filetype/s/BufSetOption filetype=/User LSPDefaultConfig/' \
+            -e '/-group-lsp /s/BufSetOption filetype=/User LSPDefaultLanguageId/'
+    }
+}
+
 ### Requests ###
 
 define-command lsp-start -docstring "Start kakoune-lsp session" %{ nop %sh{
@@ -2663,11 +2684,22 @@ map global goto y '<esc>:lsp-type-definition<ret>' -docstring 'type definition'
 
 define-command lsp-enable -docstring "Default LSP integration" %{
     hook -group lsp global BufSetOption filetype=(.*) %{
+        lsp-if-no-servers %{
             trigger-user-hook "LSPDefaultConfig=%val{hook_param_capture_1}"
         }
+        lsp-if-no-language-id %{
+            trigger-user-hook "LSPDefaultLanguageId=%val{hook_param_capture_1}"
+        }
+    }
     lsp-enable-impl global
     evaluate-commands -buffer * %{
-        trigger-user-hook "LSPDefaultConfig=%opt{filetype}" # For kak somefile -e '%eval{kak-lsp --kakoune}'
+        # For kak somefile -e '%eval{kak-lsp --kakoune}'
+        lsp-if-no-servers %{
+            trigger-user-hook "LSPDefaultConfig=%opt{filetype}"
+        }
+        lsp-if-no-language-id %{
+            trigger-user-hook "LSPDefaultLanguageId=%opt{filetype}"
+        }
         hook -group lsp buffer BufSetOption (?:lsp_servers|lsp_config|lsp_server_configuration)=.* lsp-did-change-config
     }
     hook -group lsp global BufClose .* lsp-did-close
@@ -2688,7 +2720,12 @@ define-command lsp-disable -docstring "Disable LSP" %{
 }
 
 define-command lsp-enable-window -docstring "Default LSP integration in the window scope" %{
+    lsp-if-no-servers %{
         trigger-user-hook "LSPDefaultConfig=%opt{filetype}"
+    }
+    lsp-if-no-language-id %{
+        trigger-user-hook "LSPDefaultLanguageId=%opt{filetype}"
+    }
     lsp-enable-impl window
     hook -group lsp window WinClose .* lsp-did-close
     hook -group lsp window WinSetOption (?:lsp_servers|lsp_config|lsp_server_configuration)=.* lsp-did-change-config
diff --git a/rc/servers.kak b/rc/servers.kak
index 391cd919..9d30dc56 100644
--- a/rc/servers.kak
+++ b/rc/servers.kak
@@ -2,17 +2,16 @@
 #
 # For help, see the docstring of the 'lsp_servers' option.
 
-hook -group lsp-filetype-c-family global User LSPDefaultConfig=(?:c|cpp|objc) %{
-    lsp-if-no-servers %{
+lsp-load-default-config %{
+
+hook -group lsp-filetype-c-family global BufSetOption filetype=(?:c|cpp|objc) %{
     set-option buffer lsp_servers "
         [clangd]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" compile_commands.json .clangd .git .hg $(: kak_buffile)}""
     "
 }
-}
 
-hook -group lsp-filetype-clojure global User LSPDefaultConfig=clojure %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-clojure global BufSetOption filetype=clojure %{
     set-option buffer lsp_servers "
         [clojure-lsp]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" project.clj .git .hg $(: kak_buffile)}""
@@ -22,47 +21,37 @@ hook -group lsp-filetype-clojure global User LSPDefaultConfig=clojure %{
         # source-paths-ignore-regex = [""resources.*"", ""target.*""]
     "
 }
-}
 
-hook -group lsp-filetype-cmake global User LSPDefaultConfig=make %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-cmake global BufSetOption filetype=make %{
     set-option buffer lsp_servers "
         [cmake-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" CMakeLists.txt .git .hg $(: kak_buffile)}""
     "
 }
-}
 
-hook -group lsp-filetype-crystal global User LSPDefaultConfig=crystal %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-crystal global BufSetOption filetype=crystal %{
     set-option buffer lsp_servers "
         [crystalline]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" shard.yml $(: kak_buffile)}""
     "
 }
-}
 
-hook -group lsp-filetype-css global User LSPDefaultConfig=(?:css|less|scss) %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-css global BufSetOption filetype=(?:css|less|scss) %{
     set-option buffer lsp_servers "
         [vscode-css-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" package.json .git .hg $(: kak_buffile)}""
         args = [""--stdio""]
     "
 }
-}
 
-hook -group lsp-filetype-d global User LSPDefaultConfig=(?:d|di) %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-d global BufSetOption filetype=(?:d|di) %{
     set-option buffer lsp_servers "
         [dls]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .git dub.sdl dub.json $(: kak_buffile)}""
     "
 }
-}
 
-hook -group lsp-filetype-dart global User LSPDefaultConfig=dart %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-dart global BufSetOption filetype=dart %{
     set-option buffer lsp_servers "
         [dart-lsp]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" pubspec.yaml .git .hg $(: kak_buffile)}""
@@ -71,10 +60,8 @@ hook -group lsp-filetype-dart global User LSPDefaultConfig=dart %{
         args = [""-c"", ""dart \""$(dirname \""$(command -v dart)\"")\""/snapshots/analysis_server.dart.snapshot --lsp""]
     "
 }
-}
 
-hook -group lsp-filetype-elixir global User LSPDefaultConfig=elixir %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-elixir global BufSetOption filetype=elixir %{
     set-option buffer lsp_servers "
         [elixir-ls]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" mix.exs $(: kak_buffile)}""
@@ -84,10 +71,8 @@ hook -group lsp-filetype-elixir global User LSPDefaultConfig=elixir %{
         # dialyzerEnable = true
     "
 }
-}
 
-hook -group lsp-filetype-elm global User LSPDefaultConfig=elm %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-elm global BufSetOption filetype=elm %{
     set-option buffer lsp_servers "
         [elm-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" elm.json $(: kak_buffile)}""
@@ -101,20 +86,16 @@ hook -group lsp-filetype-elm global User LSPDefaultConfig=elm %{
         elmTestPath = ""elm-test""
     "
 }
-}
 
-hook -group lsp-filetype-elvish global User LSPDefaultConfig=elvish %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-elvish global BufSetOption filetype=elvish %{
     set-option buffer lsp_servers "
         [elvish]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .git .hg $(: kak_buffile)}""
         args = [""-lsp""]
     "
 }
-}
 
-hook -group lsp-filetype-erlang global User LSPDefaultConfig=erlang %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-erlang global BufSetOption filetype=erlang %{
     set-option buffer lsp_servers "
         [erlang_ls]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" rebar.config erlang.mk .git .hg $(: kak_buffile)}""
@@ -122,10 +103,8 @@ hook -group lsp-filetype-erlang global User LSPDefaultConfig=erlang %{
         # how to configure. This default config should work in most cases though.
     "
 }
-}
 
-hook -group lsp-filetype-go global User LSPDefaultConfig=go %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-go global BufSetOption filetype=go %{
     set-option buffer lsp_servers "
         [gopls]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" Gopkg.toml go.mod .git .hg $(: kak_buffile)}""
@@ -142,10 +121,8 @@ hook -group lsp-filetype-go global User LSPDefaultConfig=go %{
         usePlaceholders = true
     "
 }
-}
 
-hook -group lsp-filetype-haskell global User LSPDefaultConfig=haskell %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-haskell global BufSetOption filetype=haskell %{
     set-option buffer lsp_servers "
         [haskell-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" hie.yaml cabal.project Setup.hs stack.yaml '*.cabal' $(: kak_buffile)}""
@@ -157,10 +134,8 @@ hook -group lsp-filetype-haskell global User LSPDefaultConfig=haskell %{
         # haskell.formattingProvider = ""ormolu""
     "
 }
-}
 
-hook -group lsp-filetype-html global User LSPDefaultConfig=html %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-html global BufSetOption filetype=html %{
     set-option buffer lsp_servers "
         [vscode-html-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" package.json $(: kak_buffile)}""
@@ -171,10 +146,8 @@ hook -group lsp-filetype-html global User LSPDefaultConfig=html %{
         # javascript.format.semicolons = ""insert""
     "
 }
-}
 
-hook -group lsp-filetype-javascript global User LSPDefaultConfig=(?:javascript|typescript) %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-javascript global BufSetOption filetype=(?:javascript|typescript) %{
     set-option -add buffer lsp_servers "
         [typescript-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" package.json tsconfig.json jsconfig.json .git .hg $(: kak_buffile)}""
@@ -216,10 +189,8 @@ hook -group lsp-filetype-javascript global User LSPDefaultConfig=(?:javascript|t
     #     codeAction.showDocumentation = { enable = false }
     # "
 }
-}
 
-hook -group lsp-filetype-java global User LSPDefaultConfig=java %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-java global BufSetOption filetype=java %{
     set-option buffer lsp_servers "
         [jdtls]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" mvnw gradlew .git .hg $(: kak_buffile)}""
@@ -228,20 +199,16 @@ hook -group lsp-filetype-java global User LSPDefaultConfig=java %{
         # ""java.format.insertSpaces"" = true
     "
 }
-}
 
-hook -group lsp-filetype-json global User LSPDefaultConfig=json %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-json global BufSetOption filetype=json %{
     set-option buffer lsp_servers "
         [vscode-json-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" package.json $(: kak_buffile)}""
         args = [""--stdio""]
     "
 }
-}
 
-hook -group lsp-filetype-julia global User LSPDefaultConfig=julia %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-julia global BufSetOption filetype=julia %{
     set-option buffer lsp_servers "
         # Requires Julia package ""LanguageServer""
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" Project.toml .git .hg $(: kak_buffile)}""
@@ -274,10 +241,8 @@ hook -group lsp-filetype-julia global User LSPDefaultConfig=julia %{
         # ""julia.lint.run"" = ""true""
     "
 }
-}
 
-hook -group lsp-filetype-latex global User LSPDefaultConfig=latex %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-latex global BufSetOption filetype=latex %{
     set-option buffer lsp_servers "
         [texlab]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .git .hg $(: kak_buffile)}""
@@ -308,10 +273,8 @@ hook -group lsp-filetype-latex global User LSPDefaultConfig=latex %{
         }
     "
 }
-}
 
-hook -group lsp-filetype-lua global User LSPDefaultConfig=lua %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-lua global BufSetOption filetype=lua %{
     set-option buffer lsp_servers "
         [lua-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .git .hg $(: kak_buffile)}""
@@ -321,10 +284,8 @@ hook -group lsp-filetype-lua global User LSPDefaultConfig=lua %{
         # diagnostics.enable = true
     "
 }
-}
 
-hook -group lsp-filetype-markdown global User LSPDefaultConfig=markdown %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-markdown global BufSetOption filetype=markdown %{
     set-option -add buffer lsp_servers "
         [marksman]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .marksman.toml $(: kak_buffile)}""
@@ -336,28 +297,22 @@ hook -group lsp-filetype-markdown global User LSPDefaultConfig=markdown %{
     #     args = [""lsp""]
     # "
 }
-}
 
-hook -group lsp-filetype-nim global User LSPDefaultConfig=nim %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-nim global BufSetOption filetype=nim %{
     set-option buffer lsp_servers "
         [nimlsp]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" '*.nimble' .git .hg $(: kak_buffile)}""
     "
 }
-}
 
-hook -group lsp-filetype-nix global User LSPDefaultConfig=nix %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-nix global BufSetOption filetype=nix %{
     set-option buffer lsp_servers "
         [nil]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" flake.nix shell.nix .git .hg $(: kak_buffile)}""
     "
 }
-}
 
-hook -group lsp-filetype-ocaml global User LSPDefaultConfig=ocaml %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-ocaml global BufSetOption filetype=ocaml %{
     set-option buffer lsp_servers "
         [ocamllsp]
         # Often useful to simply do a `touch dune-workspace` in your project root folder if you have problems with root detection
@@ -367,10 +322,8 @@ hook -group lsp-filetype-ocaml global User LSPDefaultConfig=ocaml %{
         # codelens.enable = false
     "
 }
-}
 
-hook -group lsp-filetype-php global User LSPDefaultConfig=php %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-php global BufSetOption filetype=php %{
     set-option buffer lsp_servers "
         [intelephense]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .htaccess composer.json $(: kak_buffile)}""
@@ -380,29 +333,23 @@ hook -group lsp-filetype-php global User LSPDefaultConfig=php %{
         storagePath = ""/tmp/intelephense""
     "
 }
-}
 
-hook -group lsp-filetype-protobuf global User LSPDefaultConfig=protobuf %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-protobuf global BufSetOption filetype=protobuf %{
     set-option buffer lsp_servers "
         [pls] # https://github.com/lasorda/protobuf-language-server
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .git .hg $(: kak_buffile)}""
     "
 }
-}
 
-hook -group lsp-filetype-purescript global User LSPDefaultConfig=purescript %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-purescript global BufSetOption filetype=purescript %{
     set-option buffer lsp_servers "
         [purescript-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" spago.dhall spago.yaml package.json .git .hg $(: kak_buffile)}""
         args = [""--stdio""]
     "
 }
-}
 
-hook -group lsp-filetype-python global User LSPDefaultConfig=python %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-python global BufSetOption filetype=python %{
     set-option -add buffer lsp_servers "
         [pylsp]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" requirements.txt setup.py pyproject.toml .git .hg $(: kak_buffile)}""
@@ -426,10 +373,8 @@ hook -group lsp-filetype-python global User LSPDefaultConfig=python %{
     #     fixAll = true
     # "
 }
-}
 
-hook -group lsp-filetype-r global User LSPDefaultConfig=r %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-r global BufSetOption filetype=r %{
     set-option buffer lsp_servers "
         [r-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" DESCRIPTION .git .hg $(: kak_buffile)}""
@@ -437,10 +382,8 @@ hook -group lsp-filetype-r global User LSPDefaultConfig=r %{
         args = [""--slave"", ""-e"", ""languageserver::run()""]
     "
 }
-}
 
-hook -group lsp-filetype-racket global User LSPDefaultConfig=racket %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-racket global BufSetOption filetype=racket %{
     set-option buffer lsp_servers "
         [racket-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" info.rkt $(: kak_buffile)}""
@@ -448,19 +391,15 @@ hook -group lsp-filetype-racket global User LSPDefaultConfig=racket %{
         args = [""-l"", ""racket-langserver""]
     "
 }
-}
 
-hook -group lsp-filetype-reason global User LSPDefaultConfig=reason %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-reason global BufSetOption filetype=reason %{
     set-option buffer lsp_servers "
         [ocamllsp]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" package.json Makefile .git .hg $(: kak_buffile)}""
     "
 }
-}
 
-hook -group lsp-filetype-rust global User LSPDefaultConfig=rust %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-rust global BufSetOption filetype=rust %{
     set-option buffer lsp_servers "
         [rust-analyzer]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" Cargo.toml $(: kak_buffile)}""
@@ -484,10 +423,8 @@ hook -group lsp-filetype-rust global User LSPDefaultConfig=rust %{
         check.command = ""clippy""
     "
 }
-}
 
-hook -group lsp-filetype-ruby global User LSPDefaultConfig=ruby %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-ruby global BufSetOption filetype=ruby %{
     set-option buffer lsp_servers "
         [solargraph]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" Gemfile $(: kak_buffile)}""
@@ -498,11 +435,9 @@ hook -group lsp-filetype-ruby global User LSPDefaultConfig=ruby %{
         # diagnostics = false
     "
 }
-}
 
 # See https://scalameta.org/metals/docs/integrations/new-editor
-hook -group lsp-filetype-scala global User LSPDefaultConfig=scala %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-scala global BufSetOption filetype=scala %{
     set-option buffer lsp_servers "
         [metals]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" build.sbt .scala-build $(: kak_buffile)}""
@@ -520,30 +455,24 @@ hook -group lsp-filetype-scala global User LSPDefaultConfig=scala %{
         inlayHints.typeParameters.enable = true
     "
 }
-}
 
-hook -group lsp-filetype-sh global User LSPDefaultConfig=sh %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-sh global BufSetOption filetype=sh %{
     set-option buffer lsp_servers "
         [bash-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .git .hg $(: kak_buffile)}""
         args = [""start""]
     "
 }
-}
 
-hook -group lsp-filetype-svelte global User LSPDefaultConfig=svelte %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-svelte global BufSetOption filetype=svelte %{
     set-option buffer lsp_servers "
         [svelteserver]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" package.json tsconfig.json jsconfig.json .git .hg $(: kak_buffile)}""
         args = [""--stdio""]
     "
 }
-}
 
-hook -group lsp-filetype-terraform global User LSPDefaultConfig=terraform %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-terraform global BufSetOption filetype=terraform %{
     set-option buffer lsp_servers "
         [terraform-ls]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" '*.tf' $(: kak_buffile)}""
@@ -553,20 +482,16 @@ hook -group lsp-filetype-terraform global User LSPDefaultConfig=terraform %{
         # rootModulePaths = []
     "
 }
-}
 
-hook -group lsp-filetype-toml global User LSPDefaultConfig=toml %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-toml global BufSetOption filetype=toml %{
     set-option buffer lsp_servers "
         [taplo]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .git .hg $(: kak_buffile)}""
         args = [""lsp"", ""stdio""]
     "
 }
-}
 
-hook -group lsp-filetype-yaml global User LSPDefaultConfig=yaml %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-yaml global BufSetOption filetype=yaml %{
     set-option buffer lsp_servers "
         [yaml-language-server]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" .git .hg $(: kak_buffile)}""
@@ -578,16 +503,13 @@ hook -group lsp-filetype-yaml global User LSPDefaultConfig=yaml %{
         # format.enable = true
     "
 }
-}
 
-hook -group lsp-filetype-zig global User LSPDefaultConfig=zig %{
-    lsp-if-no-servers %{
+hook -group lsp-filetype-zig global BufSetOption filetype=zig %{
     set-option buffer lsp_servers "
         [zls]
         root = ""%sh{eval ""$kak_opt_lsp_find_root"" build.zig $(: kak_buffile)}""
     "
 }
-}
 
 ### Language ID ###
 
@@ -596,22 +518,24 @@ declare-option -docstring %{LSP languageId, usually same as filetype option
 See https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentItem
 } str lsp_language_id
 
-hook -group lsp global User LSPDefaultConfig=(.*) %{
+hook -group lsp global BufSetOption filetype=(.*) %{
     set-option buffer lsp_language_id %val{hook_param_capture_1}
 }
 
-hook -group lsp global User LSPDefaultConfig=(?:c|cpp) %{
+hook -group lsp global BufSetOption filetype=(?:c|cpp) %{
     set-option buffer lsp_language_id c_cpp
 }
-hook -group lsp global User LSPDefaultConfig=javascript %{
+hook -group lsp global BufSetOption filetype=javascript %{
     set-option buffer lsp_language_id javascriptreact
 }
-hook -group lsp global User LSPDefaultConfig=protobuf %{
+hook -group lsp global BufSetOption filetype=protobuf %{
     set-option buffer lsp_language_id proto
 }
-hook -group lsp global User LSPDefaultConfig=sh %{
+hook -group lsp global BufSetOption filetype=sh %{
     set-option buffer lsp_language_id shellscript
 }
-hook -group lsp global User LSPDefaultConfig=typescript %{
+hook -group lsp global BufSetOption filetype=typescript %{
     set-option buffer lsp_language_id typescriptreact
 }
+
+}
```

topisani:
I think all of the proposed options are totally workable, and suggest we just pick one and move on for now. I would then like to look at writing some kakscript to more declaratively declare servers, and then be able to do something like the following to inject a single server, and change some of its default options:
```kakscript
hook global BufSetOption filetype=c %{
  lsp-server-use clangd
  lsp-server-set-config clangd.includeStyle llvm
  lsp-find-root .clangd compile_commands.json .git
}
```
But this would be a QOL layer on top of the "low level" options defined here. Point is, i think this interface is best kept simple for now (like with what this PR currently proposes), and then we can implement more fancy stuff on top of it later, if it turns out to be needed. With this current system, any user specific issues should be solveable using things like `disabled_hooks` and `rmhook`

krobelus:
Yeah, I don't think the hook ordering issue matters that much.
I'll pick one that has a nicer-looking default config, but I'm happy to revisit in future.

I agree that the next thing is to explore ways to reuse (default) server definitions, override parts thereof, and configure roots in a way that's less ugly (but for roots I'd like it if the new thing obsoleted `KAK_LSP_PROJECT_ROOT_RUST` which was added because for some project setups the root patterns are not enough.).

I don't expect that the core `lsp_servers` format needs changing; the simple filetype hooks should keep working.
But we definitely want to add some nice porcelain

vbauerster:
> I forgot to mention that the goal of this change is to make the default config copy-paste friendly.

One way to achieve this is to provide command line option to list default config for specific language. For example:
```
$ kak-lsp --server-config=go
[gopls]
...
```
We can list just config string or ready to use kakoune script either `provide-module` or `hook` implementation which user can copy paste. I incline to believe that `provide-module` is better approach because of ordering guarantee, but I'm ok with `hook` implementation. Either case user has to copy paste if one needs custom config.

krobelus:
I think the a `provide-module` approach would be nice but it seems like a lossy abstraction.
If the user wants to use two different sets of servers for two files of the same filetype (but different project or similar) then it wouldn't work. For that case, the user needs to be in control of the hooks.

It's a judgement call how important these cases are; I think for now it's fine to have multiple competing styles; all one needs to do is replace `rc/servers.kak`.

topisani:
> all one needs to do is replace rc/servers.kak.
could we maybe get an option to `kak-lsp --kakoune --no-default-servers`, or another way to easily opt out of _all_ default language configs? In practice, just not emiting rc/servers.kak

krobelus:
> could we maybe get an option to kak-lsp --kakoune --no-default-servers, or another way to easily opt out of all default language configs? In practice, just not emiting rc/servers.kak

We offer two ways to load `rc/*.kak`:
1 `eval %sh{kak-lsp --kakoune}`
2. clone into autoload / load via plugin manager / source manually

It's kind of unfortunate that we have both.
Option 1 can be more convenient for quick installation, trying out different versions, avoiding user errrors.

Option 2 is more conventional for Kakoune plugins.
Additionally, it allows to edit `rc/lsp.kak`, make a quick change and restart without recompiling (I use this a lot)

Given that there is no best approach, I'm not sure if we want to add something like `kak-lsp --no-default-servers` that only works in one of them.
It seems better to use `remove-hooks global lsp-filetype-.*` (or `disabled_hooks`) or source files manually; users should learn those tricks anyway, for general kakoune usage.

I guess it used to be easier to remove all default servers, so maybe you're right.

krobelus:
The packaging question is interesting but it can be a rabbit hole.

If we could hide the compilation step and the whole binary from the user, that would be nice
because cloning to ~/.config/kak/autoload/ would just work.
That's how it works for plugins that use a scripting language instead of a compiled like Rust.

I wonder if that would work for us.

Packages would still install a `kak-lsp` binary, so we wouldn't need to compile but merely hardcode the absolute path to `kak-lsp` in `lsp.kak`.
This would imply that packagers would add something like `/usr/share/kak/autoload/lsp/{lsp,servers}.kak`.
Note this means that only `/bin/kak` will have LSP support but `src/kak` will not (because it has a different autoload directory).
The obvious way to make it work for `src/kak` is to clone `kak-lsp` separately into their `~/.config/autoload` or into `./rc`.

In order to override a kakoune-lsp installed in this way, one would need to uninstall it, or fine-tune `~/.config/autoload` directory, which seems fair I guess.

Non-packaged installs would install it to `target/release` as needed.

I don't see a good way to tell if we should recompile. Maybe never, or if `rc/*.kak` don't match `kak-lsp --kakoune` ...

Not sure yet. It would definitely be a breaking change.
But it seems to make more sense for a Kakoune plugin to live in autoload than in $PATH (the package would still install the binary but only for running the kak-lsp server from outside Kakoune).

---

Else `kak-lsp` will remain our canonical interface.

I agree we should try to add a more obvious way to disable default servers,
even if docs already state how one can do it.

krobelus:
If we use autoload, then we can also make `:doc lsp` work.

> The obvious way to make it work for src/kak is to clone kak-lsp separately into their ~/.config/autoload or into ./rc.

alternatively, `ls -s /usr/share/kak/rc/kakoune-lsp path/to/kakoune/rc/kakoune-lsp`.

Quite hard to decide between convention versus convenience (+ avoiding breakage).

PR #783, opened 2024-10-16
Match all settings tables in servers.kak

arrufat:
Follow up of #777


rc/servers.kak
```diff
@@ -151,7 +151,7 @@ hook -group lsp-filetype-html global BufSetOption filetype=html %{
         root_globs = ["package.json"]
         args = ["--stdio"]
         settings_section = "_"
-        [html-language-server.settings._]
+        [vscode-html-language-server.settings._]
```
krobelus, line 154: huh so we're broken for html. I should really have double-checked these. That'll be another patch release then
Thanks!

PR #791, opened 2024-10-28
fix two types in julia lsp configuration

rowanxshi:
1. `"true"` to `true`
2. move `root_globs` option under `[julia-language-server]` heading

### DIFF
--- a/rc/servers.kak
+++ b/rc/servers.kak
@@ -222,10 +222,10 @@ hook -group lsp-filetype-json global BufSetOption filetype=json %{
 hook -group lsp-filetype-julia global BufSetOption filetype=julia %{
     set-option buffer lsp_servers %{
         # Requires Julia package "LanguageServer"
-        root_globs = ["Project.toml", ".git", ".hg"]
         # Run: `julia --project=@kak-lsp -e 'import Pkg; Pkg.add("LanguageServer")'` to install it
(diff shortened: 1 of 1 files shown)


rc/servers.kak
```diff
@@ -222,10 +222,10 @@ hook -group lsp-filetype-json global BufSetOption filetype=json %{
 hook -group lsp-filetype-julia global BufSetOption filetype=julia %{
     set-option buffer lsp_servers %{
         # Requires Julia package "LanguageServer"
-        root_globs = ["Project.toml", ".git", ".hg"]
         # Run: `julia --project=@kak-lsp -e 'import Pkg; Pkg.add("LanguageServer")'` to install it
         # Configuration adapted from https://github.com/neovim/nvim-lspconfig/blob/bcebfac7429cd8234960197dca8de1767f3ef5d3/lua/lspconfig/julials.lua
         [julia-language-server]
+        root_globs = ["Project.toml", ".git", ".hg"]
```
krobelus, line 228: Thanks!
FWIW, now that we can also specifiy `root` directly (as alternative to `root_globs`),
we can reintroduce a smarter root detection logic without undermining kakoune-lsp's concept of the project root.
It was removed in ebd370f4 (Remove Julia root detection hacks, 2024-08-09)

PR #793, opened 2024-10-30
WIP: allow users to override `SymbolKind` rendering

igor-ramazanov:
Adds a new optional `[<server_name>.symbols_mapping]` nested table to the `lsp_servers` which is a mapping from `SymbolKind`s to any user-defined strings to make the `lsp-document-symbol` and `lsp-goto-document-symbol` more readable.

More things needed:
* [x] Implement for `lsp-document-symbol`
* [ ] Implement for `lsp-goto-document-symbol`
* [ ] Examples in `rc/servers.kak`
* [ ] Documentation
* [ ] Perhaps, some refactoring

With the new `[rust.symbols_mapping]`:
```
  hook -group lsp-custom-servers global BufSetOption filetype=rust %{
      [rust]
      filetypes = ["rust"]
      root_globs = ["Cargo.toml"]
      command = "rust-analyzer"
      [rust.symbols_mapping]
      Constant = "const"
      Enum = "enum"
      EnumMember = ""
      Field = ""
      Function = "fn"
      Interface = "trait"
      Method = "fn"
      Module = "mod"
      Object = ""
      Struct = "struct"
      TypeParameter = "type"
    }
  }
```

Before:
```
src/controller.rs
%:46:6:    QuoteState (Enum)
%:47:5:    ├── OutsideArg (EnumMember)
%:48:5:    ├── InsideArg (EnumMember)
%:49:5:    └── InsideArgSQ (EnumMember)
%:52:12:   ParserState (Struct)
%:53:5:    ├── session (Field)
%:54:9:    ├── buf (Field)
%:55:5:    ├── offset (Field)
%:56:5:    └── state (Field)
%:59:6:    impl ParserState (Object)
%:60:12:   └── new (Function)
%:70:4:    next_string (Function)
%:85:4:    process (Function)
%:113:7:   FromString (Interface)
%:114:10:  ├── Err (TypeParameter)
%:115:8:   └── from_string (Function)
...
```

After:
```
src/controller.rs
%:46:6:    enum QuoteState
%:47:5:    ├── OutsideArg
%:48:5:    ├── InsideArg
%:49:5:    └── InsideArgSQ
%:52:12:   struct ParserState
%:53:5:    ├── session
%:54:9:    ├── buf
%:55:5:    ├── offset
%:56:5:    └── state
%:59:6:    impl ParserState
%:60:12:   └── fn new
%:70:4:    fn next_string
%:85:4:    fn process
%:113:7:   trait FromString
%:114:10:  ├── type Err
%:115:8:   └── fn from_string
...
```

### DIFF
--- a/src/language_features/document_symbol.rs
+++ b/src/language_features/document_symbol.rs
@@ -244,6 +244,7 @@ pub fn format_symbol<T: Symbol<T>>(
         ctx: &Context,
         single_file: bool,
         prefix: &mut Vec<bool>,
+        symbols_mapping: &HashMap<String, String>,
     ) {
         let length = items.len();
         let is_root = prefix.is_empty();
@@ -265,7 +266,20 @@ pub fn format_symbol<T: Symbol<T>>(
                     .join("");
                 format!("{}{}", prefixing_hierarchy_symbols, last_hierarchy_symbol)
             };
-            let description = format!("{}{} ({:?})", hierarchy, symbol.name(), symbol.kind());
+
+            let mapped_symbol = symbols_mapping
+                .get(format!("{:?}", symbol.kind()).as_str())
+                .map_or(
+                    format!("{} ({:?})", symbol.name(), symbol.kind()),
+                    |mapped_kind| {
+                        if mapped_kind.is_empty() {
+                            symbol.name().to_string()
+                        } else {
+                            format!("{} {}", mapped_kind, symbol.name())
+                        }
+                    },
+                );
+            let description = format!("{}{}", hierarchy, mapped_symbol);
             let mut filename_path = PathBuf::default();
             let filename = symbol_filename(meta, symbol, &mut filename_path);
             let position = get_kakoune_position_with_fallback(
@@ -290,7 +304,16 @@ pub fn format_symbol<T: Symbol<T>>(
 
             let children = symbol.children();
             prefix.push(is_last);
-            format_symbol_at_depth(output, children, meta, server, ctx, single_file, prefix);
+            format_symbol_at_depth(
+                output,
+                children,
+                meta,
+                server,
+                ctx,
+                single_file,
+                prefix,
+                symbols_mapping,
+            );
             prefix.pop();
         }
     }
@@ -303,6 +326,12 @@ pub fn format_symbol<T: Symbol<T>>(
         ctx,
         single_file,
         &mut vec![],
+        &meta
+            .language_server
+            .get(&server.name)
+            .map_or(HashMap::new(), |language_server_config| {
+                language_server_config.symbols_mapping.clone()
+            }),
     );
     if single_file {
         // Align symbol names (first column is %:line:col).
--- a/src/types.rs
+++ b/src/types.rs
@@ -82,6 +82,8 @@ pub struct LanguageServerConfig {
     pub settings_section: Option<String>,
     pub settings: Option<Value>,
     pub offset_encoding: Option<OffsetEncoding>,
+    #[serde(default)]
+    pub symbols_mapping: HashMap<String, String>,
     pub experimental: Option<Value>,
     // This does nothing, but is kept so we can still parse old configs.
     #[allow(dead_code)]


src/types.rs
```diff
@@ -82,6 +82,8 @@ pub struct LanguageServerConfig {
     pub settings_section: Option<String>,
     pub settings: Option<Value>,
     pub offset_encoding: Option<OffsetEncoding>,
+    #[serde(default)]
+    pub symbols_mapping: HashMap<String, String>,
```
igor-ramazanov, line 86:
```suggestion
    pub symbols_mapping: HashMap<SymbolKind, String>,
```
Would be nice, but not sure how.

src/language_features/document_symbol.rs
```diff
@@ -265,7 +266,20 @@ pub fn format_symbol<T: Symbol<T>>(
                     .join("");
                 format!("{}{}", prefixing_hierarchy_symbols, last_hierarchy_symbol)
             };
-            let description = format!("{}{} ({:?})", hierarchy, symbol.name(), symbol.kind());
+
+            let mapped_symbol = symbols_mapping
+                .get(format!("{:?}", symbol.kind()).as_str())
```
igor-ramazanov, line 271: Seems like an ugly way to convert `SymbolKind` to string.

src/language_features/document_symbol.rs
```diff
@@ -265,7 +266,20 @@ pub fn format_symbol<T: Symbol<T>>(
...
+                .map_or(
+                    format!("{} ({:?})", symbol.name(), symbol.kind()),
```
igor-ramazanov, line 273: The existing behavior.

src/language_features/document_symbol.rs
```diff
@@ -265,7 +266,20 @@ pub fn format_symbol<T: Symbol<T>>(
...
+                    |mapped_kind| {
+                        if mapped_kind.is_empty() {
+                            symbol.name().to_string()
```
igor-ramazanov, line 276: If the user-provided string is empty, then maybe that's an intention.

src/language_features/document_symbol.rs
```diff
@@ -265,7 +266,20 @@ pub fn format_symbol<T: Symbol<T>>(
...
+                        } else {
+                            format!("{} {}", mapped_kind, symbol.name())
```
igor-ramazanov, line 278: A new case: prepend a user-string.

src/types.rs
krobelus: we'd need to write an

	impl<'de> Deserialize<'de> for SymbolKind

should be fairly simple but I think we might wanna try a different UI

src/language_features/document_symbol.rs
krobelus: > although some users may want to append the string, instead of prepending.

right.

I'm not sure we want to build a declarative config language for this,
because it will be limited.

We don't really need the full power of rust here;
we could do these text replacements in kakoune script.

I think we can achieve the same thing by overriding
`lsp-show-document-symbol`

Granted, that doesn't allow using different configurations multiple servers
in the same filetype like `lsp_servers` does.  But this issue is present
in a lot of our configuration switches.. multiple servers in LSP are hacky
and I don't expect much need for different per-server configuration here.
I think in future we should try to find a solution for this that doesn't
require stuffing everything into this weird `lsp_servers` TOML table in.

So I think we should try to find a way to customize `lsp-show-document-symbol`
per filetype (or per server) in a way that allows the user to run arbitrary
code.

perhaps something like (untested)

```kakrc
hook global BufCreate \*goto\* %{
	hook -once BufSetOption filetype=lsp-document-symbol %{
		hook -once NormalIdle .* %{
			execute-keys -draft %{%s \(Enum\)$<ret>d<a-b>ienum <esc>}
		}
	}
}
```

I didn't try this so I don't have a strong opinion yet.  I can see that
there are also upsides to the declarative approach: the UI boundary
is clearer and it's easier to not accidentally break it when changing
some internals.

Actually we should do this to get rid of one nested hook above

```diff
diff --git a/rc/lsp.kak b/rc/lsp.kak
index 5d38882be2..04236f8763 100644
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -1225,7 +1225,7 @@
 }
 
 define-command -hidden lsp-show-document-symbol -params 3 -docstring "Render document symbols" %{
-    lsp-show-goto-buffer *goto* lsp-document-symbol %arg{1} %arg{3}
+    lsp-show-goto-buffer *symbols* lsp-document-symbol %arg{1} %arg{3}
     evaluate-commands -try-client %opt[toolsclient] %{
         set-option -add buffer path %arg{1} # for gf on the file name
         set-option buffer lsp_buffile %arg{2}
```

igor-ramazanov: I like this approach! I'll try to hack around the KakScript and see if it's possible to implement everything in terms of Kakoune UI only.

krobelus: One approach is to let the user (or default config) define a buffer local alias, but having to find the name to alias is extra burden. Matching against the buffer name is more obvious. Since the buffer is not a fifo buffer, it's not super obvious when the buffer will be populated (it won't be populated yet in `BufCreate`).
Maybe we should make it a fifo buffer for that reason, then we can use `BufCloseFifo`..
Either way, I think it's not tragic if it's not super obvious since the main way users learn about this is by copying the default Rust config.

krobelus:
Ah I totally missed that we want to use it elsewhere. Maybe not `lsp-goto-document-symbol` but totally in `lsp-breadcrumbs`.

So I'm good with the approach. Happy to merge as-is; I'll try adding it to breadcrumbs

PR #794, opened 2024-10-30
Fix crashing on `window/showMessageRequest`

igor-ramazanov:
It's been crashing for me with the `state.next` returning quoted numbers like `"3"` which then unsuccessfully parsed as `toml::Value`.

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -503,9 +503,9 @@ fn dispatch_fifo_request(
         "textDocument/switchSourceHeader" => Box::new(()),
         "window/showMessageRequest/showNext" => Box::new(()),
         "window/showMessageRequest/respond" => {
-            let id: toml::Value = toml::from_str(&state.next::<String>()).unwrap();
+            let id = jsonrpc_core::Id::Str(state.next::<String>());
             Box::new(MessageRequestResponse {
-                message_request_id: jsonrpc_core::Id::deserialize(id).unwrap(),
+                message_request_id: id,
                 item: state
                     .next::<Option<String>>()
                     .map(|s| toml::from_str(&s).unwrap()),


src/controller.rs
```diff
@@ -503,9 +503,9 @@ fn dispatch_fifo_request(
         "textDocument/switchSourceHeader" => Box::new(()),
         "window/showMessageRequest/showNext" => Box::new(()),
         "window/showMessageRequest/respond" => {
-            let id: toml::Value = toml::from_str(&state.next::<String>()).unwrap();
+            let id = jsonrpc_core::Id::Str(state.next::<String>());
             Box::new(MessageRequestResponse {
-                message_request_id: jsonrpc_core::Id::deserialize(id).unwrap(),
+                message_request_id: id,
```
krobelus, line 508: Thanks and sorry about that; we're missing a lot of test coverage.
Fortunately it's easy to test this manually with `metals` on an empty `*.scala` file.
I'll push a slightly different fix since this one might not work if the server treats `id: "123"` differently from `id: 123`.

Something else I noticed is that I see a glitch when multiple messages are sent; it looks like

```
LSP: info from server metals: ⚠️ Compiled t_f4dd477a3a (0.12s)      🔔1 t/test.scala 1:4  1 sel - client0@[foo]
LSP: info from server metals: ⚠️ Compiled t_f4dd477a3a (0.12s)      🔔1 t/test.scala 1:4  1 sel - client0@[foo]
```

but when I hold shift and copy those two lines, the text is actually joined in one line, and the trailing `]`

```
LSP: info from server metals: ⚠️ Compiled t_f4dd477a3a (0.12s)      🔔1 t/test.scala 1:4  1 sel - client0@[fooLSP: info from server metals: ⚠️ Compiled t_f4dd477a3a (0.12s)                                                                                      ⌛🔔1 t/test.scala 1:4  1 sel - client0@[foo
```

it looks like Kakoune and my terminal disagree on the width of 🔔 or ⚠️ (probably the latter since that seems to includes a variation selector).
I'll look into it..

igor-ramazanov: Thanks!

igor-ramazanov: >Something else I noticed is that I see a glitch when multiple messages are sent; it looks like

Hm, yeah, I remember I had the same problem in the past which I fixed [by changing this line](https://github.com/kakoune-lsp/kakoune-lsp/blob/ce7077cec4ac8e3f017467c97c5a8c12fb6be235/rc/servers.kak#L469) to `statusBarProvider = "log-message"` ([docs](https://scalameta.org/metals/docs/integrations/new-editor#statusbarprovider)).

However, recently I've changed it back to `statusBarProvider = "show-message"` and have no issues anymore.

I suspect, it's a bug with an old `metals` version. Mine is `1.4.0` and I can not reproduce it.

krobelus: it's a well-known type of bug: the terminal and Kakoune disagree on the width of `⚠️`.
In my case, the terminal (foot, which has a custom implementation of `wcwidth()` says it's width 2 and Kakoune (which uses the glibc one) thinks it's width 1.

FWIW this is how I tested the glibc version:

```c++
#define _XOPEN_SOURCE
#include <locale.h>
#include <stdio.h>
#include <wchar.h>

int main() {
  setlocale(LC_ALL, "");
#define TEST(s) wprintf(s " %d\n", wcswidth(s, wcslen(s)))
  TEST(L"🔎");
  TEST(L"💡");
  TEST(L"⌛");
  TEST(L"⚠"); // 1
  TEST(L"⚠️"); // 1 but should be 2
}
```

Unfortunately I keep getting network timeouts on the glibc bug tracker.

See also [this comment](https://github.com/kovidgoyal/kitty/issues/1978#issuecomment-575430679) for example.
This situation is pretty unfortunate, not sure when it will get fixed. In kakoune-lsp we generally work around this by falling back to ascii chars if the system wcwidth is confirmed to be wrong.

> Hm, yeah, I remember I had the same problem in the past which I fixed [by changing this line](https://github.com/kakoune-lsp/kakoune-lsp/blob/ce7077cec4ac8e3f017467c97c5a8c12fb6be235/rc/servers.kak#L469) to statusBarProvider = "log-message" ([docs](https://scalameta.org/metals/docs/integrations/new-editor#statusbarprovider)).

Nice, that sounds like the thing to do as default, until we can expect this issue to be fixed in the user's libc.
Alternatively we could apply our `expected_width_or_fallback` trick but that requires a Unicode database (today we just use it for 3 hard-code characters).

krobelus: reported glibc bug as https://sourceware.org/bugzilla/show_bug.cgi?id=32322 though I wonder if it's better to use a separate library for that

PR #796, opened 2024-11-03
add default configuration for tinymist

thoughtpatterns:
includes an expansion to set the formatter column width. preview is disabled due to being experimental.

### DIFF
--- a/rc/servers.kak
+++ b/rc/servers.kak
@@ -511,6 +511,21 @@ hook -group lsp-filetype-toml global BufSetOption filetype=toml %{
     }
 }
 
+hook -group lsp-filetype-typst global BufSetOption filetype=typst %{
+    set-option buffer lsp_servers %exp{
+        [tinymist]
+        root_globs = [".git", ".hg"]
+        args = ["lsp"]
+        settings_section = "tinymist"
+        [tinymist.settings.tinymist]
+        # See https://myriad-dreamin.github.io/tinymist/configurations.html
+        exportPdf = "onDocumentHasTitle"
+        formatterMode = "typstyle"
+        formatterPrintWidth = %opt{autowrap_column}
+        previewFeature = "disable"
+    }
+}
+
 hook -group lsp-filetype-yaml global BufSetOption filetype=yaml %{
     set-option buffer lsp_servers %{
         [yaml-language-server]
--- a/rc/servers.kak
+++ b/rc/servers.kak
@@ -512,18 +512,18 @@ hook -group lsp-filetype-toml global BufSetOption filetype=toml %{
 }
 
 hook -group lsp-filetype-typst global BufSetOption filetype=typst %{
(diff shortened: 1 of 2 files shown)


rc/servers.kak
```diff
@@ -511,6 +511,21 @@ hook -group lsp-filetype-toml global BufSetOption filetype=toml %{
     }
 }
 
+hook -group lsp-filetype-typst global BufSetOption filetype=typst %{
+    set-option buffer lsp_servers %exp{
```
krobelus, line 515: thanks! I think we should use `""` for now instead of `%exp{}` because we still support Kakoune v2021.

rc/servers.kak
```diff
@@ -511,6 +511,21 @@ hook -group lsp-filetype-toml global BufSetOption filetype=toml %{
...
+        [tinymist]
+        root_globs = [".git", ".hg"]
+        args = ["lsp"]
+        settings_section = "tinymist"
+        [tinymist.settings.tinymist]
```
krobelus, line 520: I think for consistency with others this should be `[tinymist.settings._]` (and `settings_section = "_"`), unless the server really requests the `tinymist` section via `workspace/configuration`, which I doubt because the vscode config file has no such section but only keys like `"tinymist.foo": ...`.

Put another way, the `tinymist` language server probably has no idea that the section is called `tinymist`.
The VS Code plugin likely muddies the waters by collecting all `"tinymist.foo"` keys into one object for `initializeOptions` and `didChangeConfiguration`.
Other editors need to model that ugliness; I think to reduce confusion we should make it explicit that we don't need the `tinymist` name for that.

rc/servers.kak
```diff
@@ -511,6 +511,21 @@ hook -group lsp-filetype-toml global BufSetOption filetype=toml %{
...
+        # See https://myriad-dreamin.github.io/tinymist/configurations.html
+        exportPdf = "onDocumentHasTitle"
+        formatterMode = "typstyle"
+        formatterPrintWidth = %opt{autowrap_column}
```
krobelus, line 524: we try to avoid hard dependencies on other plugins (even stdlib ones);
to avoid breakage if someone doesn't have `autowrap.kak` in their autoload, can you re-declare this option after this line:

`declare-option -docstring 'name of the client in which documentation is to be displayed' str docsclient`

rc/servers.kak
thoughtpatterns: sure! would i just replace the %exp{ } with " ", or should i escape what's inside also?

krobelus: One way is to replace, and double up any `"` inside (see also `;doc expansions`). It's very ugly unfortunately.
So how about a different approach:

```
set-option buffer lsp_servers %{ ... }
set-option -add buffer lsp_servers "formatterPrintWidth = %opt{autowrap_column}"
```

if that's not too confusing..

PR #804, opened 2024-11-23
Add lsp-filetype-sql and sqls language-server

jtrv:


rc/servers.kak
```diff
@@ -553,6 +553,13 @@ hook -group lsp-filetype-sh global BufSetOption filetype=sh %{
     }
 }
 
+hook -group lsp-filetype-sql global BufSetOption filetype=sql %{
+    set-option buffer lsp_servers %{
+        [sqls]
+        root_globs = [".git", ".hg"]
```
krobelus, line 559: I haven't tried this one much but I think it did send invalid JSON (which crashed kak-lsp, I guess we should fix that).
I think this was becaus emy file contained invalid SQL, I'll try to report a bug eventually.
Also clippy showed an error telling me that I wasn't connected. (I use https://gitlab.com/Screwtapello/kakoune-repl-buffer for executing queries). Might want to sort that out too before enabling this by default. Though I guess if the user has explicitly installed `sqls` that might be a strong enough signal that they want to use it more often than not? I'm not sure

PR #806, opened 2024-11-23
Add lsp-filetype-prisma and prisma-language-server

jtrv:


krobelus:
I didn't really know what prisma is but I suppose if people want to use that we can add it.

The fish one (https://github.com/kakoune-lsp/kakoune-lsp/pull/809) is probably fine too, I forgot about that one
I can merge them later this week.

PR #807, opened 2024-11-23
Add markdown-oxide language-server

jtrv:


krobelus:
looks like Helix enables this one by default (along marksman).

Unfortunately we have a bug where, if one language server is not installed, we disable all language servers for the current buffer (`lsp-block-in-buffer`)
We should fix that, then we can enable multiple servers by default (though that still is kind of hacky)

PR #809, opened 2024-11-23
Add lsp-filetype-fish and fish-lsp language server

jtrv:


krobelus:
Interesting.. I'd have implemented this as part of fish itself.
But this is using tree-sitter, which seems to work great for most cases.

I think it's important that this one doesn't make everyday use cases[^1] worse;
I briefly tried it and there are a couple issues I noticed:
1. `set g lsp_auto_highlight_references true` will highlight entire blocks which is pretty loud since we use a bold font here (should we use italics instead?). Maybe that should be changed to only ever highlight the identifier at cursor
2. `lsp-hover` on a command like `sed` shows a dead link to https://fishshell.com/docs/current/cmds/sed.html
3. There are some false positive warnings, for example on `sh -c 'echo $HOME'` (or `eval 'echo $HOME'`).
   I understand that these warnings help people in many cases but I'm not yet sure how I feel about this.
   fish itself doesn't really have warnings, only errors.
4. false positive warning: `Conditional command should include a silence option` on things like
  `if set -l fishdir (status fish-path | string match -vr /bin/)` 

I'm not yet sure what should be done about this (cc @ndonfris, sorry for not creating proper bug reports yet)

[^1]: I'm mainly thinking of `alt-e` in fish or editing config.fish; both should not show a lot of false-positive warnings. I don't write other fish scripts myself, except for completion scripts in the fish repo

ndonfris:
All good! I appreciate the feedback. 

I've had a lot of issues trying to build this project in other ecosystems, and have found the most success with TypeScript and wasm using [web-tree-sitter](https://www.npmjs.com/package/web-tree-sitter).

I am also unsure with how some of the lsp related dependencies would work with fish-shell's current internal tooling but I'm interested to hear alternative ideas. Personally, I think separating the language server from the internal shell language makes the most sense.

---

Now some notes on your issues,

1. It is unclear to me if this is an error with the lsp or kakune. I'm happy to accommodate any requested changes you'd like to see for better integration with kakune. 

2. >lsp-hover on a command like sed shows a dead link to https://fishshell.com/docs/current/cmds/sed.html

    I believe this must be an newly introduced error on one of the more recent versions, as the links to fish-shell documentation are meant to only be generated by the `builtin -n` commands. So this should be a quick fix.

3. Yes this is a known error planning to be patched in the next release. You should be able disable this specific diagnostic with 

```fish
set -gx fish_lsp_diagnostic_disable_error_codes 2001
``` 

> You can disable any specific diagnostic by adding it to the `fish_lsp_diagnostic_disable_error_codes` list. The error codes are listed in the source code [here](https://github.com/ndonfris/fish-lsp/blob/master/src/diagnostics/errorCodes.ts), and you can disable other [features](https://github.com/ndonfris/fish-lsp/tree/master#server-configuration-optional) causing false positives.

4. This should also be a simple fix as the current diagnostic just checks that if you're trying to check if a variable is being set in a command that it has a silenced option. I've seen the syntax you described in the `set` manpage, but wasn't quite sure how to use it. Looks like all that needs to be added is a second check that a subshell doesn't follow the `if set` syntax (in the tree-sitter related code of fish-lsp).

The current warning was for checking if a variable has been set. _i.e._

```fish
if set -q some_var
    # do something with $some_var
end
```

---

As for your footnotes, I opted towards setting the fish-lsp specific options via shell variables for this specific reasoning. This allows you to set whatever options you'd like per specific ways you'd want to use the fish-lsp. In the example mentioned, this would be done by wrapping `edit_command_buffer` in a simple function with options disabled. 

```fish
function edit_fish_lsp_command_buffer
    set -lx fish_lsp_disabled_handlers diagnostics
    set -lx fish_lsp_diagnostic_disable_error_codes 2001 2002 2003 # ....
    edit_command_buffer
end

bind \ee edit_fish_lsp_command_buffer
```

The primary goal of providing warnings/diagnostics in the current version is in hope of adding code-actions support to the language server (which is in the works, currently). Two upcoming warnings will be for unused definitions (variables inside a function, functions that aren't autoloaded), and unreachable code blocks. As you point out, lots of the features I'm aiming to implement are more for the crowd that uses fish for it's scripting capabilities.

Please feel free to keep providing error reports and feedback, as it's tough to test all the edge cases on my own!

jtrv:
> I think it's important that this one doesn't make everyday use cases[1](#user-content-fn-1-eb0088b1fdc4757059668092b97a4295) worse; ...
>     1. I'm mainly thinking of `alt-e` in fish or editing config.fish; both should not show a lot of false-positive warnings. I don't write other fish scripts myself, except for completion scripts in the fish repo [↩](#user-content-fnref-1-eb0088b1fdc4757059668092b97a4295)

I agree, maybe we should opt to comment it so others can still find it for their configs if interested, what do you think?

krobelus:
I think it's fine to enable it by default (the user did install it after all) as long as we deal with these issues (first)

PR #820, opened 2025-01-19
Document Links

topisani:
Barebones implementation for now, there are some low hanging fruit:
 - Add to hover messages
 - Add line flags
 - Add default mapping


rc/lsp.kak
```diff
@@ -978,6 +989,17 @@ define-command -hidden lsp-code-lens-request %{
     }
 }
 
+define-command lsp-document-link -docstring "follow a document link from the current selection" %{
```
krobelus, line 992: `current` -> `main` I guess.

rc/lsp.kak
```diff
@@ -1828,6 +1851,7 @@
     add-highlighter "%arg{1}/lsp_references" ranges lsp_references
     add-highlighter "%arg{1}/lsp_semantic_tokens_ranges" ranges lsp_semantic_tokens_ranges
     add-highlighter "%arg{1}/lsp_snippets_placeholders" ranges lsp_snippets_placeholders
+    add-highlighter "%arg{1}/lsp_document_links" ranges lsp_document_links
```
krobelus, line 1854: should also add the corresponding `remove-highlighter` in `lsp-disable-impl`, for consistency

src/capabilities.rs
```diff
@@ -470,6 +470,7 @@ pub const CAPABILITY_SEMANTIC_TOKENS: &str = "lsp-semantic-tokens";
 pub const CAPABILITY_SIGNATURE_HELP: &str = "lsp-signature-help";
 pub const CAPABILITY_TYPE_DEFINITION: &str = "lsp-type-definition";
 pub const CAPABILITY_WORKSPACE_SYMBOL: &str = "lsp-workspace-symbol";
+pub const CAPABILITY_DOCUMENT_LINKS: &str = "lsp-document-links";
```
krobelus, line 473: this should match the name of the command, so `lsp-document-link` (no s)

Actually we should probably lose the s almost everywhere in this PR
(making it `CAPABILITY_DOCUMENT_LINK`, `mod document_link`)
this should be consistent with most others like `document_symbol`.

rc/lsp.kak
```diff
@@ -105,6 +105,8 @@
 # Face for inlay hints.
 set-face global InlayHint cyan+d
 set-face global InlayCodeLens cyan+d
+# face for document links
+set-face global LspDocumentLink +u
```
krobelus, line 109: would be good to be consistent about the "Lsp" prefix (others like `InlayHint` don't have it).

I don't know much about faces but maybe we should prefer a generic name that can be used outside LSP.

Looking at the one precedent I can think of: doc.kak hardcodes the underline link
highlighting (`doc_render_ranges "%val{selection_desc}|default+u"`).

I'd be happy with hardcoding.

But if we want it configurable, I'd call it `Link`, because that's consistent with our `InfoLink`.

rc/lsp.kak
```diff
@@ -266,6 +268,14 @@
     lsp-menu %arg{@}
 }
 
+define-command -hidden lsp-follow-document-link -params 1.. -docstring "Called on :lsp-document-link" %{
+    lsp-menu %arg{@}
```
krobelus, line 272: Adding override points that are unlikely to ever be used feels somewhat odd. (I can imagine someone might override lsp-menu but not so much the individual lsp-menu users).

Probably it's fine to keep it, because it's consistent with the existing structure.
Not sure yet.

src/language_features/document_links.rs
```diff
@@ -0,0 +1,213 @@
...
+    let eligible_servers: Vec<_> = ctx
+        .servers(&meta)
+        .filter(|srv| attempt_server_capability(ctx, *srv, &meta, CAPABILITY_DOCUMENT_LINKS))
+        .collect();
+    if eligible_servers.is_empty() {
+        return;
+    }
+
+    let (first_server, _) = *eligible_servers.first().unwrap();
+    let first_server = first_server.to_owned();
```
krobelus, line 37: `first_server` is unused

src/language_features/document_links.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    ctx.document_links.insert(buffile.clone(), links);
+
+    let command = format!("set-option buffer lsp_document_links {version} {range_specs}");
+    let command = format!(
+        "evaluate-commands -buffer {} %§{}§",
+        editor_quote(buffile),
+        command.replace('§', "§§")
+    );
+    ctx.exec(EditorMeta::default(), command);
```
krobelus, line 109: looks correct; it's super confusing that `ctx.exec(meta, ...)` only ever uses `meta` for `meta.client`.

src/language_features/document_links.rs
```diff
@@ -0,0 +1,213 @@
...
+        .iter()
+        .filter(|(server_id, link)| {
+            let ServerSettings {
+                offset_encoding, ..
+            } = ctx.server(*server_id);
+            let range = kakoune_range_to_lsp(&range, &document.text, *offset_encoding);
+            ranges_overlap(link.range, range)
+        })
+        .map(|(a, b)| (*a, b.clone()))
+        .collect::<Vec<_>>();
```
krobelus, line 166: it seems odd that the earlier case (where links are unresolved) only follows one link,
and here we follow all of the available (resolved) ones.
Is there a specific use case you have in mind?

I'd expect us to first resolve all unresolved links, and then offer all
(resolved) links, no matter whether we just resolved them, or whether they
were previously resolved.

I did not check whether there is some kind of guarantee by LSP that we can
rely on, but I doubt there is. So we probably want to handle the case where
we have both resolved and unresolved links exist.

(In theory, we could even cache the resolved links in `ctx.document_links`,
so we don't have to resolve them again. But I would not do that, since it
probably adds complexity for no significant gain).

src/language_features/document_links.rs
```diff
@@ -0,0 +1,213 @@
...
+    if let Some((server_id, link)) = ctx
+        .document_links
+        .get(&meta.buffile)
+        .and_then(|links| {
+            links.iter().find(|(server_id, link)| {
+                let ServerSettings {
+                    offset_encoding, ..
+                } = ctx.server(*server_id);
+                let range = kakoune_range_to_lsp(&range, &document.text, *offset_encoding);
+                ranges_overlap(link.range, range)
```
krobelus, line 137: should probably deduplicate this `filter()` call (maybe extract a function if we stick with two separate code paths here)

src/language_features/document_links.rs
```diff
@@ -0,0 +1,213 @@
...
+    let commands = links
+        .iter()
+        .filter(|(_, link)| link.target.is_some())
+        .map(|(_, link)| {
+            let target = link.target.as_ref().unwrap();
+            let label = editor_quote(target.as_str());
+            let path = Some(target)
+                .filter(|x| x.scheme() == "file")
+                .and_then(|x| x.to_file_path().ok())
+                .map(|x| x.to_string_lossy().to_string());
```
krobelus, line 186: super nit:

I guess our error handling is all over the place :(
If we started a project from scratch we'd definitely add an easy way to log
these errors, since they probably indicate a bug in the LSP server.

I guess if `to_string_lossy` is forced to do lossy conversion, we at least
get part of the name, so the user will see a somewhat reasonable error message.
But logging and propagating some error to Kakoune seems like it would be ideal,
because AFAICT, Kakoune will not be able to open the right file if we put
the Unicode replacement character in there.

src/language_features/document_links.rs
```diff
@@ -0,0 +1,213 @@
...
+            let command = match path {
+                Some(path) => format!("edit -existing {}", &editor_quote(&path)),
```
krobelus, line 188: Ok I guess URL is guaranteed to be absolute. Else we'd want to add a `--` here.

src/language_features/document_links.rs
```diff
@@ -0,0 +1,213 @@
...
+                None => format!("lsp-open-url {}", &editor_quote(target.as_str())),
+            };
+            (label, command)
+        })
+        .collect_vec();
+
+    match commands.len() {
+        0 => ctx.show_error(meta, "no document link in selection"),
+        1 => ctx.exec(meta, commands.into_iter().next().unwrap().1),
```
krobelus, line 197: (I think we do support `menu -auto-single` but I agree with explicitly short-circuiting here)

krobelus:
Seems nice!

Which server implements this? (so I can try :)

> Add to hover messages

agreed, that would fit in nicely

> Add line flags

Right. Probably only if there are no diagnostics or code lenses (?) on that line.
The existing logic which prioritizes diagnostics over lenses is

```kak
    let line_flags = diagnostics
        .merge_join_by(lenses, |left, right| left.0.cmp(&right.0))
        .map(|r| match r {
            EitherOrBoth::Left((line, diagnostic_label)) => (line, diagnostic_label),
            EitherOrBoth::Right((line, lens_label)) => (line, lens_label),
            EitherOrBoth::Both((line, diagnostic_label), _) => (line, diagnostic_label),
        })
```

> Add default mapping

Haven't scanned all bindings but `lsp L` (close to code lenses `lsp l`) or `lsp <ret>` comes to mind

topisani:
> Which server implements this? (so I can try :)
I implemented it for https://github.com/kylebonnici/dts-lsp, but that isnt very easy to test, as configuring it is a bit involved. It looks like https://github.com/artempyanykh/marksman might support it, but i havent tried it yet.

krobelus:
On Mon, Jan 20, 2025 at 01:28:00AM -0800, Tobias Pisani wrote:
> > Which server implements this? (so I can try :)
> I implemented it for https://github.com/kylebonnici/dts-lsp, but that isnt very easy to test, as configuring it is a bit involved. It looks like https://github.com/artempyanykh/marksman might support it, but i havent tried it yet.

I see. This feature sounds like `lsp-definition` but with extra highlighting for valid jump targets.

When there are multiple targets, `lsp-definition` shows a scratch buffer, not a menu.
Perhaps we should unify this in future

topisani:
> I see. This feature sounds like `lsp-definition` but with extra highlighting for valid jump targets. 

I think this is partially true. At least in theory i think the point is also to support linking to other resources, i.e. documentation. Hence why i added the unused lsp-open-url hook.

But in general, looking closer at it i cannot really find much support in servers, so it might make sense to just keep this basic implementation for now, and expand if we get more usecases. The way dts-lsp uses it, it should really support gotoDefinition too, so i might add that there also.

topisani:
> I see. This feature sounds like `lsp-definition` but with extra highlighting for valid jump targets. 

I think this is partially true. At least in theory i think the point is also to support linking to other resources, i.e. documentation. Hence why i added the unused lsp-open-url hook.

But in general, looking closer at it i cannot really find much support in servers, so it might make sense to just keep this basic implementation for now, and expand if we get more usecases. The way dts-lsp uses it, it should really support gotoDefinition too, so i might add that there also.  

edit: turns out clangd uses it for include files, thats an easy way to test it

krobelus:
For `file://` URIs it's probably better for users if they can use `gd` instead of having to learn a new shortcut. This already works in clangd today.

I guess `gd` doesn't include a preview of the file names but that can be provided in `lsp-hover` (actually the file name is already part of clangd's hover info on include directives.. showing it twice might be awkward but okay).

I wonder if using `gd` for both `lsp-definition` and `lsp-document-link`
is ever ambiguous.  I guess in practice it depends on what VSCode does. If
it uses ctrl-click for both (with no way to choose precedence), that would
probably discourage servers from making the two conflicts.

It's unclear what we should do on `https//` URIs, but that goes for `gd` as much as `lsp-document-link`.

From quick testing with clangd, I get the menu below.
If we keep this, we should probably strip the `file://` schema and reversing the URL-encoding, for a human readable name.

`'file:///usr/include/c%2B%2B/14.2.1/cstdio'   'file:///usr/include/stdio.h'`

krobelus:
I guess for `dts-lsp` we should be using `textDocument/definition`.

For clangd, it looks like the motivation was stated in https://github.com/clangd/clangd/issues/217

> Clangd does currently support go-to-definition here, which gets the job done but isn't as discoverable (the editor doesn't know where to show the links).

so I guess we don't have a way to take advantage of this currently.
Maybe we can teach Kakoune to emit OSC 8 hyperlinks; then that sounds like a nice improvement when using `clangd`.
Especially with terminals like `foot` that have a hint mode for copying links (and a `file:///$hostname/` prefix will be stripped).
So this would be a quick way to copy absolute paths to included files without moving the cursor.. nothing major but might be nice

PR #857, opened 2025-10-03
Change root glob resolution

khoover:
A bit subjective, but to me when the project is looking for roots it should stop at the first directory matching any of the globs, rather than going through each glob in order and matching the first directory containing it.

It leads to counterintuitive stuff like a Python project that uses `setup.py` and `requirements.txt` not being recognized as the root for itself because a parent contains a `pyproject.toml` as a uv workspace.

### DIFF
--- a/src/project_root.rs
+++ b/src/project_root.rs
@@ -37,9 +37,9 @@ fn roots_by_marker(to_editor: &ToEditorSender, roots: &[String], path: &str) -> 
         src.pop();
     }
 
-    for root in roots {
-        let mut pwd = src.clone();
-        loop {
+    let mut pwd = src.clone();
+    loop {
+        for root in roots {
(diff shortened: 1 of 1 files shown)


krobelus:
> Ah, forgot this monorepo had a requirements.txt in the git root
> because bazel. Yeah, it skipped right over the specific package's
> `pyproject.toml` and went to the git root because of that. Still
> think the behaviour isn't ideal, even if it's what other setups use.

possibly; I don't think we're gonna break backwards compatibility,
at least not for every single language unless we have a good rationale.

If we can argue that our python configuration is wrong, then,
by all means we can change that (to use different order,
or even a different algorithm).
Again, if we have enough spare time, we should check what other people
(VSCode/Helix) do.

We could add better/more intuitive names like root_globs_greedy or
root_globs_nongreedy (I'm sure there are better ones)

PR #859, opened 2025-10-13
Move cursor to start of selection

gbrlsnchs:
## Issue

* Go-to jump selections have the cursor positioned at the end of the selection
* Error find jump selections have the cursor positioned at the end of the selection

This is specially annoying on languages that select the whole function instead of just its name (i.e. Python and Ruby). Nonetheless, having the cursor at the beginning of the selection, even if it's in the same line, is  the superior UX decision anyway, as things are read from left to right.

## Solution

The `select` command takes arguments like `1.5,1.10` for, in this example, to select from line 1, column 5 to line 1, column 10. This ascending order places the cursor at the end of the selection. By inverting the arguments, we get the same selection but with the cursor correctly positioned at the start of the selection.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -1568,7 +1568,9 @@ Jump to the next or previous diagnostic error" %{
                 selection="$first"
             fi
         fi
-        printf 'select %s\n' "$selection"
+        start="${selection%,*}"
+        end="${selection#*,}"
+        printf 'select %s\n' "${end},${start}"
     }
 }
 


rc/lsp.kak
```diff
@@ -1568,7 +1568,9 @@ Jump to the next or previous diagnostic error" %{
                 selection="$first"
             fi
         fi
-        printf 'select %s\n' "$selection"
+        start="${selection%,*}"
+        end="${selection#*,}"
+        printf 'select %s\n' "${end},${start}"
```
gbrlsnchs: Same rationale here, inverting end and start will keep the same selection, but place the cursor at the beginning of it.

rc/lsp.kak
```diff
@@ -1568,7 +1568,9 @@ Jump to the next or previous diagnostic error" %{
                 selection="$first"
             fi
         fi
-        printf 'select %s\n' "$selection"
```
krobelus: > * Go-to jump selections have the cursor positioned at the end of the selection

right

 > * Error find jump selections have the cursor positioned at the end of the selection

 > This is specially annoying on languages that select the
 > whole function instead of just its name (i.e. Python and
 > Ruby).

I wonder which servers. Our default pylsp (v1.12.2) works fine for me, it selects only the function name.
IIRC gopls had this problem when jumping to packages,
but no longer does, now it only selects the `package statement`.

The only server I'm aware of is `rust-analyzer`; if you write

	use std::collections::HashMap;

and jump to `std` or `collections`,
then it will select the entire `mod.rs`.
I'm not sure if that's reasonable behavior.
For inline modules (`mod foo`) only the module name is selected.
Maybe we should report an issue to rust-analyzer,
because an empty range might be better than selecting the whole file.

 > Nonetheless, having the cursor at the beginning of the
 > selection, even if it's in the same line, is  the superior UX
 > decision anyway, as things are read from left to right.

Looks like Helix agrees with you.
For small selections, I don't have a strong opinion either way,
so I'm fine with this change.

Maybe add a changelog entry, and perhaps put the PR description into the
commit message (can merge all the behavior-changing
commits into one I think, so no commit introduces inconsistencies)

rc/lsp.kak
```diff
@@ -1568,7 +1568,9 @@ Jump to the next or previous diagnostic error" %{
...
+        start="${selection%,*}"
+        end="${selection#*,}"
```
krobelus: no need for quotes here BTW (irrespective of the variable value)

src/types.rs
```diff
@@ -568,7 +568,7 @@ impl Display for KakounePosition {
 
 impl Display for KakouneRange {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{},{}", self.start, self.end)
+        write!(f, "{},{}", self.end, self.start)
```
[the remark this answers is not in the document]
gbrlsnchs: I'm not opposed to following this suggestion, but I feel like this behavior should be "the rule" and, when (and if) we find a case that has been disrupted by this, we could introduce a new struct and treat that as an exception. WDYT?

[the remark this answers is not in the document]
krobelus: Yeah I guess it makes sense as default and then we can skip the extra type for now.
If I remove the `Display` implementation I get 11 compile errors. I assume you checked them;
I skimmed them and it seems like we are probably fine (unless someone is relying on the format of `lsp_inline_diagnostics` etc).

krobelus:
This broke lsp-object and lsp-selection-range,
both of which already had their custom logic to determine when to flip the selection.

It's possible that some of them should be changed in future,
but it should be an intentional change.  Merging the rest for now. Thanks!

PR #861, opened 2025-10-20
Add $/lean/plainTermGoal lsp method

Lqnk4:
Same as #860 but for the $/lean/plainTermGoal lsp method

The `lean-get-term-goal` command works in exactly the same way as the `lean-get-goal` command but instead fetches the term's expected type

I would like this lsp method to be merged since it offers the "Expected Type" feature found in the [lean.nvim](https://github.com/Julian/lean.nvim) infoview.

### DIFF
--- a/src/language_features/lean.rs
+++ b/src/language_features/lean.rs
@@ -5,6 +5,12 @@ use crate::util::*;
 use indoc::formatdoc;
 use lsp_types::request::Request;
 use lsp_types::*;
+use serde::Deserialize;
+use serde::Serialize;
+
+pub trait LeanPlainGoalResponse: Serialize + for<'de> Deserialize<'de> {
+    fn get_rendered(self) -> String;
+}
 
 #[derive(Debug, Clone)]
 pub struct EditorPlainGoalParams {
@@ -24,6 +30,11 @@ pub struct PlainGoalResponse {
     pub goals: Vec<String>,
     pub rendered: String,
 }
+impl LeanPlainGoalResponse for PlainGoalResponse {
+    fn get_rendered(self) -> String {
+        self.rendered
+    }
+}
 
 pub struct PlainGoalRequest {}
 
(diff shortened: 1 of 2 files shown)


src/language_features/lean.rs
```diff
@@ -77,3 +77,76 @@ pub fn plain_goal(meta: EditorMeta, params: EditorPlainGoalParams, ctx: &mut Con
         },
     );
 }
+
+#[derive(Debug, Clone)]
+pub struct EditorPlainTermGoalParams {
+    pub position: KakounePosition,
+    pub buffer: String,
+}
```
krobelus: I think we should reuse `EditorPlainGoalParams` here,
since the two are unlikely to deviate.

src/language_features/lean.rs
```diff
@@ -77,3 +77,76 @@ pub fn plain_goal(meta: EditorMeta, params: EditorPlainGoalParams, ctx: &mut Con
...
+pub struct PlainTermGoalResponse {
+    pub goal: String,
+    pub range: Range,
+}
+
+pub struct PlainTermGoalRequest {}
+
+impl Request for PlainTermGoalRequest {
+    type Params = PlainTermGoalParams;
+    type Result = Option<PlainTermGoalResponse>;
```
krobelus: I think we should reuse `PlainGoalParams` as-is here rather than defining our own type,
that might be unconventional but I think it would help with code sharing

src/language_features/lean.rs
```diff
@@ -77,3 +77,76 @@ pub fn plain_goal(meta: EditorMeta, params: EditorPlainGoalParams, ctx: &mut Con
...
+                    .unwrap()
+                }],
+            )
+        })
+        .collect();
+
+    ctx.call::<PlainTermGoalRequest, _>(
+        meta,
+        RequestParams::Each(req_params),
+        move |ctx: &mut Context, _meta, results| {
```
krobelus: rather than copying code, we should extract a function.
At least for the response handling (which is not just boilerplate),
but probably for the entire `plain_term_goal()`.

It will be a generic function; the two callers will pass PlainGoalRequest and PlainTermGoalRequest as concrete type parameters.

	fn plain_goal_impl<T: Request>(meta: EditorMeta, params: EditorPlainGoalParams, ctx: &mut Context) {
		...
		ctx.call::<T>(...);
	}

krobelus:
nice work! I think that's a great solution. Unlike C++, Rust requires explicitly specifying what generics types can do - this allows it to type-check the implementation of the generic function. This is good, because things will fail faster, and the explicit annotation helps the reader. If we wanted to save characters we could use macros..

PR #864, opened 2025-11-02
Add inlay hint textedit support

ficd0:
~~Added a new method kakoune/inlay-hint-apply-textedit. There is now a Kakoune command that calls this method. It applies the TextEdits associated with the inlayHint that's nearest the cursor.~~

There is now a "nearest" and "selected" version of the command. Nearest applies the hint that's closest to the cursor, on the same line. Selected applies all hints that fall within the selections. Both commands now use `selections_desc` and operate per-selection.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -356,6 +356,16 @@ lsp-inlay-hints-enable global
 
 You can change the hints' face with `set-face global InlayHint <face>`.
 
+Some language servers support "applying" inlay hints: inserting them as text into your document.
+For example, when using `basedpyright` to write Python,
+inlay hints which represent type hints can be explicitly inserted into your code.
+
+We provide two commands for applying inlay hints:
+`lsp-inlay-hint-apply-nearest` and `lsp-inlay-hint-apply-selected`.
+They both operate _per selection_.
+`nearest` applies whichever hint is closest to the cursor, *on the same line*.
+`selected` applies *all* hints that fall inside the selection.
+
 === Semantic Tokens
 
 kakoune-lsp supports the semanticTokens feature for semantic highlighting. If the language server supports it, you can enable it with:
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -1232,6 +1232,14 @@ define-command -hidden lsp-inlay-hints -docstring "lsp-inlay-hints: request inla
     }
 }
 
+define-command lsp-inlay-hint-apply-nearest -docstring "Apply the textedit for the inlay hint nearest the cursor, for each selection" %{
+    lsp-send kakoune/inlay-hint-apply-nearest %val{selection_count} %val{selections_desc}
+}
+
+define-command lsp-inlay-hint-apply-selected -docstring "Apply the textedit for the inlay hints overlapping with the selections" %{
+    lsp-send kakoune/inlay-hint-apply-selected %val{selection_count} %val{selections_desc}
+}
+
 # CCLS Extension
 
 define-command ccls-navigate -docstring "Navigate C/C++/ObjectiveC file" -params 1 %{
--- a/src/context.rs
+++ b/src/context.rs
@@ -79,6 +79,7 @@ pub struct Context {
     pub diagnostics: HashMap<String, Vec<(ServerId, Diagnostic)>>,
     pub documents: HashMap<String, Document>,
     pub dynamic_config: DynamicConfig,
+    pub inlay_hints: HashMap<String, Vec<(ServerId, InlayHint)>>,
     pub language_servers: BTreeMap<ServerId, ServerSettings>,
     pub route_cache: HashMap<(ServerName, RootPath), ServerId>,
     pub outstanding_requests:
@@ -118,6 +119,7 @@ impl Context {
             diagnostics: Default::default(),
             documents: Default::default(),
             dynamic_config: DynamicConfig::default(),
+            inlay_hints: Default::default(),
             language_servers: BTreeMap::new(),
             route_cache: HashMap::new(),
             outstanding_requests: HashMap::default(),
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -38,7 +38,7 @@ use ccls::{EditorCallParams, EditorInheritanceParams, EditorMemberParams, Editor
 use code_lens::{text_document_code_lens, CodeLensOptions};
 use crossbeam_channel::{after, never, tick, Receiver, Select, Sender};
 use indoc::formatdoc;
-use inlay_hints::InlayHintsOptions;
+use inlay_hints::{InlayHintApplyParams, InlayHintsOptions};
 use itertools::Itertools;
 use jsonrpc_core::{Call, ErrorCode, MethodCall, Output, Params};
 use lean::EditorPlainGoalParams;
@@ -424,6 +424,20 @@ fn dispatch_fifo_request(
         "kakoune/exit" => Box::new(()),
         "kakoune/goto-document-symbol" => Box::new(GotoSymbolParams {
             goto_symbol: state.next()?,
+        }),
+        "kakoune/inlay-hint-apply-nearest" => Box::new(InlayHintApplyParams {
+            selections_desc: {
+                let selection_count = state.next()?;
+                state.next_vec(selection_count)?
+            },
+            kind: inlay_hints::InlayHintApplyKind::Nearest,
+        }),
+        "kakoune/inlay-hint-apply-selected" => Box::new(InlayHintApplyParams {
+            selections_desc: {
+                let selection_count = state.next()?;
+                state.next_vec(selection_count)?
+            },
+            kind: inlay_hints::InlayHintApplyKind::Selected,
         }),
         "kakoune/next-or-previous-symbol" => {
             let num_symbol_kinds = state.next()?;
@@ -1806,6 +1820,12 @@ fn dispatch_editor_request(request: EditorRequest, ctx: &mut Context) -> Control
             }
             let params: CompletionParameters = params.unbox();
             let _ = std::fs::write(params.response_fifo, commands.as_bytes());
+        }
+        "kakoune/inlay-hint-apply-nearest" => {
+            inlay_hints::inlay_hint_apply(meta, params.unbox(), ctx);
+        }
+        "kakoune/inlay-hint-apply-selected" => {
+            inlay_hints::inlay_hint_apply(meta, params.unbox(), ctx);
         }
         "kakoune/next-or-previous-symbol" => {
             document_symbol::next_or_prev_symbol(meta, params.unbox(), ctx);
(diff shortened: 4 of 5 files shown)


README.asciidoc
```diff
@@ -354,6 +354,13 @@ lsp-inlay-hints-enable global
 
 You can change the hints' face with `set-face global InlayHint <face>`.
 
+When supported by the LSP, you can "apply" the inlay hint nearest the cursor:
+
+[source,kak]
+----
+lsp-inlay-hint-apply-textedit
```
krobelus: interesting, which server supports these?
I imagine the purpose is autofixes for issues flagged by inlay hints.
Though there are already code actions for some of those.. but I guess the UI difference is that the inlay hints are clickable in VSCode.
(And some servers even use code lenses for autofixes..)

If it's something that commonly-used servers support, maybe add a changelog entry.

src/context.rs
```diff
@@ -118,6 +119,7 @@ impl Context {
             diagnostics: Default::default(),
             documents: Default::default(),
             dynamic_config: DynamicConfig::default(),
+            inlay_hints: Default::default(),
```
krobelus: I'm mildly surprised that we don't reliably delete from the per-buffer maps.
There's some code for "ctx.diagnostics" to empty diagnostcs;
I was wondering if we should also reset other data here:

```diff
diff --git a/src/text_sync.rs b/src/text_sync.rs
index db412e8d9c..4d34adc8cf 100644
--- a/src/text_sync.rs
+++ b/src/text_sync.rs
@@ -67,6 +67,8 @@
     // Resets metadata for buffer.
     ctx.documents.insert(meta.buffile.clone(), document);
     ctx.diagnostics.insert(meta.buffile.clone(), Vec::new());
+    ctx.code_lenses.insert(meta.buffile.clone(), Vec::new());
+    ctx.inlay_hints.insert(meta.buffile.clone(), Vec::new());
 
     let req_params = DidChangeTextDocumentParams {
         text_document: VersionedTextDocumentIdentifier {
```

this should be fine for code lenses because by default we send
lsp-code-lens-request always immediately after lsp-did-change.

Not sure if it's the same for inlay hints,
but either way we should try to resolve any inconsistency between these.

and in text_document_did_close(), we can remove the per-buffer data.

Probably we should create a single map for all per-buffer data
(include ctx.documents)

Of course this thread is unrelated, better suited for a future PR.

src/controller.rs
```diff
@@ -417,6 +417,9 @@ fn dispatch_fifo_request(
             position_line: state.next()?,
         }),
         "kakoune/exit" => Box::new(()),
+        "kakoune/inlay-hint-apply-textedit" => Box::new(InlayHintApplyParams {
```
krobelus: should sort alphabetically here

src/language_features/inlay_hints.rs
```diff
@@ -71,6 +72,10 @@ pub fn inlay_hints_response(
         Some(document) => document,
         None => return,
     };
+
+    ctx.inlay_hints
+        .insert(meta.buffile.clone(), inlay_hints.clone());
```
krobelus: probably we shouldn't clone inlay hints here,
since it's not really needed.

```diff
diff --git a/src/language_features/inlay_hints.rs b/src/language_features/inlay_hints.rs
index fad1d275a7..95b956cd59 100644
--- a/src/language_features/inlay_hints.rs
+++ b/src/language_features/inlay_hints.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use itertools::Itertools;
 use lsp_types::{
     request::InlayHintRequest, InlayHint, InlayHintLabel, InlayHintParams, Position, Range,
@@ -73,11 +75,8 @@
         None => return,
     };
 
-    ctx.inlay_hints
-        .insert(meta.buffile.clone(), inlay_hints.clone());
-
     let ranges = inlay_hints
-        .into_iter()
+        .iter()
         .map(
             |(
                 server_id,
@@ -89,13 +88,13 @@
                     ..
                 },
             )| {
-                let server = ctx.server(server_id);
+                let server = ctx.server(*server_id);
                 let position =
                     lsp_position_to_kakoune(&position, &document.text, server.offset_encoding);
                 let label = match label {
-                    InlayHintLabel::String(s) => s,
+                    InlayHintLabel::String(s) => Cow::Borrowed(s),
                     InlayHintLabel::LabelParts(parts) => {
-                        parts.iter().map(|x| x.value.as_str()).collect()
+                        Cow::Owned(parts.iter().map(|x| x.value.as_str()).collect())
                     }
                 };
                 let padding_left = if padding_left.unwrap_or(false) {
@@ -115,6 +114,9 @@
             },
         )
         .join(" ");
+
+    ctx.inlay_hints.insert(meta.buffile.clone(), inlay_hints);
+
     let version = meta.version;
     let command = format!("set-option buffer lsp_inlay_hints {version} {ranges}");
     let command = format!(
```

src/language_features/inlay_hints.rs
```diff
@@ -119,3 +124,60 @@ pub fn inlay_hints_response(
...
+
+#[derive(Debug)]
+pub struct InlayHintApplyParams {
+    pub selection_desc: String,
+}
+
+pub fn apply_inlay_hint(meta: EditorMeta, params: InlayHintApplyParams, ctx: &mut Context) {
+    let document = match ctx.documents.get(&meta.buffile) {
+        Some(doc) => doc,
+        None => return, // buffer is gone
```
krobelus: clippy should yell at you to use let-else here `let Some(document)  = ctx.docuemtns.get() else { return; };`

README.asciidoc
ficd0: Main one I can think of is `basedpyright`, which allows explicitly adding in the type hints shown in inlay hints.

ficd0:
I've applied the changes, except for the manhattan distance stuff -- frankly, I'm a little clueless about what to do there.

src/language_features/inlay_hints.rs
```diff
@@ -119,3 +124,60 @@ pub fn inlay_hints_response(
...
+    // find matching inlay hint
+    let (server_id, hint) = match hints.iter().min_by_key(|(server_id, hint)| {
+        let server = ctx.server(*server_id);
+        let kak_pos =
+            lsp_position_to_kakoune(&hint.position, &document.text, server.offset_encoding);
+        // manhattan distance from cursor
+        let dl = kak_pos.line.abs_diff(cursor.line) as u64;
+        let dc = kak_pos.column.abs_diff(cursor.column) as u64;
+        dl.saturating_mul(1000) + dc
+    }) {
```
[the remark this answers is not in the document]
ficd0: Whoops, I didn't see this earlier. I think this is very reasonable. I'll take a shot at implementing this in a bit.

README.asciidoc
```diff
@@ -354,6 +354,8 @@ lsp-inlay-hints-enable global
 
 You can change the hints' face with `set-face global InlayHint <face>`.
 
+Some LSPs support inserting the inlay hints as text into your document. This can be useful when using `basedpyright` to write Python code, since it can make your type hints explicit. To do this, put the primary cursor next to the hint, and run `lsp-inlay-hint-apply-textedit`.
```
krobelus: We're inconsistent about line wrapping,
but I think we should use semantic line breaks (like, one sentence per line),
mostly for easier editing and better diffs
https://nick.groenen.me/notes/one-sentence-per-line/

(also because Kakoune doesn't have great support for softwrapping yet haha)

rc/lsp.kak
```diff
@@ -1226,6 +1226,10 @@ define-command -hidden lsp-inlay-hints -docstring "lsp-inlay-hints: request inla
     }
 }
 
+define-command lsp-inlay-hint-apply-textedit -docstring "Apply the textedit for the inlay hint nearest the cursor" %{
```
krobelus: I suppose we could avoid adding magic behavior based on selection size,
by defining two different commands

1. lsp-inlay-hint-apply-textedit (apply nearest edit)
2. lsp-inlay-hint-apply-textedits (apply all edits)

or equivalently, do it with a flag like "-all", not sure which is better.

rc/lsp.kak
```diff
@@ -1226,6 +1226,10 @@ define-command -hidden lsp-inlay-hints -docstring "lsp-inlay-hints: request inla
...
+    lsp-send kakoune/inlay-hint-apply-textedit %val{selection_desc}
```
krobelus: if we want to handle all selections, we'd want to pass `selections_desc'.

But lik esome other comments, feel free to leave it for later if you don't want to spend the extra work now.

src/controller.rs
```diff
@@ -416,10 +416,33 @@ fn dispatch_fifo_request(
         "kakoune/breadcrumbs" => Box::new(BreadcrumbsParams {
             position_line: state.next()?,
         }),
+        "kakoune/did-change-option" => {
```
krobelus: (doesn't matter here but in general it's better to put such sorting changes in an earlier commit)

src/language_features/inlay_hints.rs
```diff
@@ -119,3 +126,46 @@ pub fn inlay_hints_response(
...
+
+    // for now let's use cursor as the target
+    let (_, cursor) = parse_kakoune_range(&params.selection_desc);
+
+    let Some((server_id, hint)) = ctx
+        .inlay_hints
+        .get(&meta.buffile)
+        .unwrap()
+        .iter()
+        .min_by_key(|(server_id, hint)| {
```
krobelus: let's maybe consider using *all* closest matches,
instead of an arbitrary one

src/controller.rs
```diff
@@ -457,26 +471,6 @@ fn dispatch_fifo_request(
         "kakoune/textDocument/codeLens" => Box::new(CodeLensOptions {
             selection_desc: state.next()?,
         }),
-        "kakoune/did-change-option" => {
```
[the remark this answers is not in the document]
ficd0: I'm actually not sure, this is weird because I don't remember deleting this. If you recall in my previous draft I sorted the match arms alphabetically, perhaps the combination of that with rebasing on master and squashing my new commits in caused this. I'll go ahead and put this back in

src/language_features/inlay_hints.rs
```diff
@@ -119,3 +128,91 @@ pub fn inlay_hints_response(
...
+        };
+
+        if edits_by_server.is_empty() {
+            ctx.show_error(meta.clone(), "no textedits available to apply");
+            continue;
+        }
+
+        let uri = Url::from_file_path(&meta.buffile).unwrap();
+        for (server_id, edits) in edits_by_server {
+            apply_text_edits(server_id, meta.clone(), uri.clone(), edits, ctx);
```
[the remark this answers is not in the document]
ficd0: OK, I'll add a FIXME comment to the code for now

PR #865, opened 2025-11-11
Add LSP 3.17 pull diagnostics support

saifulapm:
Implement textDocument/diagnostic and workspace/diagnostic with automatic refresh on didOpen, didChange, and didSave events.

Add lsp-workspace-diagnostics command for workspace-wide diagnostics. Handle workspace/diagnostic/refresh and eslint/status notifications.


src/capabilities.rs
```diff
@@ -679,6 +690,27 @@ pub fn capabilities(meta: EditorMeta, ctx: &mut Context) {
             .entry("lsp-diagnostics".to_string())
             .or_default()
             .push(server_name);
+        probe_feature(to_editor, entry, &mut features, CAPABILITY_DIAGNOSTIC);
+
+        // Check for workspace diagnostics support
+        if server_has_capability(to_editor, server_settings, CAPABILITY_DIAGNOSTIC) {
+            if let Some(ref caps) = server_settings.capabilities {
+                let supports_workspace = match &caps.diagnostic_provider {
+                    Some(DiagnosticServerCapabilities::Options(opts)) => opts.workspace_diagnostics,
```
krobelus: squash the formatting commit into the other one

src/capabilities.rs
```diff
@@ -679,6 +690,27 @@ pub fn capabilities(meta: EditorMeta, ctx: &mut Context) {
...
+                    Some(DiagnosticServerCapabilities::RegistrationOptions(opts)) => {
+                        opts.diagnostic_options.workspace_diagnostics
```
krobelus: we're sort of duplicating code here (i.e. we check `opts.diagnostic_options.workspace_diagnostics` in two places);
how about defining a `CAPABILITY_WORKSPACE_DIAGNOSTICS = "lsp-workspace-diagnostics"`,
and use that in both places?

src/controller.rs
```diff
@@ -518,7 +518,9 @@ fn dispatch_fifo_request(
         | "textDocument/typeDefinition" => Box::new(PositionParams {
             position: state.next()?,
         }),
-        "textDocument/diagnostics" | "textDocument/documentSymbol" => Box::new(()),
+        "textDocument/diagnostics" | "workspace/diagnostics" | "textDocument/documentSymbol" => {
```
krobelus: I think the addition should be sorted next to the other `workspace/` cases.

src/diagnostics.rs
```diff
@@ -394,3 +395,507 @@ pub fn format_related_information(
...
+    if req_params.is_empty() {
+        return;
+    }
+
+    ctx.call::<DocumentDiagnosticRequest, _>(
+        meta,
+        RequestParams::Each(req_params),
+        move |ctx, meta, results| {
+            for (server_id, report) in results {
+                handle_document_diagnostic_response(server_id, report, &meta, ctx);
```
krobelus: Hmm I'm not sure if it's a good idea to run one `ctx.exec()` per server.
It doesn't really matter in practice as multi-servers is an edge case,
but ideally we coalesce them all in one big command.

src/diagnostics.rs
```diff
@@ -394,3 +395,507 @@ pub fn format_related_information(
...
+    );
+}
+
+fn handle_document_diagnostic_response(
+    server_id: ServerId,
+    report: DocumentDiagnosticReportResult,
+    meta: &EditorMeta,
+    ctx: &mut Context,
+) {
+    let buffile = &meta.buffile;
```
krobelus: so this request is supposed to retrieve diagnostics,
but only for the requested file.
In that case, it makes sense that we're storing them per-file.

A tangentially related issue is that we currently fail to clean 
them up when the buffere is deleted, see
https://github.com/kakoune-lsp/kakoune-lsp/pull/864#discussion_r2484524719
We should fix that eventually..

saifulapm:
Closing this PR. I'll keep this feature in my own fork for personal use. Thanks for the review feedback\!

PR #877, opened 2026-01-20
Add support for `metals/status` LSP extension

igor-ramazanov:
# Metals LSP extensions
Editor clients can opt into receiving Metals-specific JSON-RPC requests and notifications. Metals extensions are not defined in LSP and are not strictly required for the Metals server to function but it is recommended to implement them to improve the user experience.

See:
https://scalameta.org/metals/docs/integrations/new-editor#metalsstatus

## `metals/status`
The Metals status notification is sent from the server to the client to notify about non-critical and non-actionable events that are happening in the server. Metals status notifications are a complement to `window/showMessage` and `window/logMessage`.
Unlike `window/logMessage`, status notifications should always be visible in the user interface.
Unlike `window/showMessage`, status notifications are not critical meaning that they should not demand too much attention from the user.

In general, Metals uses status notifications to update the user about ongoing events in the server such as batch compilation in the build server or when a successful connection was established with the build server.

### DIFF
--- a/src/language_features/metals.rs
+++ b/src/language_features/metals.rs
@@ -1,0 +1,57 @@
+use jsonrpc_core::Params;
+
+use crate::{context::Context, editor_quote, EditorMeta, ServerId, LAST_CLIENT};
+
+/// https://scalameta.org/metals/docs/integrations/new-editor#metalsstatus
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct MetalsStatusParams {
+    /// The text to display in the status bar.
+    pub text: String,
+    /// If true, show the status bar.
+    pub show: Option<bool>,
+    /// If true, hide the status bar.
+    pub hide: Option<bool>,
+    /// If set, display this message when user hovers over the status bar.
+    pub tooltip: Option<String>,
+    /// If set, execute this command when the user clicks on the status bar item.
+    pub command: Option<String>,
+}
+
+/// Maps 'metals/status' to 'lsp-show-message-info'.
+/// Doesn't support the 'command' parameter as it's not clear how to action on the notification.
+pub fn status(server_id: ServerId, meta: EditorMeta, params: Params, ctx: &mut Context) {
+    let params: MetalsStatusParams = params.parse().expect("Failed to parse semhl params");
+
+    let show = params.show.unwrap_or(false) && !params.hide.unwrap_or(false);
+
+    if show {
+        let have_client = meta.client.is_some();
+        let last_client = LAST_CLIENT.lock().unwrap();
+        let client = if have_client {
+            ""
+        } else {
+            &last_client
+                .as_ref()
+                .map(|client| client.as_str())
+                .unwrap_or_default()
+        };
+        let msg = format!(
+            "{}{}",
+            params.text,
+            params
+                .tooltip
+                .map(|tooltip| format!(": {}", tooltip))
+                .unwrap_or("".to_string())
+        );
+
+        ctx.exec(
+            meta,
+            format!(
+                "evaluate-commands -verbatim -try-client '{}' lsp-show-message-info {} {}",
+                client,
+                editor_quote(&ctx.server(server_id).name),
+                editor_quote(&msg)
+            ),
+        );
+    }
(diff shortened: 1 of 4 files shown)


src/language_features/metals.rs
```diff
@@ -0,0 +1,57 @@
+use jsonrpc_core::Params;
```
krobelus, line 1: I don't know if this is a sensible LSP extension;
probably it would be better to create something that could be standardized eventually.
I wonder if metals could do downgrade to `window/showMessage` automatically if `metals/showMessage` is not supported by the client..
Not sure..

If not, it seems okay to add this. Merge as you like.

src/language_features/metals.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// https://scalameta.org/metals/docs/integrations/new-editor#metalsstatus
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct MetalsStatusParams {
+    /// The text to display in the status bar.
+    pub text: String,
+    /// If true, show the status bar.
+    pub show: Option<bool>,
+    /// If true, hide the status bar.
+    pub hide: Option<bool>,
```
krobelus, line 13: So if neither of show or hide is set, the status bar state is supposed to stay the same?
I guess Kakoune doesn't really have the concept of hidden/shown status bar;
it's only shown when new text is added.

We could technically keep track of whether the last metals/status showed the status bar,
and if it did, then the current `metals/status` updates the status bar even if `show` is not set

We could do something similar to `hide`,
but that's even less relevant because status is cleared quickly anyway.

No need to address this.
This is LSP message is clearly overengineered for the Kakoune use case :)

src/language_features/metals.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// If set, execute this command when the user clicks on the status bar item.
+    pub command: Option<String>,
+}
+
+/// Maps 'metals/status' to 'lsp-show-message-info'.
+/// Doesn't support the 'command' parameter as it's not clear how to action on the notification.
+pub fn status(server_id: ServerId, meta: EditorMeta, params: Params, ctx: &mut Context) {
+    let params: MetalsStatusParams = params.parse().expect("Failed to parse semhl params");
+
+    let show = params.show.unwrap_or(false) && !params.hide.unwrap_or(false);
```
krobelus, line 25: I don't strongly care but I'd think that if `show` is false,
we should still call `lsp-show-message-info`,
except without a client context, so we don't change the user's status bar,
but still log to `*debug*`.

src/language_features/metals.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    if show {
```
krobelus, line 27: should use early return? (`if !show { return }`)

src/language_features/metals.rs
```diff
@@ -0,0 +1,57 @@
...
+        let have_client = meta.client.is_some();
+        let last_client = LAST_CLIENT.lock().unwrap();
+        let client = if have_client {
+            ""
+        } else {
+            &last_client
+                .as_ref()
+                .map(|client| client.as_str())
+                .unwrap_or_default()
+        };
```
krobelus, line 37: ah this logic is copied from src/show_message.rs.
Would be good to extract it into a function.

src/language_features/metals.rs
igor-ramazanov: Thanks for the review!

I also think the complexity/benefits ratio for this particular PR isn't that good, and I also wish `metals` would move away from using LSP extensions.

> I wonder if metals could do downgrade to window/showMessage automatically if metals/showMessage is not supported by the client..

Yes, it does, and that's how it works now. It supports one of `window/showMessage`, `window/logMessage` or `metals/status`.

My overall idea was to eventually support as much metals extensions as possible to provide a complete Scala programming experience.

After giving it a bit of a thought, I started to think that maybe it's better if I keep it in a privately maintained fork and see if I can support more of their LSP extensions before proposing PRs.

It's good to keep it simple, and makes a sense to wait a bit if/when there'd be an implementation with more benefits.

PR #883, opened 2026-02-27
completion: support server-defined triggerCharacters

### LINKED ISSUE #882: Completion: support server-defined `triggerCharacters`
Some LSP servers register `triggerCharacters` in their completion capabilities
to tell the client which characters should immediately trigger a completion
request. kakoune-lsp currently ignores these characters — completion is driven
entirely by the `InsertIdle` hook and `lsp_completion_trigger` /
`lsp_completion_fragment_start` options on the Kakoune side.

This works well for typical identifier completion (`\w+`), but I've found that
servers relying on non-word trigger characters don't get completions in certain
cases.

### Example: markdown-oxide

markdown-oxide registers these trigger characters:

    [ ] ( # > <space>

I observe the following failures:

| Input | Expected | Actual | Why |
|---|---|---|---|
| `[title](` | file path completions | no completion menu | `(` is not `\w`, `lsp_completion_fragment_start` fails to find a token start |
| `[[^` | block reference completions | no completion menu | `^` is not `\w` |
| `> [!` | callout completions | no completion menu | `!` is not `\w` |
| `word ` (with trailing space) | completions after space trigger | no completion menu | `lsp_completion_trigger` rejects: preceding char is space |

These cases pass `lsp_completion_trigger` (except the space case) because the
preceding char is not whitespace, but `lsp_completion_fragment_start` cannot
find a `\w` token to track back to.

The workaround is to set `lsp_extra_word_chars` per server, but this also
affects Kakoune's `w`/`b`/`e` motions and doesn't scale across servers.

### Reproduction

1. Install markdown-oxide and configure kakoune-lsp for it
2. Open a Markdown file inside an Obsidian vault (with `.obsidian/` directory)
3. Type `[title](` — no completion menu appears
4. Check `*debug*` buffer — no completion request is sent to the server

### What other clients do

Helix [checks the server's `trigger_characters` directly](https://github.com/helix-editor/helix/blob/master/helix-term/src/handlers/completion.rs)
and fires a `TriggerChar` completion event when the typed text ends with one of
them.

### Possible approach

**MVP**: After initialization, extract `completion_provider.trigger_characters`
from the server capabilities and expose them as a Kakoune option (e.g.
`lsp_completion_trigger_characters`). In `lsp-completion`, when the cursor
follows a trigger character, skip the `lsp_completion_fragment_start` check and
use the cursor position as the completion offset.

**Optional follow-ups**:
- Set `context_support: Some(true)` and send `CompletionContext` with
  `trigger_kind: TriggerCharacter` and `trigger_character` in the request
- Consider using `InsertChar` hook for trigger-character completions instead of
  waiting for `InsertIdle`
- Handle whitespace triggers (e.g. space) that currently conflict with
  `lsp_completion_trigger`'s default `\S.\z` check

### Questions

- Is this something you'd be interested in? I'm happy to work on a PR if the
  general direction makes sense.
- Would you prefer the trigger character check in Kakoune script (modifying
  `lsp_completion_trigger`) or in Rust?
- Should this be opt-in (per-server option) to avoid behavior changes for
  existing setups?
- How should whitespace triggers be handled, given `lsp_completion_trigger`'s
  default rejects spaces?

Yus314:
Closes #882.

## Summary

When an LSP server registers `triggerCharacters` in its completion capabilities,
kakoune-lsp now detects them and fires completion requests accordingly, with
proper `CompletionContext` sent to each server.

### Changes

1. **Kakoune script** (`rc/lsp.kak`):
   - `lsp-completion` now has two paths:
     - **Path 1**: If the character before the cursor matches one of the
       server's trigger characters (`lsp_completion_trigger_regex`), fire a
       completion request immediately, bypassing `lsp_completion_trigger`.
       This handles whitespace triggers and non-word triggers like `(`, `[`, etc.
     - **Path 2** (fallback): The existing `lsp_completion_trigger` /
       `lsp_completion_fragment_start` logic, unchanged.
   - Trigger character extraction uses `<a-k>` as a guard (keeping the
     selection only if the line ends with a trigger character) followed by
     cursor movement (`<a-:>;h`) to isolate the character. This avoids shell
     subprocess spawning on the `InsertIdle` hot path.
   - The trigger character is passed to the Rust side via the FIFO.

2. **Rust side**:
   - `send_completion_trigger_characters()` extracts trigger characters from all
     servers' capabilities and sets `lsp_completion_trigger_regex` (the union,
     in a `BTreeSet` for dedup).
   - Per-server `CompletionContext` with `TriggerCharacter` kind is sent only to
     servers that actually registered the trigger character.
   - `context_support: true` is now advertised in client capabilities.

3. **filterText for snippet completions** (`completion.rs`):
   - Snippet completions now use `filter_text` (if provided by the server)
     instead of the `insert_text` for Kakoune's completion filtering. This
     fixes cases where the filter text differs from the snippet body.

### Design decisions

- **InsertIdle, not InsertChar**: Per maintainer guidance, InsertIdle is kept
  to avoid issues with pasting in insert mode.
- **Union of trigger chars**: Multi-server trigger characters are unioned.
  Per-server filtering happens on the Rust side when constructing
  `CompletionContext`.
- **No `Invoked` kind**: Kakoune has no concept of manual completion invocation.

### Test plan

- With markdown-oxide: `[title](` now triggers file path completions
- With rust-analyzer: `.` triggers method completions with `CompletionContext`
- Non-trigger characters still use the existing completion path
- Multiple LSP servers in the same buffer work correctly (each gets context
  only for its own trigger characters)

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -128,6 +128,7 @@ declare-option -docstring "Completion request is sent only when this expression 
 # By default, it tracks back to the first punctuation or whitespace.
 declare-option -docstring "Select from cursor to the start of the term being completed" str lsp_completion_fragment_start %{execute-keys "<esc><a-h>s\$?[\w%opt{lsp_extra_word_chars}]+.\z<ret>"}
 declare-option -hidden str lsp_extra_word_chars
+declare-option -hidden str lsp_completion_trigger_regex
 # Update lsp_extra_word_chars whenever extra_word_chars changes.
 # We could avoid this if we are not enabled, but this doesn't trigger that often and that would complicate initialization.
 hook -group lsp-extra-word-chars global WinSetOption extra_word_chars=.* %{
@@ -698,25 +699,40 @@ define-command -hidden lsp-did-change -docstring "Notify language server about b
 }
 
 define-command -hidden lsp-completion -docstring "Request completions for the main cursor position" %{
+    declare-option -hidden str lsp_completion_offset
+    declare-option -hidden str lsp_trigger_character
+
+    set-option window lsp_trigger_character ''
+    set-option window lsp_completion_offset %val{cursor_column}
+
     try %{
-        # Fail if preceding character is a whitespace (by default; the trigger could be customized).
-        evaluate-commands -draft %opt{lsp_completion_trigger}
-
-        # Kakoune requires completions to point fragment start rather than cursor position.
-        # We try to detect it and put into lsp_completion_offset and then pass via completion.offset
-        # parameter to the kakoune-lsp server so it can use it when sending completions back.
-        declare-option -hidden str lsp_completion_offset
-
-        set-option window lsp_completion_offset %val{cursor_column}
-        evaluate-commands -draft %{
-            try %{
-                evaluate-commands %opt{lsp_completion_fragment_start}
-                set-option window lsp_completion_offset %val{cursor_column}
+        try %{
+            # Path 1: Trigger character detection
+            # In insert mode, <a-h> selects the line including trailing newline.
+            # <a-k> keeps the selection only if it ends with a trigger char
+            # before the newline; on success we extract it with <a-:> (ensure
+            # forward), ; (reduce to cursor at the trailing newline), h (back
+            # one character to the trigger char).
+            evaluate-commands -draft %{
+                execute-keys "<a-h><a-k>[%opt{lsp_completion_trigger_regex}]\n\z<ret>"
+                execute-keys "<a-:>;h"
+                set-option window lsp_trigger_character %val{selection}
+            }
+        } catch %{
+            # Path 2: Normal completion (no trigger character or no match)
+            # Standard trigger check (default: preceding character is not whitespace).
+            evaluate-commands -draft %opt{lsp_completion_trigger}
+            # Detect fragment start position.
+            evaluate-commands -draft %{
+                try %{
+                    evaluate-commands %opt{lsp_completion_fragment_start}
+                    set-option window lsp_completion_offset %val{cursor_column}
+                }
             }
         }
 
         lsp-send textDocument/completion %val{cursor_line} %val{cursor_column} \
-            %opt{lsp_completion_offset}
+            %opt{lsp_completion_offset} %opt{lsp_trigger_character}
     }
 }
 
--- a/src/capabilities.rs
+++ b/src/capabilities.rs
@@ -10,6 +10,7 @@ use lsp_types::notification::*;
 use lsp_types::request::*;
 use lsp_types::*;
 use std::collections::BTreeMap;
+use std::collections::BTreeSet;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::process;
@@ -175,7 +176,7 @@ pub fn initialize(meta: EditorMeta, ctx: &mut Context, servers: Vec<ServerId>) {
                                         CompletionItemKind::TYPE_PARAMETER,
                                     ]),
                                 }),
-                                context_support: Some(false),
+                                context_support: Some(true),
                                 insert_text_mode: None,
                                 completion_list: None,
                             }),
@@ -408,7 +409,7 @@ pub fn initialize(meta: EditorMeta, ctx: &mut Context, servers: Vec<ServerId>) {
         })
         .collect();
 
-    ctx.call::<Initialize, _>(meta, RequestParams::Each(req_params), move |ctx, _meta, results| {
+    ctx.call::<Initialize, _>(meta, RequestParams::Each(req_params), move |ctx, meta, results| {
         let results: HashMap<_,_> = results.into_iter().collect();
 
         let to_editor = ctx.to_editor().clone();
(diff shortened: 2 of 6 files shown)


rc/lsp.kak
```diff
@@ -698,25 +699,40 @@ define-command -hidden lsp-did-change -docstring "Notify language server about b
...
-                evaluate-commands %opt{lsp_completion_fragment_start}
-                set-option window lsp_completion_offset %val{cursor_column}
+    try %{
+        try %{
+            # Path 1: Trigger character detection
+            # In insert mode, <a-h> selects the line including trailing newline.
+            # <a-k> keeps the selection only if it ends with a trigger char
+            # before the newline; on success we extract it with <a-:> (ensure
+            # forward), ; (reduce to cursor at the trailing newline), h (back
+            # one character to the trigger char).
```
krobelus, line 715: let's remove comments that only explain what the code is doing, readers can look that up in `:doc keys`

rc/lsp.kak
```diff
@@ -698,25 +699,40 @@ define-command -hidden lsp-did-change -docstring "Notify language server about b
...
+            evaluate-commands -draft %{
+                execute-keys "<a-h><a-k>[%opt{lsp_completion_trigger_regex}]\n\z<ret>"
```
krobelus, line 717: the `\n` makes no sense to me, why would we not allow completions if not at the end of the line?

Also, this logic will succeed if `lsp_completion_trigger_regex` is empty, which means that servers who don't support this feature will almost never enter Path 2.
This seems like a serious logic flaw that needs to be fixed.

rc/lsp.kak
```diff
@@ -698,25 +699,40 @@ define-command -hidden lsp-did-change -docstring "Notify language server about b
...
+                execute-keys "<a-:>;h"
+                set-option window lsp_trigger_character %val{selection}
+            }
+        } catch %{
+            # Path 2: Normal completion (no trigger character or no match)
```
krobelus, line 722: this is a bit surprising. I think that if `lsp_completion_trigger_regex` is non-empty, we should not  need the `%opt{lsp_extra_word_chars}` hack.

Unfortunately the LSP spec is not helpful, on `CompletionOptions.triggerCharacters` it says:

 > /// Characters that make up identifiers don't need to be listed here.

So yeah I guess we still need this fallback if no trigger character is matches

src/capabilities.rs
```diff
@@ -445,10 +446,47 @@ pub fn initialize(meta: EditorMeta, ctx: &mut Context, servers: Vec<ServerId>) {
...
 
+pub fn send_completion_trigger_characters(meta: &EditorMeta, ctx: &Context) {
+    let mut trigger_chars = BTreeSet::new();
+    for &server_id in &meta.servers {
+        let server = ctx.server(server_id);
+        if let Some(caps) = &server.capabilities {
+            if let Some(provider) = &caps.completion_provider {
+                if let Some(chars) = &provider.trigger_characters {
+                    for ch in chars {
+                        if ch.chars().count() == 1 {
```
krobelus, line 462: This seems very weird, why do we need to check for length 1?
Why is triggerCharacters a list of Strings instead of a list of chars?
Is it because javascript/typescript doesn't have a native char type?

If we only expect strings of exactly one character, we should add `assert_eq!(ch.char_count(), 1)` instead.

Else, if a length > 1 string is possibly allowed here, we should rather support this right away. 

we'd only need to adjust the regex to be `(?:%opt{lsp_completion_trigger_regex})`
where `lsp_completion_trigger_regex` is `"|".join(re_escape(tcs) for tcs in trigger_characters)`,
and `re_escape` is simple to implement; we only need to backslash-escape the characters mentioned in `:doc regex`:

 > Every character except the syntax characters \^$.*+?[]{}|(). match themselves.

src/capabilities.rs
```diff
@@ -445,10 +446,47 @@ pub fn initialize(meta: EditorMeta, ctx: &mut Context, servers: Vec<ServerId>) {
...
+            "<" => "<lt>".to_string(),
+            "[" => "\\[".to_string(),
+            "\\" => "\\\\".to_string(),
+            "]" => "\\]".to_string(),
+            "^" => "\\^".to_string(),
+            other => other.to_string(),
+        })
+        .collect();
+    let command = format!(
+        "evaluate-commands -buffer {} -verbatim -- set-option buffer lsp_completion_trigger_regex {}",
```
krobelus, line 483: it would make more sense to include the brackets here (`[{}]`) because that's where we do the escaping.
The default value for `lsp_completion_trigger_regex` can remain the empty string.

src/capabilities.rs
```diff
@@ -445,10 +446,47 @@ pub fn initialize(meta: EditorMeta, ctx: &mut Context, servers: Vec<ServerId>) {
...
+        editor_quote(&meta.buffile),
+        editor_quote(&escaped),
+    );
```
krobelus, line 486: no need to send this command to the editor if `trigger_chars` is empty.

src/controller.rs
```diff
@@ -524,6 +524,14 @@ fn dispatch_fifo_request(
             position: state.next()?,
             completion: EditorCompletion {
                 offset: state.next()?,
+                trigger_character: {
+                    let tc: String = state.next()?;
+                    if tc.is_empty() {
+                        None
+                    } else {
+                        Some(tc)
```
krobelus, line 532: no need for the manual `is_empty()` check, a mere  `trigger_character: state.next()?,` should be enough because we have an `impl FromString for Option<String>`.

PR #884, opened 2026-02-28
Encode brackets for file paths

gbrlsnchs:
Files with brackets in their name were failing due to the lack of encoding. This PR encodes them and uses the new `Uri` struct for `lsp-types`.

### DIFF
--- a/src/text_edit.rs
+++ b/src/text_edit.rs
@@ -2,6 +2,7 @@ use crate::context::*;
 use crate::editor_transport::ToEditorSender;
 use crate::position::*;
 use crate::types::*;
+use crate::util::uri_to_file_path;
 use crate::util::*;
 use indoc::formatdoc;
 #[cfg(test)]
@@ -71,7 +72,7 @@ impl TextEditish<OneOf<TextEdit, AnnotatedTextEdit>> for OneOf<TextEdit, Annotat
 pub fn apply_text_edits<T: TextEditish<T>>(
     server_id: ServerId,
     meta: EditorMeta,
-    uri: Url,
+    uri: Uri,
     edits: Vec<T>,
     ctx: &mut Context,
 ) {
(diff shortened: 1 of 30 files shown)


src/text_edit.rs
```diff
@@ -119,22 +119,21 @@ pub fn apply_text_edits_try_deferred<T: TextEditish<T>>(
...
     server_id: ServerId,
-    uri: &Url,
+    uri: &Uri,
     text_edits: Vec<T>,
     language_id: &LanguageId,
     ctx: &mut Context,
 ) -> std::io::Result<()> {
-    let path = uri.to_file_path().unwrap();
-    let filename = path.to_str().unwrap();
+    let filename = uri_to_file_path(uri).to_string_lossy().into_owned();
```
krobelus, line 134: feels like you made all these additions of `to_string_lossy()` to avoid definining a temporary variable -- but that would not be a good reason

src/text_edit.rs
```diff
@@ -457,9 +456,7 @@ pub fn apply_text_edits_to_buffer<T: TextEditish<T>>(
         );
     }
 
-    let maybe_buffile = uri
-        .and_then(|uri| uri.to_file_path().ok())
-        .and_then(|path| path.to_str().map(|buffile| buffile.to_string()));
+    let maybe_buffile = uri.map(|uri| uri_to_file_path(&uri).to_string_lossy().into_owned());
```
krobelus, line 459: using lossy conversion for a file path seems wrong.
The bigger issue is that we're using String everywhere which can't express non-Unicode data,
but we're probably not gonna be able to fix this anytime soon.
So for now I'll maybe keep the existing behavior when for invalid UTF-8.

src/text_sync.rs
```diff
@@ -226,8 +227,9 @@ fn event_file_changes(
                 | notify::EventKind::Other => continue,
             };
             if watch_request.pattern.matches_path(path) {
+                let path = path.to_str().unwrap();
```
krobelus, line 230: this is unnecessary back-and-forth conversion;
fixing that by changing the signature to `file_path_to_uri(path: impl AsRef<std::path::Path>)`

src/util.rs
```diff
@@ -127,3 +130,18 @@ where
...
+    let s = url.as_str().replace('[', "%5B").replace(']', "%5D");
+    Uri::from_str(&s).unwrap()
+}
+
+/// Parse a file:// URI and return the filesystem path.
+pub fn uri_to_file_path(uri: &Uri) -> PathBuf {
+    url::Url::parse(uri.as_str())
+        .ok()
+        .and_then(|url| url.to_file_path().ok())
+        .unwrap_or_else(|| PathBuf::from(uri.path().as_str()))
```
krobelus, line 146: the catch-all fallback seems weird; why is it necessary at all?
Existing code uses `Uri::to_file_path().unwrap()` everywhere.
I guess the newly added `url::Url::parse` might fail (why though),
but `to_file_path()` should never fail, so we should at least reduce the scope of the fallback.

I'll remove the fallback and merge (sorry for the delay), but LMK if we need it for any reason.

PR #885, opened 2026-03-02
Add support for Deno virtual documents

gbrlsnchs:
> [!IMPORTANT]
> Needs https://github.com/kakoune-lsp/kakoune-lsp/pull/884 first as base.

More details in the commits themselves but, basically, some servers use virtual documents to show definitions. Deno does that because the source of its core is embedded in Deno itself, not in files.

This merge request allows handling those virtual documents by creating temporary files for them. Same path, same hash, which means the same temporary file can be used between different projects.

Also, this merge request also correctly maps those temporary files back to their virtual URI, so that the LSP can still function accordingly inside those files.

The way the code is structured allows easily plugging in virtual document handlers for other LSP servers as well.

### DIFF
--- a/rc/lsp.kak
+++ b/rc/lsp.kak
@@ -2266,10 +2266,15 @@ hook -group lsp-goto global WinSetOption filetype=(lsp-(?:diagnostics|document-s
 define-command -hidden lsp-make-register-relative-to-root %{
     evaluate-commands -save-regs / %{
         try %{
-            # Is it an absolute path?
-            execute-keys <a-k>\A/<ret>
+            # Is it a virtual path?
+            execute-keys <a-k>\A\w+:/<ret>
         } catch %{
-            set-register a "%opt{lsp_project_root}%reg{a}"
+            try %{
+                # Is it an absolute path?
+                execute-keys <a-k>\A/<ret>
+            } catch %{
+                set-register a "%opt{lsp_project_root}%reg{a}"
+            }
         }
     }
 }
@@ -2284,17 +2289,35 @@ define-command -hidden lsp-goto-jump -docstring %{
         try %{
             evaluate-commands -draft -save-regs / %{
                 set-register / ^\h*\K([^:\n]+):(\d+)\b(?::(\d+)\b)?(?::([^\n]+))
-                execute-keys <semicolon>xs<ret>
+                try %{
+                    execute-keys <semicolon>xs<ret>
+                } catch %{
+                    # Try virtual path pattern.
+                    set-register / ^\h*\K(.+):(\d+):(\d+):([^\n]+)
+                    execute-keys <semicolon>xs<ret>
+                }
                 set-register a "%reg{1}"
                 set-register b "%reg{2}"
                 set-register c "%reg{3}"
                 lsp-make-register-relative-to-root
             }
             set-option buffer jump_current_line %val{cursor_line}
-            evaluate-commands -try-client %opt{jumpclient} -verbatim -- edit -existing -- %reg{a} %reg{b} %reg{c}
-            try %{ focus %opt{jumpclient} }
+            try %{
+                lsp-open-file %reg{a} %reg{b} %reg{c}
+            } catch %{
+                lsp-open-virtual-file %reg{a} %reg{b} %reg{c} %reg{b} %reg{c}
+            }
         }
     }
+}
+
+define-command -hidden lsp-open-file -params 3 %{
+    evaluate-commands -try-client %opt{jumpclient} -verbatim -- edit -existing -- %arg{@}
+    try %{ focus %opt{jumpclient} }
+}
+
+define-command -hidden lsp-open-virtual-file -params 5 %{
+    evaluate-commands -try-client %opt{jumpclient} -- lsp-do-send kakoune/open-virtual-file %arg{@}
 }
 
 define-command -hidden lsp-document-symbol-jump -docstring %{
--- a/src/context.rs
+++ b/src/context.rs
@@ -2,6 +2,7 @@ use crate::editor_transport::{self, ToEditorSender};
 use crate::language_server_transport::LanguageServerTransport;
 use crate::text_sync::CompiledFileSystemWatcher;
 use crate::thread_worker::Worker;
+use crate::util::file_path_to_uri;
 use crate::{filetype_to_language_id_map, types::*};
 use jsonrpc_core::{self, Call, Error, Failure, Id, Output, Success, Value, Version};
 use lsp_types::notification::{Cancel, Notification};
@@ -13,6 +14,7 @@ use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, HashMap, HashSet, VecDeque};
 use std::convert::TryInto;
 use std::path::PathBuf;
+use std::str::FromStr;
 use std::time;
 
 // Copy of Kakoune's timestamped buffer content.
@@ -78,6 +80,7 @@ pub struct Context {
     pub config: Config,
     pub diagnostics: HashMap<String, Vec<(ServerId, Diagnostic)>>,
     pub documents: HashMap<String, Document>,
+    pub virtual_documents: HashSet<String>,
     pub dynamic_config: DynamicConfig,
     pub inlay_hints: HashMap<String, Vec<(ServerId, InlayHint)>>,
     pub language_servers: BTreeMap<ServerId, ServerSettings>,
@@ -118,6 +121,7 @@ impl Context {
             config,
             diagnostics: Default::default(),
             documents: Default::default(),
+            virtual_documents: Default::default(),
             dynamic_config: DynamicConfig::default(),
             inlay_hints: Default::default(),
             language_servers: BTreeMap::new(),
@@ -419,6 +423,13 @@ impl Context {
         message: impl AsRef<str>,
     ) {
         editor_transport::show_error(&self.to_editor, meta, response_fifo, message);
+    }
+
+    pub fn uri_for_buffer(&self, path: &str) -> Uri {
+        match self.virtual_documents.contains(path) {
+            true => Uri::from_str(path).unwrap(),
+            false => file_path_to_uri(path),
+        }
     }
 
     fn next_batch_id(&mut self) -> BatchNumber {
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -21,7 +21,6 @@ use crate::language_features::lean::{
 };
 use crate::language_features::{selection_range, *};
 use crate::log::DEBUG;
-use crate::progress;
 use crate::project_root::find_project_root;
 use crate::show_message::{self, MessageRequestResponse};
 use crate::text_sync::*;
@@ -34,6 +33,7 @@ use crate::workspace::{
 use crate::{context::*, set_logger};
 use crate::{diagnostics, do_cleanup};
 use crate::{language_server_transport, LAST_CLIENT};
+use crate::{progress, util};
 use ccls::{EditorCallParams, EditorInheritanceParams, EditorMemberParams, EditorNavigateParams};
 use code_lens::{text_document_code_lens, CodeLensOptions};
 use crossbeam_channel::{after, never, tick, Receiver, Select, Sender};
@@ -491,6 +491,25 @@ fn dispatch_fifo_request(
             debug!(&state.to_editor, "Applied option change {}", hook_param);
             return Some(());
         }
+        "kakoune/open-virtual-file" => Box::new({
+            let uri: String = state.next()?;
+            let start_line: u32 = state.next()?;
+            let start_column: u32 = state.next()?;
+            let end_line: u32 = state.next().or(Some(start_line))?;
+            let end_column: u32 = state.next().or(Some(start_column))?;
+
+            OpenVirtualFileParams {
+                uri,
+                start: KakounePosition {
+                    line: start_line,
+                    column: start_column,
+                },
+                end: KakounePosition {
+                    line: end_line,
+                    column: end_column,
+                },
+            }
+        }),
         "rust-analyzer/expandMacro" => Box::new(PositionParams {
             position: state.next()?,
         }),
@@ -1329,6 +1348,9 @@ fn route_request(
     if request_method == notification::Exit::METHOD {
         return None;
     }
+    if request_method == "kakoune/open-virtual-file" {
+        return None;
+    }
     if !meta.session.is_empty() && &meta.session != ctx.session() {
         info!(
             ctx.to_editor(),
@@ -1338,7 +1360,7 @@ fn route_request(
         );
         return Some(ControlFlow::Break(()));
     }
-    if !meta.buffile.starts_with('/') {
+    if !meta.buffile.starts_with('/') && !ctx.virtual_documents.contains(&meta.buffile) {
         report_error_no_server_configured(
             ctx,
             meta,
@@ -1653,6 +1675,7 @@ fn dispatch_incoming_editor_request(request: EditorRequest, ctx: &mut Context) -
             // InsertIdle is not triggered while the completion pager is active, so let's
             // smuggle completion-related requests through.
             && method != request::ResolveCompletionItem::METHOD
+            && method != "kakoune/open-virtual-file"
     {
         assert!(request.response_fifo.is_none());
         // Wait for buffer update.
@@ -1838,6 +1861,13 @@ fn dispatch_editor_request(request: EditorRequest, ctx: &mut Context) -> Control
         }
         "kakoune/textDocument/codeLens" => {
             code_lens::resolve_and_perform_code_lens(meta, params.unbox(), ctx);
+        }
+        "kakoune/open-virtual-file" => {
+            let OpenVirtualFileParams { uri, start, end } = params.unbox();
+            if let Some(contents) = ctx.documents.get(&uri) {
+                let range = KakouneRange { start, end };
+                util::open_virtual_file(ctx, meta, &uri, range, &contents.text.to_string());
+            }
         }
         request::Formatting::METHOD => {
             formatting::text_document_formatting(meta, response_fifo, params.unbox(), ctx);
@@ -2136,7 +2166,11 @@ fn ensure_did_open(request: &EditorRequest, ctx: &mut Context) {
         return;
     };
     if !buffile.starts_with('/') {
-        assert_eq!(&request.method, notification::Exit::METHOD);
+        assert!(
+            request.method == notification::Exit::METHOD
+                || &request.method == "kakoune/open-virtual-file"
+                || ctx.virtual_documents.contains(buffile)
+        );
         return;
     }
     if request.method == notification::DidChangeTextDocument::METHOD {
--- a/src/language_features/deno.rs
+++ b/src/language_features/deno.rs
@@ -1,0 +1,81 @@
+use std::collections::HashMap;
+
+use itertools::Itertools;
+use lsp_types::{request::Request, Location, TextDocumentIdentifier, Uri};
+use ropey::Rope;
+
+use crate::{
+    context::{Context, Document, RequestParams},
+    language_features::goto::{goto_location, goto_locations},
+    types::{EditorMeta, ServerId},
+};
+
+pub const SCHEME: &str = "deno";
+
+struct VirtualTextDocument {}
+
+#[derive(Deserialize, Serialize, Clone)]
+#[serde(rename_all = "camelCase")]
+struct VirtualTextDocumentParams {
+    text_document: TextDocumentIdentifier,
+}
+
+impl Request for VirtualTextDocument {
+    type Params = VirtualTextDocumentParams;
+    type Result = String;
+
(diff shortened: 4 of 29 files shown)


rc/lsp.kak
```diff
@@ -664,6 +664,10 @@ hook -group lsp-scratch-buffers global WinDisplay \*debug\* %{
     lsp-block-in-buffer
 }
 
+hook -group lsp-virtual-documents global BufCreate .*/kakoune-lsp/.*/virtual/.*/[0-9a-f]+\..* %{
```
krobelus: why do we need to write this to a temporary file, as opposed to creating a scratch buffer?
We currently hard-disable all LSP interaction for scratch buffers but we could allow it scratch buffers with names matching `\*lsp-.*`.
Note that `:edit` works for scratch buffers just fine.

src/language_features/deno.rs
```diff
@@ -0,0 +1,134 @@
...
+                .zip(results.into_iter())
+                .map(|((_, uri), (_, content))| (uri.clone(), content))
+                .collect();
+
+            // Write temp files for each unique URI.
+            let tmp_paths: HashMap<&str, PathBuf> = unique_uris
+                .iter()
+                .map(|(_, uri)| {
+                    let content = content_map.get(uri).unwrap();
+                    let tmp_path = util::create_virtual_definition_file("deno", uri, content);
```
krobelus: it's surprising that we never remove these "temporary" paths.

src/language_features/deno.rs
```diff
@@ -0,0 +1,134 @@
...
+                let (server_id, Location { uri, range }) = &locations[0];
+                let content = content_map.get(uri).unwrap();
+                let tmp_path = tmp_paths.get(uri.as_str()).unwrap();
+
+                // Select range that the server wants to show.
+                let text = Rope::from_str(content);
+                let server = ctx.server(*server_id);
+                let kak_range = lsp_range_to_kakoune(range, &text, server.offset_encoding);
+                let command = format!(
+                    "evaluate-commands -try-client %opt{{jumpclient}} -- %[
```
krobelus: forgot the backslash at the end of the line here?
Also we usually use `formatdoc!` for these I guess..
which made me realize.. why not reuse the `edit_at_range` function?

src/language_features/deno.rs
```diff
@@ -0,0 +1,134 @@
...
+                        edit -existing {}; \
+                        select {}; \
+                        execute-keys <c-s>vv \
+                    ]",
+                    editor_quote(&tmp_path.to_string_lossy()),
```
krobelus: don't use lossy conversion for path names. Better to crash instead of doing the wrong thing (until we don't handle this properly)

src/language_features/deno.rs
```diff
@@ -0,0 +1,134 @@
...
+                    let content = content_map.get(uri).unwrap();
+                    let rope = Rope::from_str(content);
+                    let server = ctx.server(*server_id);
+                    let pos = lsp_range_to_kakoune(range, &rope, server.offset_encoding).start;
+                    if range.start.line as usize >= rope.len_lines() {
+                        return "".into();
+                    }
+                    let tmp_path = tmp_paths.get(uri.as_str()).unwrap();
+                    format!(
+                        "{}:{}:{}:{}",
```
krobelus: similar code exists in `src/language_features/goto.rs`.
Please consider extracting a function rather than introducing a code clone.

In fact, is there any reason we can't reuse this code snippet, with small modifications?

```rust
    match locations.len() {
        0 => {}
        1 => {
            goto_location(meta, &locations[0], ctx);
        }
        _ => {
            goto_locations(meta, &locations, ctx);
        }
    }
```

src/language_features/goto.rs
```diff
@@ -74,32 +74,60 @@ pub fn edit_at_range(buffile: &str, range: KakouneRange, in_normal_mode: bool) -
...
-        let server = ctx.server(*server_id);
-        let range = lsp_range_to_kakoune(range, &contents, server.offset_encoding);
+fn goto_location(meta: EditorMeta, (server_id, location): (ServerId, Location), ctx: &mut Context) {
+    let Location { uri, range } = location;
+
+    // Handle virtual location.
+    {
+        if uri
+            .scheme()
+            .map_or(false, |scheme| scheme.eq_lowercase(deno::SCHEME))
```
krobelus: Use `is_some_and`. Same below

src/language_features/goto.rs
```diff
@@ -74,32 +74,60 @@ pub fn edit_at_range(buffile: &str, range: KakouneRange, in_normal_mode: bool) -
...
+        let (deno_locations, _): (Vec<_>, Vec<_>) =
+            virtual_locations
+                .into_iter()
+                .partition(|(_, Location { uri, .. })| {
+                    uri.scheme()
+                        .map_or(false, |scheme| scheme.eq_lowercase(deno::SCHEME))
+                });
+        deno::handle_virtual_locations(&meta, ctx, deno_locations);
+
+        // NOTE: Handlers for other languages should be added here.
```
krobelus: I think other schemes are impossible today, so maybe we should panic upon seeing them here?

src/util.rs
```diff
@@ -127,3 +131,38 @@ where
...
+        .ok()
+        .and_then(|url| url.to_file_path().ok())
+        .unwrap_or_else(|| PathBuf::from(uri.path().as_str()))
+}
+
+/// Creates a temporary file for a virtual defintion (for example, used by Deno LSP, where some
+/// defitinions are not in disk when using goto).
+pub fn create_virtual_definition_file(name: &str, uri: &Uri, content: &str) -> PathBuf {
+    let path = uri_to_file_path(uri);
+    let ext = path.extension().and_then(|p| p.to_str()).unwrap_or("ts");
```
krobelus: ts as in typescript? I don't think `src/util.rs` should contain language-specific or server-specific code.
Either remove the fallback or panic if no extension is given.

src/util.rs
```diff
@@ -127,3 +131,38 @@ where
...
+    let mut hasher = DefaultHasher::new();
+    uri.as_str().hash(&mut hasher);
+    let hash = hasher.finish();
+    let mut tmp_path = temp_dir();
+
+    tmp_path.push("virtual");
+    tmp_path.push(name);
+    let _ = std::fs::create_dir_all(&tmp_path);
+
+    tmp_path.push(format!("{:x}.{}", hash, ext));
```
krobelus: not sure if hashing is appropriate here, maybe it's better to use the input name except replace dangerous characters (like `/`) with some multi-char escape sequence.
This should make sure there is no collision (i.e. the escaping function is injective).

gbrlsnchs:
Thank you for the review, I will rebase and address your feedback. :-)

rc/lsp.kak
gbrlsnchs: I have to use a FIFO buffer, right? In order to write the content.

gbrlsnchs: Also: there are scenarios where there are multiple options that end up triggering a goto prompt. How would I handle that while also using a scratch buffer?

krobelus:
> I have to use a FIFO buffer, right? In order to write the content.

no, see lsp-show-goto-buffer for an example.

krobelus:
but fifo buffer is useful if we may edit a buffer that may already
be open, because in that case, "edit! -fifo" will maintain the buffer
visible in whatever client its currently shown.  Though this is quite
a subtle quirk, it would be nice if we made this explicit (but that
would require a change in Kakoune)

rc/lsp.kak
krobelus: > Also: there are scenarios where there are multiple options that end up triggering a goto prompt. How would I handle that while also using a scratch buffer?

I'm not sure I understand the problem. We can still use `edit` or `buffer` to switch to an existing buffer if it already exists (otherwise buffer creation would fail if we reuse the same name)

rc/lsp.kak
```diff
@@ -2266,10 +2266,15 @@ hook -group lsp-goto global WinSetOption filetype=(lsp-(?:diagnostics|document-s
 define-command -hidden lsp-make-register-relative-to-root %{
     evaluate-commands -save-regs / %{
         try %{
-            # Is it an absolute path?
-            execute-keys <a-k>\A/<ret>
+            # Is it a virtual path?
+            execute-keys <a-k>\A\w+:/<ret>
```
krobelus, line 2270: since we currently only support deno, can we hardcode `deno:` instead of `\w+:`?
And maybe fold this into a single "absolute path" check (assuming all deno URIs are absolute)

	execute-keys <a-k>\a(?:deno://)?/<ret>

This check will not recognize relative deno paths like `deno://some/relative/path` but those are probably not allowed,
(or not even valid URIs?) so it seems like a GIGO scenario where it's okay if we do weird things.

rc/lsp.kak
```diff
@@ -2284,19 +2289,37 @@ define-command -hidden lsp-goto-jump -docstring %{
         try %{
             evaluate-commands -draft -save-regs / %{
                 set-register / ^\h*\K([^:\n]+):(\d+)\b(?::(\d+)\b)?(?::([^\n]+))
-                execute-keys <semicolon>xs<ret>
+                try %{
+                    execute-keys <semicolon>xs<ret>
+                } catch %{
+                    # Try virtual path pattern.
+                    set-register / ^\h*\K(.+):(\d+):(\d+):([^\n]+)
```
krobelus, line 2296: hmm. I wonder if it's possible to update the original regex instead, in a way that is simpler overall.
That might help with updating `lsp-{next,previous}-location`

rc/lsp.kak
```diff
@@ -2284,19 +2289,37 @@ define-command -hidden lsp-goto-jump -docstring %{
...
                 set-register c "%reg{3}"
                 lsp-make-register-relative-to-root
             }
             set-option buffer jump_current_line %val{cursor_line}
-            evaluate-commands -try-client %opt{jumpclient} -verbatim -- edit -existing -- %reg{a} %reg{b} %reg{c}
-            try %{ focus %opt{jumpclient} }
+            try %{
+                lsp-open-file %reg{a} %reg{b} %reg{c}
+            } catch %{
+                lsp-open-virtual-file %reg{a} %reg{b} %reg{c} %reg{b} %reg{c}
```
krobelus, line 2308: no call to `focus` in the virtual file case?
I think we could maybe hardcode it in Rust sources for now
(after `lsp-unblock-in-buffer;`). Or define a callback function.

rc/lsp.kak
```diff
@@ -2284,19 +2289,37 @@ define-command -hidden lsp-goto-jump -docstring %{
...
+            }
         }
     }
 }
 
+define-command -hidden lsp-open-file -params 3 %{
+    evaluate-commands -try-client %opt{jumpclient} -verbatim -- edit -existing -- %arg{@}
```
krobelus, line 2315: missing `-verbatim` (or quoting) to avoid double evaluation (if arg contains % expansions or similar).

same below

rc/lsp.kak
```diff
@@ -2284,19 +2289,37 @@ define-command -hidden lsp-goto-jump -docstring %{
...
+    try %{ focus %opt{jumpclient} }
+}
+
+define-command -hidden lsp-open-virtual-file -params 5 %{
+    evaluate-commands -try-client %opt{jumpclient} -- lsp-do-send kakoune/open-virtual-file %arg{@}
```
krobelus, line 2320: hmm normally we don't call `lsp-do-send` directly, we always go through `lsp-send`.
But I guess this is the most convenient way to make things work from a scratch buffer.

src/context.rs
```diff
@@ -421,6 +425,13 @@ impl Context {
         editor_transport::show_error(&self.to_editor, meta, response_fifo, message);
     }
 
+    pub fn uri_for_buffer(&self, path: &str) -> Uri {
```
krobelus, line 428: `path` is a misleading name; this should rather be `buffile` or `bufname`.

Note that in recent Kakoune versions, `buffile` is empty for scratch buffers,
so we need to make sure to fall back to `bufname`.. not sure why we don't do that already.

I guess it's because `"".starts_with('/')` is false..
but note that `ctx.virtual_documents.contains(&meta.buffile)` will easily and incorrectly be true for **all** scratch buffers because `meta.buffile` is empty for scratch buffers.

So you want to use `%val{bufname}`, not `%val{buffile}`, so make sure to pass that around.

src/controller.rs
```diff
@@ -491,6 +491,25 @@ fn dispatch_fifo_request(
             debug!(&state.to_editor, "Applied option change {}", hook_param);
             return Some(());
         }
+        "kakoune/open-virtual-file" => Box::new({
+            let uri: String = state.next()?;
+            let start_line: u32 = state.next()?;
+            let start_column: u32 = state.next()?;
+            let end_line: u32 = state.next().or(Some(start_line))?;
+            let end_column: u32 = state.next().or(Some(start_column))?;
```
krobelus, line 499: why add separate parameters for `end_line`/`end_column`, is anyone really using them?
`edit` doesn't support them either.. its paramaters are only for simple use cases.
For flexibility, it's best to run other commands after "edit".
In case of `lsp-open-virtual-file` -- should we ever need it -- that might need a hook or callback argument

src/controller.rs
```diff
@@ -1329,6 +1348,9 @@ fn route_request(
     if request_method == notification::Exit::METHOD {
         return None;
     }
+    if request_method == "kakoune/open-virtual-file" {
```
krobelus, line 1351: extract a constant, since this is used in a few places?

src/controller.rs
```diff
@@ -1839,6 +1862,13 @@ fn dispatch_editor_request(request: EditorRequest, ctx: &mut Context) -> Control
         "kakoune/textDocument/codeLens" => {
             code_lens::resolve_and_perform_code_lens(meta, params.unbox(), ctx);
         }
+        "kakoune/open-virtual-file" => {
+            let OpenVirtualFileParams { uri, start, end } = params.unbox();
+            if let Some(contents) = ctx.documents.get(&uri) {
+                let range = KakouneRange { start, end };
+                util::open_virtual_file(ctx, meta, &uri, range, &contents.text.to_string());
```
krobelus, line 1869: it's unusual to put any logic in this switch statement.
Why not put the `ctx.documents.get` check inside `open_virtual_file`?

src/controller.rs
```diff
@@ -2136,7 +2166,11 @@ fn ensure_did_open(request: &EditorRequest, ctx: &mut Context) {
         return;
     };
     if !buffile.starts_with('/') {
-        assert_eq!(&request.method, notification::Exit::METHOD);
+        assert!(
+            request.method == notification::Exit::METHOD
+                || &request.method == "kakoune/open-virtual-file"
```
krobelus, line 2171: oh the `&` is unnecessary

src/language_features/deno.rs
```diff
@@ -0,0 +1,81 @@
...
+                m
+            });
+
+    ctx.call::<VirtualTextDocument, _>(
+        meta.clone(),
+        RequestParams::Each(req_params),
+        move |ctx, _, results| {
+            for ((_, uri), (_, content)) in unique_uris.into_iter().zip(results.into_iter()) {
+                _ = ctx.virtual_documents.insert(uri.to_string());
+                _ = ctx.documents.insert(
```
krobelus, line 63: shouldn't we replace the document (with new content) if it already exists.
Insert does nothing in that case.

src/language_features/goto.rs
```diff
@@ -73,31 +74,98 @@ pub fn edit_at_range(buffile: &str, range: KakouneRange, in_normal_mode: bool) -
...
+                        uri.scheme()
+                            .is_some_and(|scheme| scheme.eq_lowercase(deno::SCHEME))
+                    });
+
+            return deno::handle_virtual_locations(
+                meta,
+                ctx,
+                deno_locations,
+                non_virtual_locations,
+            );
```
krobelus, line 146: I don't really understand why `goto_locations` needs to know about the `virtual_locations` parameter.
Looks like there is only one caller that actually passes non-empty virtual locations,
and in that case we typically pass back control to `deno::handle_virtual_locations`.
So how about handling `deno://` virtual locations in the caller (to keep deno-specific code away from `goto.rs`),
and passing any `other://` virtual locations via the regular `locations` parameter?

src/language_features/goto.rs
```diff
@@ -73,31 +74,98 @@ pub fn edit_at_range(buffile: &str, range: KakouneRange, in_normal_mode: bool) -
...
+
+        locations = non_virtual_locations;
+    } else {
+        locations.extend(virtual_locations);
+    }
+
     let select_location = locations
         .iter()
-        .chunk_by(|(_, Location { uri, .. })| uri_to_file_path(uri))
+        .chunk_by(|(_, Location { uri, .. })| (uri_to_file_path(uri), uri))
```
krobelus, line 156: Chunking by the pair `(uri_to_file_path(uri), uri)` is weird.
Better chunk by `uri` (which is equivalent), and move the file path computation into the map callback.

src/language_features/goto.rs
```diff
@@ -73,31 +74,98 @@ pub fn edit_at_range(buffile: &str, range: KakouneRange, in_normal_mode: bool) -
...
+                .scheme()
+                .is_some_and(|scheme| scheme.eq_lowercase("file"));
+
+            let path_str = if is_file {
+                path.to_str().unwrap()
+            } else {
+                uri.as_str()
+            };
+
             let contents = match get_file_contents(path_str, ctx) {
```
krobelus, line 169: get_file_contents doesn't support virtual file URIs yet so it's a bit strange that we're calling it here, but I guess it's fine

src/language_features/goto.rs
```diff
@@ -109,11 +177,21 @@ fn goto_locations(meta: EditorMeta, locations: &[(ServerId, Location)], ctx: &mu
                     if range.start.line as usize >= contents.len_lines() {
                         return "".into();
                     }
+
+                    let file_path = if uri
+                        .scheme()
+                        .is_some_and(|scheme| scheme.eq_lowercase("file"))
```
krobelus, line 183: this ought to reuse above definition of `is_file`.

src/language_features/goto.rs
```diff
@@ -109,11 +177,21 @@ fn goto_locations(meta: EditorMeta, locations: &[(ServerId, Location)], ctx: &mu
...
+                    {
+                        short_file_path(path_str, ctx.main_root(&meta))
+                    } else {
+                        uri.as_str()
```
krobelus, line 187: I guess you updated `lsp-goto-jump` to handle URIs in the first component,
but won't this also break assumptions in `lsp-select-next`, `lsp-select-previous`?

src/util.rs
```diff
@@ -154,3 +156,34 @@ pub fn uri_to_file_path(uri: &Uri) -> PathBuf {
         .to_file_path()
         .unwrap()
 }
+
+pub fn open_virtual_file(
```
krobelus, line 160: `util` is not a great module name; maybe we should put this in `deno.rs` for now..

src/util.rs
```diff
@@ -154,3 +156,34 @@ pub fn uri_to_file_path(uri: &Uri) -> PathBuf {
...
+    meta: EditorMeta,
+    uri: &str,
+    range: KakouneRange,
+    contents: &str,
+) {
+    let fifo = mkfifo(ctx.to_editor());
+    let client = meta.client.as_ref().unwrap();
+    let command = format!(
+        "%[
+            edit! -readonly -fifo {fifo} {uri}; \
```
krobelus, line 171: this quoting would be broken if `uri` contained `]`.
Maybe better to use `editor_quote` or `editor_escape`.

src/util.rs
```diff
@@ -154,3 +156,34 @@ pub fn uri_to_file_path(uri: &Uri) -> PathBuf {
...
+            hook -once buffer BufCloseFifo .* %[
```
krobelus, line 172: why use a fifo buffer instead of a plain scratch buffer (like `lsp-show-goto-buffer` does)?
I forgot..

src/util.rs
```diff
@@ -154,3 +156,34 @@ pub fn uri_to_file_path(uri: &Uri) -> PathBuf {
...
+                nop %sh[ rm {fifo} ]; \
+                evaluate-commands -try-client {client} %[
+                    edit -existing -readonly {uri}
```
krobelus, line 175: readonly is an unnecessary restriction.
No need to second guess the user.
The only reason I can think of for using `-readonly ` is the built-in `:doc` command,
whose highlighters would be broken by the user editing the buffer.
But that's not a problem here.

src/util.rs
```diff
@@ -154,3 +156,34 @@ pub fn uri_to_file_path(uri: &Uri) -> PathBuf {
...
+                    select {}; \
+                    execute-keys <c-s>vv; \
+                    lsp-unblock-in-buffer; \
```
krobelus, line 178: I wonder what if anything will break when using virtual documents.
I didn't test it yet; I guess we'll see.
I guess something like `lsp-definition` is supposed to work inside these buffers?

PR #886, opened 2026-03-03
feat: sourcekit lsp support for Swift filetype

paaloeye:
- added sourcekit-lsp configuration for Swift filetype
- root detection includes Package.swift and .xcodeproj files


krobelus:
thanks

PR #890, opened 2026-05-09
Replace nil with nixd as default nix LSP

HeitorAugustoLN:
[nixd](https://github.com/nix-community/nixd) is a more active and feature rich language server for Nix, most Nix users already use it for a while now


krobelus:
okay. Helix enables both `nixd` and `nil`; in future we should do that too but need some code changes to make sure that it's not an error if one language server is missing

HeitorAugustoLN:
> okay. Helix enables both `nixd` and `nil`; in future we should do that too but need some code changes to make sure that it's not an error if one language server is missing

~~I just found https://github.com/joaotavora/rassumfrassum, that might be useful for this, I am using this for eglot in emacs~~

Nevermind, it fails when a language server is missing

PR #897, opened 2026-08-02
Handle missing cases of TextDocumentSyncSaveOptions

natsukagami:
`include_text` with:
- a missing `include_text` should default to `false`
- just `Supported(true)` should default to `false`

This was causing some troubles with `rust-analyzer`, where I think doesn't include `include_text` as a setting,
causing us to always skip sending `didSave` notifications.


krobelus:
sorry for the long delay; with commit 2c44d2847 (Refactor didSave capability check, 2026-06-24) I only wanted to get rid of the `options` local variable, not change any behavior.. my bad. Thanks!

PR #899, opened 2026-08-17
Support LSP diagnostic pull model (textDocument/diagnostic)

Yukaii:
Hi! This PR implements the LSP 3.17+ diagnostic pull model (`textDocument/diagnostic`) alongside `workspace/diagnostic/refresh`.

### Summary
- Advertises `textDocument.diagnostic` and `workspace.diagnostics.refreshSupport` capabilities.
- When a language server advertises `diagnosticProvider`, `kak-lsp` pulls diagnostics on buffer open/change instead of expecting `textDocument/publishDiagnostics`.
- Caches opaque `resultId`s per server and buffer, supporting `Full` and `Unchanged` reports as well as related documents.
- Falls back cleanly to push diagnostics if the server doesn't support pull (e.g. `clangd`, `oxlint`).
- Handles `workspace/diagnostic/refresh` by clearing cached `resultId`s and re-pulling.
- Updates ESLint workaround to inject `nodePath: null` (required by modern `vscode-eslint-language-server`).
- Adds an integration test against `vscode-eslint-language-server`.

### How to test manually

1. Install modern ESLint server:
   ```sh
   npm install -g vscode-langservers-extracted eslint @eslint/js
   ```

2. Create a test project:
   ```sh
   mkdir /tmp/pull-test && cd /tmp/pull-test
   npm init -y
   cat > eslint.config.js << 'EOF'
   import js from "@eslint/js";
   export default [js.configs.recommended, { rules: { "no-unused-vars": "error" } }];
   EOF
   echo 'const unused = 1;' > main.js
   ```

3. In your `kakrc`:
   ```kak
       [eslint-language-server]
       root_globs = ["package.json", "eslint.config.js"]
       command = "vscode-eslint-language-server"
       args = ["--stdio"]
       workaround_eslint = true
       [eslint-language-server.settings]
       workingDirectory.mode = "auto"
   }
   ```

4. Open `main.js` in Kakoune:
   - You should see the error gutter flag on line 1.
   - Deleting `unused` or using it will clear the gutter diagnostics via pull.


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## Unreleased
 
+Additions:
```
krobelus, line 3: FWIW a previous approach was https://github.com/kakoune-lsp/kakoune-lsp/pull/865, which was not finished because it duplicated some code.
Your PR does not seem to suffer from that code duplication.

Looks pretty good. Test looks nice too.
Let me know if you need help addressing comments.

rc/servers.kak
```diff
@@ -293,10 +294,12 @@ hook -group lsp-filetype-javascript global BufSetOption filetype=(?:javascript|t
     #     rulesCustomizations = []
     #     run = "onType"
     #     validate = "on"
+    #     packageManager = "npm"
```
krobelus, line 297: Ok these are probably server defaults(?).
We should maybe also include a link to where these settings are documented
(for example `# See https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/main/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/preferences/Preferences.java`)
but of course that's unrelated to this PR, so it's not a blocker

src/controller.rs
```diff
@@ -2075,6 +2078,32 @@ fn dispatch_server_request(
...
+                    .language_servers
+                    .keys()
+                    .copied()
+                    .filter(|&id| id == server_id)
+                    .collect();
+                if servers.is_empty() {
+                    continue;
+                }
+                let mut pull_meta = meta.clone();
+                pull_meta.buffile = buffile.clone();
```
krobelus: this clone can be avoided either by replacing `&buffile` with `&pull_meta.buffile` below, or by moving it into pull_meta later.

src/controller.rs
```diff
@@ -2075,6 +2078,32 @@ fn dispatch_server_request(
...
+                pull_meta.servers = servers;
+                if let Some(doc) = ctx.documents.get(&buffile) {
+                    pull_meta.version = doc.version;
```
krobelus: but version will be wrong when the document is not present?
Also meta.filetype and meta.language_id are wrong.
Can we construct a mostly empty EditorMeta and only set what's needed?
I think we will only need to set very few fields.

src/controller.rs
```diff
@@ -2075,6 +2078,32 @@ fn dispatch_server_request(
...
+                }
+                // Invalidate cached result ids so the server recomputes.
+                ctx.diagnostic_pull_result_ids
+                    .retain(|&(sid, ref path), _| !(sid == server_id && path == &buffile));
```
krobelus: since we're iterating over all buffiles, why not do

	ctx.diagnostic_pull_result_ids
	    .retain(|&(sid, _), _| sid != server_id);

outside the loop instead?

This should be effectively the same, I doubt we'll find a case where the difference matters.

src/diagnostics.rs
```diff
@@ -7,33 +10,185 @@ use crate::util::*;
...
+    let caps = server.capabilities.as_ref()?;
+    match caps.diagnostic_provider.as_ref()? {
+        DiagnosticServerCapabilities::Options(options) => options.identifier.clone(),
+        DiagnosticServerCapabilities::RegistrationOptions(options) => {
+            options.diagnostic_options.identifier.clone()
+        }
+    }
+}
+
+pub fn server_supports_pull_diagnostics(
```
krobelus: I'd probably inline this function given that it's only used once

src/diagnostics.rs
```diff
@@ -7,33 +10,185 @@ use crate::util::*;
...
+                    partial_result_params: Default::default(),
+                }],
+            )
+        })
+        .collect();
+    if eligible_servers.is_empty() {
+        return;
+    }
+
+    let buffile = meta.buffile.clone();
```
krobelus, line 55: unnecessary clone

src/diagnostics.rs
```diff
@@ -7,33 +10,185 @@ use crate::util::*;
...
+    let version = meta.version;
```
krobelus: maybe inline this variable?
We can use `meta.version` in the callback to the same effect, as far as I can tell

src/diagnostics.rs
```diff
@@ -7,33 +10,185 @@ use crate::util::*;
...
+        meta,
+        RequestParams::Each(eligible_servers.into_iter().collect()),
+        move |ctx, meta, results| {
+            // Drop stale responses after the buffer changed again.
+            if ctx
+                .documents
+                .get(&buffile)
+                .map(|doc| doc.version)
+                .unwrap_or(-1)
+                != version
```
krobelus: The `unwrap_or(-1)` is weird; we probably want `documents.get(...).map(...) != Some(version)`,
so the result is obviously false if the document is missing

src/diagnostics.rs
```diff
@@ -7,33 +10,185 @@ use crate::util::*;
...
+            {
+                return;
+            }
+            for (server_id, result) in results {
+                handle_document_diagnostic_result(server_id, &buffile, result, ctx);
+            }
+            let _ = meta;
```
krobelus: this line seems redundant

src/diagnostics.rs
```diff
@@ -7,33 +10,185 @@ use crate::util::*;
...
+
+fn handle_document_diagnostic_result(
+    server_id: ServerId,
+    buffile: &str,
+    result: DocumentDiagnosticReportResult,
+    ctx: &mut Context,
+) {
+    let report = match result {
+        DocumentDiagnosticReportResult::Report(report) => report,
+        DocumentDiagnosticReportResult::Partial(_) => return,
```
krobelus: if the server is not supposed to send partial results (since we don't support them), then it would be nice to log a warning here, but no big deal if not

src/diagnostics.rs
```diff
@@ -7,33 +10,185 @@ use crate::util::*;
...
+    };
+
+    match report {
+        DocumentDiagnosticReport::Full(report) => {
+            if let Some(result_id) = &report.full_document_diagnostic_report.result_id {
+                ctx.diagnostic_pull_result_ids
+                    .insert((server_id, buffile.to_string()), result_id.clone());
```
krobelus: should not need this clone, I think we can instead do partial moves in a way that Rust understands to be safe: 

	let (result_id, items) = (report.full_document_diagnostic_report.result_id,
			report.full_document_diagnostic_report.items);

src/diagnostics.rs
```diff
@@ -7,33 +10,185 @@ use crate::util::*;
...
+                }
+            }
+        }
+        DocumentDiagnosticReport::Unchanged(report) => {
+            ctx.diagnostic_pull_result_ids.insert(
+                (server_id, buffile.to_string()),
+                report
+                    .unchanged_document_diagnostic_report
+                    .result_id
+                    .clone(),
```
krobelus: unnecessary clone I think

src/diagnostics.rs
```diff
@@ -7,33 +10,185 @@ use crate::util::*;
...
+    report: DocumentDiagnosticReportKind,
+    ctx: &mut Context,
+) {
+    let path = uri_to_file_path(&uri);
+    let path = path.to_str().unwrap();
+    match report {
+        DocumentDiagnosticReportKind::Full(report) => {
+            if let Some(result_id) = &report.result_id {
+                ctx.diagnostic_pull_result_ids
+                    .insert((server_id, path.to_string()), result_id.clone());
```
krobelus: same here, no need to clone, at least if we use a destructuring assignment (`let (result_id, items) = (report.result_id, result.items);`)

src/diagnostics.rs
```diff
@@ -7,33 +10,185 @@ use crate::util::*;
...
         return;
     }
-    let document = document.unwrap();
+    let params: PublishDiagnosticsParams = params.parse().expect("Failed to parse params");
+    let buffile = uri_to_file_path(&params.uri);
+    let buffile = buffile.to_str().unwrap();
+    set_server_diagnostics(server_id, buffile, params.diagnostics, ctx);
+}
+
+pub fn refresh_diagnostics_display(buffile: &str, ctx: &mut Context) {
```
krobelus: this need not be `pub` since it's only used from this file

src/text_sync.rs
```diff
@@ -67,8 +69,9 @@ pub fn text_document_did_change(
 
     // Resets metadata for buffer.
     ctx.documents.insert(meta.buffile.clone(), document);
-    ctx.diagnostics.insert(meta.buffile.clone(), Vec::new());
-
+    // Keep existing diagnostics until the next pull/push update so the UI does not
+    // flicker empty while we wait for the language server. Pull result IDs stay
+    // valid for unchanged reports.
```
krobelus, line 74: is this change something that also helps cases where the server does not support pull diagnostics?
If yes, then it should be in a separate (precursor) commit (as should be the eslint fix), to communicate this fact.
If not, then we would make this conditional on pull diagnostic support.

As far as I can remember, push-diagnostics already have the tendency to linger (in the editor gutter) because even though we delete the diagnostics here,
we probably wait for the next `textDocument/publishDiagnostics` from the server.
Maybe there is an easy fix for that without requiring pull diagnostics..

src/workspace.rs
```diff
@@ -98,14 +98,22 @@ pub fn configuration(
...
-                            return settings.clone();
+                        };
+                        if eslint_workaround && section.is_empty() {
+                            // vscode-eslint expects JSON null for unset nodePath; TOML
+                            // cannot express null, so fill it in here.
+                            let mut settings = settings.clone();
+                            if let Some(obj) = settings.as_object_mut() {
+                                obj.entry("nodePath".to_string()).or_insert(Value::Null);
+                                obj.entry("options".to_string())
+                                    .or_insert(Value::Object(Default::default()));
```
krobelus, line 114: this is quite strange of vscode-eslint; why wouldn't they accept unset nodePath and options?

Whenever we add a workaround like this we should ideally also report the problem upstream to eslint (like we did for https://github.com/microsoft/vscode-eslint/issues/1788), so we can combine efforts with other LSP clients who face the same issues

test/eslint-pull-diagnostics.sh
```diff
@@ -0,0 +1,89 @@
...
+		quiet = false
+		rulesCustomizations = []
+		run = "onType"
+		validate = "on"
+		packageManager = "npm"
+		experimental = {}
+		problems = { shortenToSingleLine = false }
+		codeAction.disableRuleComment = { enable = true, location = "separateLine" }
+		codeAction.showDocumentation = { enable = false }
+		workingDirectory.mode = "auto"
```
krobelus, line 24: IIRC most server settings in rc/servers.kak are exactly the server defaults,
so they are only added there to show the user which settings are available.
In the test we don't need this documenation, so if possible remove all explicit settings except for the ones that are necessary for this.

ficd0:
we need to handle push and pull diagnostic sets independently and merge them for display, some LSPs use both. with this PR checked out, rust-analyzer's native diagnostics appear but clippy doesn't. logs indicate those are sent via the publish method.

rc/servers.kak
```diff
@@ -283,7 +283,8 @@ hook -group lsp-filetype-javascript global BufSetOption filetype=(?:javascript|t
     # set-option buffer lsp_servers %opt{lsp_server_biome}
     # set-option buffer lsp_servers %{
     #     [eslint-language-server]
-    #     root_globs = [".eslintrc", ".eslintrc.json"]
+    #     root_globs = [".eslintrc", ".eslintrc.json", "eslint.config.js", "package.json"]
+    #     command = "vscode-eslint-language-server"
```
krobelus, line 287: why add this line, that is, why prefer `vscode-eslint-language-server` over `eslint-language-server`?
The former is a symlink to the latter, right?

src/diagnostics.rs
```diff
@@ -7,33 +8,171 @@ use crate::util::*;
...
+    let path = uri_to_file_path(&uri);
+    let path = path.to_str().unwrap();
+    match report {
+        DocumentDiagnosticReportKind::Full(report) => {
+            let (result_id, items) = (report.result_id, report.items);
+            if let Some(result_id) = result_id {
+                ctx.diagnostic_pull_result_ids
+                    .insert((server_id, path.to_string()), result_id);
+            }
+            set_server_diagnostics(server_id, path, items, ctx);
```
krobelus, line 131: AFAICT, `set_server_diagnostics()` will replace all diagnostics for `path` with the given ones.
This means that pull and push diagnostics step on each others' toes, no?
Since either one replaces all results from the other.
Shouldn't we merge them somehow?
Maybe keep two different maps `pull_diagnostics` and `push_diagnostics`?

test/eslint-pull-diagnostics.sh
```diff
@@ -0,0 +1,76 @@
...
+
+# Fix the diagnostics and ensure pull clears the gutter.
+cat > main.js << 'JS'
+console.log(1);
+JS
+printf '%s\n' 'evaluate-commands -client client0 %{
+	edit! main.js
+}' | kak -p "$test_kak_session"
+test_sleep_until '! test_tmux capture-pane -p | grep -qE "X|W "'
+test_tmux capture-pane -p | sed 1q
```
krobelus, line 75: I was doing some cleanups before I was gonna merging this;
but I'm no longer sure this is ready because the merging is not really implemented.
(Please unmark as draft when you think it's read.)
Here's my current state.

```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index 58a57cdb22..f5306e657c 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Additions:
 - Default configuration for `kotlin`, using `intellij-server`.
+- Support the LSP diagnostic pull model (`textDocument/diagnostic`) and
+  `workspace/diagnostic/refresh`. When a language server advertises
+  `diagnosticProvider`, kak-lsp pulls diagnostics after open/change instead of
+  relying on `textDocument/publishDiagnostics`.
 
 Fixes:
 - Fix a regression in 21.0.1 that stopped sending save notifications to language servers that advertise `save: true` or omit `includeText` from their save options.
diff --git a/rc/servers.kak b/rc/servers.kak
index c21bea5a76..524cbb650f 100644
--- a/rc/servers.kak
+++ b/rc/servers.kak
@@ -289,7 +289,7 @@
     # set-option buffer lsp_servers %opt{lsp_server_biome}
     # set-option buffer lsp_servers %{
     #     [eslint-language-server]
-    #     root_globs = [".eslintrc", ".eslintrc.json"]
+    #     root_globs = [".eslintrc", ".eslintrc.json", "eslint.config.js", "package.json"]
     #     args = ["--stdio"]
     #     workaround_eslint = true
     #     [eslint-language-server.settings]
@@ -299,10 +299,12 @@
     #     rulesCustomizations = []
     #     run = "onType"
     #     validate = "on"
+    #     packageManager = "npm"
     #     experimental = {}
     #     problems = { shortenToSingleLine = false }
     #     codeAction.disableRuleComment = { enable = true, location = "separateLine" }
     #     codeAction.showDocumentation = { enable = false }
+    #     workingDirectory.mode = "auto"
     # }
     # set-option buffer lsp_servers %{
     #     [tailwindcss-language-server]
diff --git a/src/capabilities.rs b/src/capabilities.rs
index cbf249a942..6f4d0cc462 100644
--- a/src/capabilities.rs
+++ b/src/capabilities.rs
@@ -115,7 +115,9 @@
                             inlay_hint: Some(InlayHintWorkspaceClientCapabilities {
                                 refresh_support: Some(true),
                             }),
-                            diagnostic: None,
+                            diagnostic: Some(DiagnosticWorkspaceClientCapabilities {
+                                refresh_support: Some(true),
+                            }),
                         }),
                         text_document: Some(TextDocumentClientCapabilities {
                             synchronization: Some(TextDocumentSyncClientCapabilities {
@@ -332,7 +334,10 @@
                                 dynamic_registration: Some(false),
                                 resolve_support: None,
                             }),
-                            diagnostic: None,
+                            diagnostic: Some(DiagnosticClientCapabilities {
+                                dynamic_registration: Some(false),
+                                related_document_support: Some(true),
+                            }),
                             inline_completion: None,
                         }),
                         window: Some(WindowClientCapabilities {
@@ -465,6 +470,7 @@
 pub const CAPABILITY_HOVER: &str = "lsp-hover";
 pub const CAPABILITY_IMPLEMENTATION: &str = "lsp-implementation";
 pub const CAPABILITY_INLAY_HINTS: &str = "lsp-inlay-hints";
+pub const CAPABILITY_PULL_DIAGNOSTIC: &str = "textDocument/diagnostic";
 pub const CAPABILITY_RANGE_FORMATTING: &str = "lsp-range-formatting";
 pub const CAPABILITY_REFERENCES: &str = "lsp-references";
 pub const CAPABILITY_RENAME: &str = "lsp-rename";
@@ -576,6 +582,7 @@
             Some(OneOf::Right(_)) => true,
             None => false,
         },
+        CAPABILITY_PULL_DIAGNOSTIC => server_capabilities.diagnostic_provider.is_some(),
         CAPABILITY_RANGE_FORMATTING => match server_capabilities.document_range_formatting_provider
         {
             Some(OneOf::Left(ok)) => ok,
@@ -684,6 +691,7 @@
             .or_default()
             .push(server_name);
         probe_feature(to_editor, entry, &mut features, CAPABILITY_INLAY_HINTS);
+        probe_feature(to_editor, entry, &mut features, CAPABILITY_PULL_DIAGNOSTIC);
 
         // NOTE controller should park request for capabilities until they are available thus it should
         // be safe to unwrap here (otherwise something unexpectedly wrong and it's better to panic)
diff --git a/src/context.rs b/src/context.rs
index 9b08965363..5fe4adabfe 100644
--- a/src/context.rs
+++ b/src/context.rs
@@ -78,6 +78,9 @@
     pub completion_last_client: Option<ClientId>,
     pub config: Config,
     pub diagnostics: HashMap<String, Vec<(ServerId, Diagnostic)>>,
+    /// Opaque `resultId` values from document diagnostic pull responses, keyed by
+    /// (server, buffile). Used as `previousResultId` on subsequent pulls.
+    pub diagnostic_pull_result_ids: HashMap<(ServerId, String), String>,
     pub documents: HashMap<String, Document>,
     pub dynamic_config: DynamicConfig,
     pub inlay_hints: HashMap<String, Vec<(ServerId, InlayHint)>>,
@@ -118,6 +121,7 @@
             completion_last_client: None,
             config,
             diagnostics: Default::default(),
+            diagnostic_pull_result_ids: Default::default(),
             documents: Default::default(),
             dynamic_config: DynamicConfig::default(),
             inlay_hints: Default::default(),
diff --git a/src/controller.rs b/src/controller.rs
index c0a95aaf30..ba8fc7e21c 100644
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -43,7 +43,7 @@
 use jsonrpc_core::{Call, ErrorCode, MethodCall, Output, Params};
 use lean::EditorPlainGoalParams;
 use libc::O_NONBLOCK;
-use lsp_types::error_codes::CONTENT_MODIFIED;
+use lsp_types::error_codes::{CONTENT_MODIFIED, SERVER_CANCELLED};
 use lsp_types::notification::Notification;
 use lsp_types::request::Request;
 use lsp_types::*;
@@ -1103,9 +1103,10 @@
                                         meta.client.clone(),
                                         &failure.id,
                                     );
-                                    if failure.error.code
-                                        == ErrorCode::ServerError(CONTENT_MODIFIED)
-                                    {
+                                    if matches!(
+                                        failure.error.code,
+                                        ErrorCode::ServerError(CONTENT_MODIFIED | SERVER_CANCELLED)
+                                    ) {
                                         debug!(
                                             ctx.to_editor(),
                                             "Error response from server {}: {:?}",
@@ -1156,9 +1157,13 @@
                                         }
                                     }
                                     match failure.error.code {
-                                        code if code
-                                            == ErrorCode::ServerError(CONTENT_MODIFIED)
-                                            || method == request::CodeActionRequest::METHOD => {}
+                                        code if matches!(
+                                            code,
+                                            ErrorCode::ServerError(
+                                                CONTENT_MODIFIED | SERVER_CANCELLED
+                                            )
+                                        ) || method
+                                            == request::CodeActionRequest::METHOD => {}
                                         code => {
                                             let msg = match code {
                                                 ErrorCode::MethodNotFound => format!(
@@ -2072,6 +2077,29 @@
             );
             Ok(serde_json::Value::Null)
         }
+        request::WorkspaceDiagnosticRefresh::METHOD => {
+            // Invalidate cached result ids so the server recomputes.
+            ctx.diagnostic_pull_result_ids
+                .retain(|&(sid, _), _| sid != server_id);
+
+            // Re-pull diagnostics for every open document that talks to this server.
+            let docs: Vec<_> = ctx
+                .documents
+                .iter()
+                .map(|(buffile, doc)| (buffile.clone(), doc.version))
+                .collect();
+            for (buffile, version) in docs {
+                let pull_meta = EditorMeta {
+                    session: ctx.session.clone(),
+                    buffile,
+                    version,
+                    servers: vec![server_id],
+                    ..Default::default()
+                };
+                diagnostics::pull_document_diagnostics(pull_meta, ctx);
+            }
+            Ok(serde_json::Value::Null)
+        }
         _ => {
             warn!(ctx.to_editor(), "Unsupported method: {}", method);
             Err(jsonrpc_core::Error::new(
diff --git a/src/diagnostics.rs b/src/diagnostics.rs
index 48b8709367..ae4c4628ea 100644
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -1,3 +1,4 @@
+use crate::capabilities::{attempt_server_capability, CAPABILITY_PULL_DIAGNOSTIC};
 use crate::context::*;
 use crate::markup::escape_kakoune_markup;
 use crate::position::*;
@@ -7,14 +8,136 @@
 use itertools::EitherOrBoth;
 use itertools::Itertools;
 use jsonrpc_core::Params;
+use lsp_types::request::DocumentDiagnosticRequest;
 use lsp_types::*;
 use std::collections::HashMap;
 use std::fmt::Write as _;
 
-pub fn publish_diagnostics(server_id: ServerId, params: Params, ctx: &mut Context) {
-    let params: PublishDiagnosticsParams = params.parse().expect("Failed to parse params");
-    let buffile = uri_to_file_path(&params.uri);
-    let buffile = buffile.to_str().unwrap();
+/// Request document diagnostics via the LSP pull model for servers that advertise
+/// `diagnosticProvider`.
+pub fn pull_document_diagnostics(meta: EditorMeta, ctx: &mut Context) {
+    let eligible_servers: Vec<_> = ctx
+        .servers(&meta)
+        .filter(|srv| attempt_server_capability(ctx, *srv, &meta, CAPABILITY_PULL_DIAGNOSTIC))
+        .map(|(server_id, server)| {
+            let identifier = server
+                .capabilities
+                .as_ref()
+                .and_then(|caps| caps.diagnostic_provider.as_ref())
+                .and_then(|provider| match provider {
+                    DiagnosticServerCapabilities::Options(options) => options.identifier.clone(),
+                    DiagnosticServerCapabilities::RegistrationOptions(options) => {
+                        options.diagnostic_options.identifier.clone()
+                    }
+                });
+            let previous_result_id = ctx
+                .diagnostic_pull_result_ids
+                .get(&(server_id, meta.buffile.clone()))
+                .cloned();
+            (
+                server_id,
+                vec![DocumentDiagnosticParams {
+                    text_document: TextDocumentIdentifier {
+                        uri: file_path_to_uri(&meta.buffile),
+                    },
+                    identifier,
+                    previous_result_id,
+                    work_done_progress_params: Default::default(),
+                    partial_result_params: Default::default(),
+                }],
+            )
+        })
+        .collect();
+    if eligible_servers.is_empty() {
+        return;
+    }
+
+    ctx.call::<DocumentDiagnosticRequest, _>(
+        meta,
+        RequestParams::Each(eligible_servers.into_iter().collect()),
+        move |ctx, meta, results| {
+            // Drop stale responses after the buffer changed again.
+            if ctx.documents.get(&meta.buffile).map(|doc| doc.version) != Some(meta.version) {
+                return;
+            }
+            for (server_id, result) in results {
+                handle_document_diagnostic_result(server_id, &meta.buffile, result, ctx);
+            }
+        },
+    );
+}
+
+fn handle_document_diagnostic_result(
+    server_id: ServerId,
+    buffile: &str,
+    result: DocumentDiagnosticReportResult,
+    ctx: &mut Context,
+) {
+    let report = match result {
+        DocumentDiagnosticReportResult::Report(report) => report,
+        DocumentDiagnosticReportResult::Partial(_) => {
+            warn!(ctx.to_editor(), "Unsupported partial diagnostic report");
+            return;
+        }
+    };
+
+    match report {
+        DocumentDiagnosticReport::Full(report) => {
+            let (result_id, items) = (
+                report.full_document_diagnostic_report.result_id,
+                report.full_document_diagnostic_report.items,
+            );
+            if let Some(result_id) = result_id {
+                ctx.diagnostic_pull_result_ids
+                    .insert((server_id, buffile.to_string()), result_id);
+            }
+            set_server_diagnostics(server_id, buffile, items, ctx);
+            handle_related_document_report(server_id, report.related_documents, ctx);
+        }
+        DocumentDiagnosticReport::Unchanged(report) => {
+            ctx.diagnostic_pull_result_ids.insert(
+                (server_id, buffile.to_string()),
+                report.unchanged_document_diagnostic_report.result_id,
+            );
+            handle_related_document_report(server_id, report.related_documents, ctx);
+        }
+    }
+}
+
+fn handle_related_document_report(
+    server_id: ServerId,
+    related_documents: Option<HashMap<Uri, DocumentDiagnosticReportKind>>,
+    ctx: &mut Context,
+) {
+    let Some(related) = related_documents else {
+        return;
+    };
+    for (uri, related_report) in related {
+        let path = uri_to_file_path(&uri);
+        let path = path.to_str().unwrap();
+        match related_report {
+            DocumentDiagnosticReportKind::Full(report) => {
+                let (result_id, items) = (report.result_id, report.items);
+                if let Some(result_id) = result_id {
+                    ctx.diagnostic_pull_result_ids
+                        .insert((server_id, path.to_string()), result_id);
+                }
+                set_server_diagnostics(server_id, path, items, ctx);
+            }
+            DocumentDiagnosticReportKind::Unchanged(report) => {
+                ctx.diagnostic_pull_result_ids
+                    .insert((server_id, path.to_string()), report.result_id);
+            }
+        }
+    }
+}
+
+fn set_server_diagnostics(
+    server_id: ServerId,
+    buffile: &str,
+    items: Vec<Diagnostic>,
+    ctx: &mut Context,
+) {
     let mut diagnostics: Vec<_> = ctx
         .diagnostics
         .remove(buffile)
@@ -22,18 +145,28 @@
         .into_iter()
         .filter(|(id, _)| id != &server_id)
         .collect();
-    let params: Vec<_> = params
-        .diagnostics
-        .into_iter()
-        .map(|d| (server_id, d))
-        .collect();
-    diagnostics.extend(params);
+    diagnostics.extend(items.into_iter().map(|d| (server_id, d)));
     ctx.diagnostics.insert(buffile.to_string(), diagnostics);
-    let document = ctx.documents.get(buffile);
-    if document.is_none() {
+    refresh_diagnostics_display(buffile, ctx);
+}
+
+// Note: Some language servers (such as rust-analyzer) use a hybrid model where native/syntax
+// diagnostics are served via pull requests (textDocument/diagnostic), while flycheck/clippy/rustc
+// diagnostics are pushed asynchronously via textDocument/publishDiagnostics (see
+// https://github.com/rust-lang/rust-analyzer/issues/18709).
+// Both pull responses and push notifications update the diagnostics collection so all diagnostics
+// are preserved and merged for display.
+pub fn publish_diagnostics(server_id: ServerId, params: Params, ctx: &mut Context) {
+    let params: PublishDiagnosticsParams = params.parse().expect("Failed to parse params");
+    let buffile = uri_to_file_path(&params.uri);
+    let buffile = buffile.to_str().unwrap();
+    set_server_diagnostics(server_id, buffile, params.diagnostics, ctx);
+}
+
+fn refresh_diagnostics_display(buffile: &str, ctx: &mut Context) {
+    let Some(document) = ctx.documents.get(buffile) else {
         return;
-    }
-    let document = document.unwrap();
+    };
     let version = document.version;
     let diagnostics = &ctx.diagnostics[buffile];
     let diagnostics_orderd_by_severity = diagnostics
diff --git a/src/text_sync.rs b/src/text_sync.rs
index 3b1b79fc3b..b834b6cd4e 100644
--- a/src/text_sync.rs
+++ b/src/text_sync.rs
@@ -4,6 +4,7 @@
     time::Duration,
 };
 
+use crate::diagnostics;
 use crate::thread_worker::Worker;
 use crate::types::*;
 use crate::util::file_path_to_uri;
@@ -43,6 +44,7 @@
     for &server_id in &meta.servers {
         ctx.notify::<DidOpenTextDocument>(server_id, params.clone());
     }
+    diagnostics::pull_document_diagnostics(meta, ctx);
 }
 
 pub fn text_document_did_change(
@@ -67,8 +69,9 @@
 
     // Resets metadata for buffer.
     ctx.documents.insert(meta.buffile.clone(), document);
-    ctx.diagnostics.insert(meta.buffile.clone(), Vec::new());
-
+    // Keep existing diagnostics until the next pull/push update so the UI does not
+    // flicker empty while we wait for the language server. Pull result IDs stay
+    // valid for unchanged reports.
     let req_params = DidChangeTextDocumentParams {
         text_document: VersionedTextDocumentIdentifier {
             uri,
@@ -83,10 +86,13 @@
     for &server_id in &meta.servers {
         ctx.notify::<DidChangeTextDocument>(server_id, req_params.clone());
     }
+    diagnostics::pull_document_diagnostics(meta, ctx);
 }
 
 pub fn text_document_did_close(meta: EditorMeta, ctx: &mut Context) {
     ctx.documents.remove(&meta.buffile);
+    ctx.diagnostic_pull_result_ids
+        .retain(|(_, path), _| path != &meta.buffile);
     let uri = file_path_to_uri(&meta.buffile);
     let params = DidCloseTextDocumentParams {
         text_document: TextDocumentIdentifier { uri },
diff --git a/src/workspace.rs b/src/workspace.rs
index 687b58ceab..69cb52e87a 100644
--- a/src/workspace.rs
+++ b/src/workspace.rs
@@ -98,14 +98,28 @@
                 .map(|section| match &settings {
                     None => Value::Null,
                     Some(settings) => {
-                        if if is_using_legacy_toml(&ctx.config) {
+                        let eslint_workaround = if is_using_legacy_toml(&ctx.config) {
                             ctx.server_config(&meta, server_name)
                                 .is_some_and(|cfg| cfg.workaround_eslint == Some(true))
                         } else {
                             ctx.server(server_id).workaround_eslint
-                        } && section.is_empty()
-                        {
-                            return settings.clone();
+                        };
+                        if eslint_workaround && section.is_empty() {
+                            // vscode-eslint expects specific fields to be present in configuration.
+                            // Fill in reasonable defaults for missing optional fields that TOML cannot easily express.
+                            let mut settings = settings.clone();
+                            if let Some(obj) = settings.as_object_mut() {
+                                obj.entry("nodePath".to_string()).or_insert(Value::Null);
+                                obj.entry("options".to_string())
+                                    .or_insert(Value::Object(Default::default()));
+                                obj.entry("experimental".to_string())
+                                    .or_insert(Value::Object(Default::default()));
+                                obj.entry("problems".to_string())
+                                    .or_insert(serde_json::json!({ "shortenToSingleLine": false }));
+                                obj.entry("rulesCustomizations".to_string())
+                                    .or_insert(Value::Array(Vec::new()));
+                            }
+                            return settings;
                         }
                         settings.get(section).unwrap_or(&Value::Null).clone()
                     }
diff --git a/test/eslint-pull-diagnostics.sh b/test/eslint-pull-diagnostics.sh
new file mode 100755
index 0000000000..397a4f22c6
--- /dev/null
+++ b/test/eslint-pull-diagnostics.sh
@@ -0,0 +1,75 @@
+#!/bin/sh
+
+# Integration test for the LSP diagnostic pull model against the modern
+# eslint-language-server (advertises diagnosticProvider).
+
+# REQUIRES: command -v eslint-language-server
+# REQUIRES: command -v eslint
+# REQUIRES: command -v npm
+
+. test/lib.sh
+
+cat >> .config/kak/kakrc << 'KAK'
+set-option global lsp_diagnostic_line_error_sign   ' X'
+set-option global lsp_diagnostic_line_warning_sign 'W '
+
+hook global BufSetOption filetype=javascript %{
+	set-option buffer lsp_servers %{
+		[eslint-language-server]
+		root_globs = ["package.json", "eslint.config.js", "eslint.config.mjs", ".eslintrc.json"]
+		args = ["--stdio"]
+		workaround_eslint = true
+		[eslint-language-server.settings]
+		workingDirectory.mode = "auto"
+		validate = "on"
+	}
+}
+KAK
+
+# Local eslint so the language server can resolve the library from the project.
+npm init -y >/dev/null 2>&1
+npm install --ignore-scripts --no-save --no-package-lock eslint@9 @eslint/js@9 >/dev/null 2>&1
+
+cat > eslint.config.mjs << 'ESLINT'
+import js from "@eslint/js";
+
+export default [
+	js.configs.recommended,
+	{
+		rules: {
+			"no-unused-vars": "error",
+			"no-undef": "error",
+		},
+	},
+];
+ESLINT
+
+cat > package.json << 'JSON'
+{
+  "name": "kak-lsp-eslint-pull-test",
+  "private": true,
+  "type": "module"
+}
+JSON
+
+cat > main.js << 'JS'
+const unused = 1;
+console.log(missing);
+JS
+
+test_tmux_kak_start 'edit main.js; set-option buffer filetype javascript'
+test_sleep_until 'test_tmux capture-pane -p | grep -qE "X|W "'
+test_tmux capture-pane -p | sed 2q
+# CHECK: {{( X|W )}}const unused = 1;
+# CHECK: {{( X|W )}}console.log(missing);
+
+# Fix the diagnostics and ensure pull clears the gutter.
+cat > main.js << 'JS'
+console.log(1);
+JS
+printf '%s\n' 'evaluate-commands -client client0 %{
+	edit! main.js
+}' | kak -p "$test_kak_session"
+test_sleep_until '! test_tmux capture-pane -p | grep -qE "X|W "'
+test_tmux capture-pane -p | sed 1q
+# CHECK: {{  }}console.log(1);
```

PR #900, opened 2026-08-30
fix didSave capability check regression

ficd0:
Fixes a regression introduced in 2c44d28 that caused `"save":{}` to be
treated as lacking save support entirely. Fixes the issue with
rust-analyzer diagnostics not refreshing correctly on save.

### DIFF
--- a/src/text_sync.rs
+++ b/src/text_sync.rs
@@ -96,21 +96,27 @@ pub fn text_document_did_close(meta: EditorMeta, ctx: &mut Context) {
     }
 }
 
+fn did_save_include_text(text_document_sync: &Option<TextDocumentSyncCapability>) -> Option<bool> {
+    match text_document_sync {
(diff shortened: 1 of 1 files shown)


krobelus:
already fixed by a345517ed (Handle missing cases of TextDocumentSyncSaveOptions, 2026-08-02), I don't think there's anything left; though the test cases are definitely nice

ficd0:
I don't know how on Earth I missed that... Could be that when I ran git pull it pulled from my fork and not from upstream? Lol. I'm glad it's fixed at least and that explains why you weren't able to reproduce my issue initially. The test cases can help prevent regression at least. I will rebase and remove everything but the test cases from this patch.

ficd0:
@krobelus would you like me to remove everything but the test cases and rebase?

krobelus:
Yeah

